PR #3, opened 2014-06-13
Add charmstore commands.

frankban:
Copy files from core.
Fix import paths, remove core dependencies and improve tests.


rogpeppe:
LGTM although I have a lurking suspicion that it might have been
worth importing the VCS history here.

PR #6, opened 2014-06-23
source distribution for charmstore deployment.

cmars:
Adapted some scripts I've used to freeze a source distribution of the charmstore, for past deployments into a restricted build env.

Since the source build manipulates its own GOPATH, and does some rather drastic things (like git reset --hard), I think its probably safer to keep this stuff in a separate makefile that probably won't be used during normal development.


Makefile.sdist
```diff
@@ -0,0 +1,35 @@
+
+GOPATH := $(shell pwd)/build
+LPUSER := $(shell bzr lp-login)
+
+PACKAGE=github.com/juju/charmstore
+DISTBRANCH=lp:~$(LPUSER)/juju-core/charmstore-sdist  # where to publish sdist
```
bac, line 6: Since we've moved juju/charmstore to github do we want to keep this branch on Launchpad still?

bac:
Casey a while back on a different charm, the webops guys suggested we store dependencies in a swift/S3 bucket rather than in a tarball (juju-gui) or in a download-cache branch on Launchpad as charmworld and others do.

I think for a substantially new deploy we'd be smart to go that route if it is preferred by webops.

Thanks for this branch. I think it gets things rolling even if we adapt it going forward.

I'm not an official review for the charmstore so these comments are just provisional until Roger or Francesco has a look.

cmars:
Sounds good, I figured this would probably end up more as a proof-of-concept than a branch to land. s3/swift sounds like a much better solution for deployment.

Cheers,
Casey

PR #11, opened 2014-06-27
charmstore: simplify serving code

rogpeppe:
We can use method expressions now, and http.StripPrefix
makes the code simpler too, although the tests need
to be changed to account for the fact that the handlers
now mutate the request. The latter is worth doing anyway
though, IMHO, to prevent possible cross-test bleedthrough.

### DIFF
--- a/server.go
+++ b/server.go
@@ -30,25 +30,19 @@ func NewServer(store *Store) (*Server, error) {
 		store: store,
 		mux:   http.NewServeMux(),
 	}
-	s.mux.HandleFunc("/charm-info", func(w http.ResponseWriter, r *http.Request) {
-		s.serveInfo(w, r)
-	})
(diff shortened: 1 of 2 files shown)


themue:
Nice, LGTM.

server.go
```diff
@@ -228,11 +211,7 @@ func (s *Server) serveCharm(w http.ResponseWriter, r *http.Request) {
 
 func (s *Server) serveStats(w http.ResponseWriter, r *http.Request) {
 	// TODO: Adopt a smarter mux that simplifies this logic.
```
johnweldon, line 213: Looks like this TODO is done now?

rogpeppe, line 213: Perhaps you're right. I'd assumed the comment pertained to the strings.Index(base, "/") logic,
but that's not really that complex, is it?

I'll remove the TODO.

PR #15, opened 2014-07-18
move all top level files to old directory

rogpeppe:
This is a purely mechanical change, moving all the existing
charmstore files into the old directory in preparation for
reconstructing the whole thing from scratch by cherry picking
the parts we need from the old code and integrating with the new.

From this point on, CI is broken
because we do not have a working charm store. That's why this
is in the v4 branch.


urosj:
LGTM

mitechie:
I'm concerned with this change. Per the cards on the board we need to provide a smooth transition from the old design, ingestion model, and api to the new, published based, api and design. We want to get our version of the charmstore deployed to store.juju.ubuntu.com and to iterate on it over time and to work on migrating clients. This does not seem like something that is working towards that end. Please correct me if I'm wrong here.

PR #18, opened 2014-07-21
internal/charmstore: add Store.AddCharm method

rogpeppe:
This is just a stop-gap measure so that tests can add charms to
the database and implement endpoints before charm archive
upload is implemented.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -1,19 +1,109 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
 package charmstore
 
 import (
+	"fmt"
+	"log"
+
+	"gopkg.in/juju/charm.v2"
 	"labix.org/v2/mgo"
+
+	"github.com/juju/charmstore/internal/mongodoc"
 )
 
 type Store struct {
-	db *mgo.Database
+	DB StoreDatabase
 }
 
 func newStore(db *mgo.Database) *Store {
-	return &Store{
-		db: db,
+	s := &Store{
+		DB: StoreDatabase{db},
+	}
+	s.ensureIndexes()
+	return s
+}
+
+func (s *Store) ensureIndexes() error {
+	indexes := []struct {
+		c *mgo.Collection
+		i mgo.Index
+	}{{
+		s.DB.Entities(),
+		mgo.Index{Key: []string{"url", "revision"}, Unique: true},
+	}}
+	for _, idx := range indexes {
+		err := idx.c.EnsureIndex(idx.i)
+		if err != nil {
+			log.Printf("error ensuring %s index: %v", idx.c.Name, err)
+			return err
+		}
+	}
+	return nil
+}
+
+// AddCharm adds a charm to the entities collection
+// associated with the given URL.
+// The charm does not have any associated content.
+// TODO fix this to add content too.
+func (s *Store) AddCharm(url *charm.URL, c charm.Charm) error {
+	return s.DB.Entities().Insert(&mongodoc.Entity{
+		URL:                     url,
+		BaseURL:                 baseURL(url),
+		CharmMeta:               c.Meta(),
+		CharmConfig:             c.Config(),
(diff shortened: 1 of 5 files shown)


internal/charmstore/store.go
```diff
@@ -1,19 +1,89 @@
...
+	})
+}
+
+func interfacesForRelations(rels map[string]charm.Relation) []string {
+	interfaces := make(map[string]bool)
+	for _, rel := range rels {
+		interfaces[rel.Interface] = true
+	}
+	result := make([]string, 0, len(interfaces))
+	for int := range interfaces {
```
bac: Since 'int' overrides the 'int' type name it is a bit confusing. Maybe 'iface' or 'interface_'?

internal/charmstore/store.go
```diff
@@ -1,19 +1,89 @@
...
+		CharmConfig:             c.Config(),
+		CharmActions:            c.Actions(),
+		CharmProvidedInterfaces: interfacesForRelations(c.Meta().Provides),
+		CharmRequiredInterfaces: interfacesForRelations(c.Meta().Requires),
+	})
+}
+
+func interfacesForRelations(rels map[string]charm.Relation) []string {
+	interfaces := make(map[string]bool)
+	for _, rel := range rels {
```
bac, line 45: A comment stating you're using the map to remove duplicates would be helpful to first-time readers.

internal/charmstore/store.go
```diff
@@ -1,19 +1,89 @@
...
+		interfaces[rel.Interface] = true
+	}
+	result := make([]string, 0, len(interfaces))
+	for int := range interfaces {
+		result = append(result, int)
+	}
+	// TODO is it worth sorting this?
```
bac: We think not.  Please remove the question.

internal/charmstore/store_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (s *StoreSuite) TestAddCharm(c *gc.C) {
+	store := newStore(s.Session.DB("foo"))
+	url := charm.MustParseURL("cs:precise/wordpress-23")
+	wordpress := testing.Charms.CharmDir("wordpress")
+	err := store.AddCharm(url, wordpress)
+	c.Assert(err, gc.IsNil)
+
+	var doc mongodoc.Entity
+	err = store.DB.Entities().Find(bson.D{{"_id", "cs:precise/wordpress-23"}}).One(&doc)
```
bac: Using FindId works and is simpler.  What do you think?
err = store.DB.Entities.FindId("cs:precise/wordpress-23").One(&doc)

internal/charmstore/store_test.go
```diff
@@ -0,0 +1,60 @@
...
+		BaseURL:                 mustParseReference("cs:wordpress"),
+		CharmMeta:               wordpress.Meta(),
+		CharmActions:            wordpress.Actions(),
+		CharmConfig:             wordpress.Config(),
+		CharmProvidedInterfaces: []string{"http", "logging", "monitoring"},
+		CharmRequiredInterfaces: []string{"mysql", "varnish"},
+	})
+
+	// Try inserting the charm again - it should fail because the charm is already
+	// there
```
bac: Add full-stop.

bac:
:+1: \* 2, with some small suggestions.  Thanks for the branch.

internal/charmstore/store.go
rogpeppe: Done

internal/charmstore/store.go
rogpeppe, line 45: Done

internal/charmstore/store.go
rogpeppe: Done

internal/charmstore/store_test.go
rogpeppe: Good suggestion. Done.

internal/charmstore/store_test.go
rogpeppe: Done

PR #19, opened 2014-07-21
charmstore: refactor external interface

rogpeppe:
We lose the version registration logic, and make versions
available to external packages without exposing
internal implementation details in the godoc.

### DIFF
--- a/server.go
+++ b/server.go
@@ -14,6 +14,7 @@ import (
 	"github.com/juju/charmstore/internal/v4"
 )
 
+// Versions of the API that can be served.
 const (
 	V4 = "v4"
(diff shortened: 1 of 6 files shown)


server.go
```diff
@@ -4,15 +4,46 @@
...
+// Versions of the API that can be served.
+const (
+	V4 = "v4"
+)
+
+var versions = map[string]func(*charmstore.Store) http.Handler{
+	V4: v4.New,
+}
+
+// Versions returns all known API versions.
```
frankban, line 26: // Versions returns all known API version strings in alphabetical order?

PR #21, opened 2014-07-21
charmstore: refactor external interface

rogpeppe:
We lose the version registration logic, and make versions
available to external packages without exposing
internal implementation details in the godoc.


rogpeppe:
reviewed as PR #19

PR #22, opened 2014-07-21
API charm config

bac:
Add API call for charm-config.
Includes Roger and Uros' work for charm-meta API call.
Update dependencies to get charm.Bundle.


urosj:
LGTM

PR #27, opened 2014-07-22
internal/router: add support for id resolving

rogpeppe:
We allow non-specific ids in API paths.
The details of the id resolution are kept
out of the router - that responsibility is
up to the API version implemention,
and will be implemented in a subsequent branch.

### DIFF
--- a/internal/router/router_test.go
+++ b/internal/router/router_test.go
@@ -165,10 +165,23 @@ var routerTests = []struct {
 		},
 	},
 	urlStr:     "http://example.com/~joe/wordpress/foo",
-	resolveURL: resolveURL("precise", 34),
+	resolveURL: newResolveURL("precise", 34),
 	expectCode: http.StatusOK,
 	expectBody: idHandlerTestResp{
 		CharmURL: "cs:~joe/precise/wordpress-34",
+	},
+}, {
+	about: "id with error on resolving",
+	handlers: Handlers{
+		Id: map[string]IdHandler{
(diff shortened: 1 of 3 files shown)


internal/router/router_test.go
```diff
@@ -255,12 +269,39 @@ var routerTests = []struct {
...
+			url.Series = series
+		}
+		if url.Revision == -1 {
+			url.Revision = revision
+		}
+		return nil
+	}
+}
+
+func resolveURLError(*charm.URL) error {
```
frankban, line 300: Is this used? Do we need to add a test for resolveUrl error?

internal/router/router_test.go
```diff
@@ -255,12 +269,39 @@ var routerTests = []struct {
 	},
 }}
 
+// resolveURL returns a URL resolver that resolves
+// unspecified series and revision to the given series
+// and revision.
+func resolveURL(series string, revision int) func(*charm.URL) error {
```
frankban: s/resolveURL/makeResolveURL/ ?

frankban:
:+1:  with possible minor changes.
Thank you!

internal/router/router_test.go
rogpeppe, line 300: good point.

internal/router/router_test.go
rogpeppe: done

rogpeppe:
:shipit:

jujugui:

PR #28, opened 2014-07-23
Implement the store.AddBundle method

frankban:
Also implement a cstomized CharmURL type so that we can easily deal with partial URLS.

Updated dependencies so that we are now using the updated names/charm.v2 packages.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -88,12 +88,18 @@ func (s *Store) AddBundle(url *charm.URL, b charm.Bundle) error {
 }
 
 func bundleCharms(data *charm.BundleData) ([]*params.CharmURL, error) {
-	urls := make([]*params.CharmURL, 0, len(data.Services))
+	// Use a map to de-duplicate the URL list: a bundle can include services
+	// deployed by the same charm.
+	urlMap := make(map[string]*params.CharmURL)
 	for _, service := range data.Services {
 		url, err := params.ParseURL(service.Charm)
 		if err != nil {
 			return nil, err
 		}
+		urlMap[url.String()] = url
+	}
+	urls := make([]*params.CharmURL, 0, len(urlMap))
+	for _, url := range urlMap {
 		urls = append(urls, url)
 	}
 	return urls, nil
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/juju/charmstore/internal/mongodoc"
 	"github.com/juju/charmstore/internal/router"
+	"github.com/juju/charmstore/params"
 )
 
 // Store represents the underlying charm store data store.
@@ -32,9 +33,10 @@ func NewStore(db *mgo.Database) *Store {
 // The charm does not have any associated content.
 // TODO fix this to add content too.
 func (s *Store) AddCharm(url *charm.URL, c charm.Charm) error {
+	charmUrl := (*params.CharmURL)(url)
 	return s.DB.Entities().Insert(&mongodoc.Entity{
-		URL:                     url,
-		BaseURL:                 baseURL(url),
+		URL:                     charmUrl,
+		BaseURL:                 baseURL(charmUrl),
 		CharmMeta:               c.Meta(),
 		CharmConfig:             c.Config(),
 		CharmActions:            c.Actions(),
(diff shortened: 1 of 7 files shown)


internal/charmstore/store.go
```diff
@@ -69,7 +72,31 @@ var errNotImplemented = fmt.Errorf("not implemented")
...
+		URL:          charmUrl,
+		BaseURL:      baseURL(charmUrl),
+		BundleData:   bundleData,
+		BundleReadMe: b.ReadMe(),
+		BundleCharms: urls,
+	})
+}
+
+func bundleCharms(data *charm.BundleData) ([]*params.CharmURL, error) {
+	urls := make([]*params.CharmURL, 0, len(data.Services))
```
rogpeppe: I think we should de-duplicate here by using a map.

something like:

```
urls := make(map[string] *params.CharmURL)
for _, _, service := range data.Services {
        url, err := params.ParseURL(service.Charm)
        if err != nil {
            return nil, err
    }
    urls[url.String()] = url
}
    // convert to slice
return urls, nil
```

params/url_test.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+var urlCodecs = []struct {
+	name      string
+	data      string
+	marshal   func(interface{}) ([]byte, error)
+	unmarshal func([]byte, interface{}) error
+}{{
+	name:      "bson",
+	data:      "%\x00\x00\x00\x02url\x00\x17\x00\x00\x00cs:trusty/wordpress-42\x00\x00",
```
rogpeppe: Rather than having raw bson data here, I'd suggest that we just check that the data round-trips ok.

internal/charmstore/store.go
frankban: Good point, fixed.

params/url_test.go
frankban: I'd also like to check the serialized data looks like what we expect. I improved the way this is tested using jc.Contains.

params/url_test.go
```diff
@@ -0,0 +1,92 @@
...
+	name:      "bson",
+	data:      "cs:trusty/wordpress-42",
+	marshal:   bson.Marshal,
+	unmarshal: bson.Unmarshal,
+}, {
+	name:      "json",
+	data:      `{"URL":"cs:trusty/wordpress-42"}`,
+	marshal:   json.Marshal,
+	unmarshal: json.Unmarshal,
+}}
```
rogpeppe, line 57: We should have a test for a charm URL without a series, as that's the whole motivation for these changes.

params/url_test.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+var urlCodecs = []struct {
+	name      string
+	data      string
+	marshal   func(interface{}) ([]byte, error)
+	unmarshal func([]byte, interface{}) error
+}{{
+	name:      "bson",
+	data:      "cs:trusty/wordpress-42",
```
rogpeppe: This feels a little odd, as it's just a part of the raw BSON. We should be able
to avoid dependency on the raw BSON encoding completely, I think.

```
data: map[string]interface{} {"URL": "cs:trusty/wordpress-42"},
```

then check with

```
var gotData map[string]interface{]
err := test.unmarshal(serialized, &gotData)
c.Assert(err, gc.IsNil)
c.Assert(gotData, jc.DeepEquals, test.data)
```

params/url_test.go
```diff
@@ -0,0 +1,92 @@
...
+	charmURL *params.CharmURL
+}
+
+var _ = gc.Suite(&urlSuite{})
+
+type urlDoc struct {
+	URL *params.CharmURL
+}
+
+var urlCodecs = []struct {
```
rogpeppe: var marshalTests = 

,ight be a more conventional name?

params/url_test.go
```diff
@@ -0,0 +1,92 @@
...
+	c.Assert(url.Schema, gc.Equals, "cs")
+	c.Assert(url.Series, gc.Equals, "precise")
+	c.Assert(url.Name, gc.Equals, "django")
+	c.Assert(url.Revision, gc.Equals, 42)
+}
+
+func (s *urlSuite) TestParseURLPartial(c *gc.C) {
+	url, err := params.ParseURL("wordpress")
+	c.Assert(err, gc.IsNil)
+	c.Assert(url.Schema, gc.Equals, "cs")
```
rogpeppe, line 90: c.Assert(url, gc.Equals, params.ParseURL{ ... })

?

rogpeppe:
LGTM modulo a few minor suggestions above.

params/url_test.go
frankban: Done.

params/url_test.go
frankban: Nice, done.

params/url_test.go
frankban, line 57: Good point. Done.

frankban:
Thanks for the review!
:shipit:

jujugui:

PR #31, opened 2014-07-25
Add charm and bundle contents to the blob store.

frankban:


internal/blobstore/blobstore.go
```diff
@@ -49,6 +51,15 @@ type ContentChallengeResponse struct {
...
 }
 
+func getTrustedChallengeResponse() *ContentChallengeResponse {
+	rand.Seed(time.Now().UnixNano())
+	return &ContentChallengeResponse{strconv.Itoa(rand.Int()), strconv.Itoa(rand.Int())}
+}
+
+// Trusted is a challenge response we trust.
+// When a trusted proof is provided, the associated content is always uploaded.
+var Trusted = getTrustedChallengeResponse()
```
rogpeppe: just

```
var Trusted = new(ContentChallengeResponse)
```

should be fine here. It's just a token - we don't care about the contents.

internal/charmstore/archive.go
```diff
@@ -0,0 +1,59 @@
...
+	io.ReadSeeker
+}
+
+func (nopCloser) Close() error {
+	return nil
+}
+
+// NopCloser returns a blobstore.ReadSeekCloser with a no-op Close method
+// wrapping the provided ReadSeeker r.
+func NopCloser(r io.ReadSeeker) blobstore.ReadSeekCloser {
```
rogpeppe: Let's not export this.

rogpeppe: Also, better at the end, so it's not the first thing you see when looking at the file.

internal/charmstore/archive.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/juju/charmstore/internal/blobstore"
+)
+
+type archiverTo interface {
+	ArchiveTo(io.Writer) error
+}
+
+// getArchive is used to turn the current charm and bundle implementations
+// into ReadSeekClosers for the corresponding archive.
```
rogpeppe: s/the/their/
?

internal/charmstore/archive.go
```diff
@@ -0,0 +1,59 @@
...
+func getArchive(c interface{}) (blobstore.ReadSeekCloser, error) {
+	var path string
+	switch c := c.(type) {
+	case archiverTo:
+		// E.g. charm.CharmDir or charm.BundleDir.
```
rogpeppe: s/E.g./For example: /

(better to avoid latinisms, I think)

internal/charmstore/store.go
```diff
@@ -5,38 +5,70 @@ package charmstore
...
 	"github.com/juju/charmstore/internal/router"
 	"github.com/juju/charmstore/params"
 )
 
-// Store represents the underlying charm store data store.
+// Store represents the underlying charm and blob data stores.
 type Store struct {
-	DB StoreDatabase
+	DB     StoreDatabase
+	BlobDB *blobstore.Store
```
rogpeppe: BlobStore ?

internal/charmstore/store_test.go
```diff
@@ -33,11 +38,16 @@ func (s *StoreSuite) TestAddCharm(c *gc.C) {
 	var doc mongodoc.Entity
 	err = store.DB.Entities().FindId("cs:precise/wordpress-23").One(&doc)
 	c.Assert(err, gc.IsNil)
+
+	// The entity doc has been correctly added to the mongo collection.
```
rogpeppe, line 42: // Check that the entity doc has been correctly added to the mongo collection.
?

internal/charmstore/store_test.go
```diff
@@ -33,11 +38,16 @@ func (s *StoreSuite) TestAddCharm(c *gc.C) {
 	var doc mongodoc.Entity
 	err = store.DB.Entities().FindId("cs:precise/wordpress-23").One(&doc)
 	c.Assert(err, gc.IsNil)
+
```
rogpeppe, line 41: We test that we can add a CharmDir here, but not a CharmArchive.
Similarly for bundles below. Perhaps make this test into a function
that accepts a Charm, so you can run the same logic both with wordpress
the charm dir and wordpress the charm archive.

rogpeppe:
LGTM, thanks, with a few trivials and a little extra test coverage.

internal/charmstore/archive.go
frankban: Done.

internal/charmstore/archive.go
frankban: Done, but why do you think it's better to avoid latinisms?

internal/charmstore/store.go
frankban: Done.

internal/charmstore/store_test.go
frankban, line 42: I usually avoid using "check", "ensure", "assert" or similar in tests. We are already in the tests context, the fact we are checking something should be already clear enough.

internal/charmstore/store_test.go
frankban, line 41: Fixed.

frankban:
:shipit:

jujugui:

PR #34, opened 2014-07-28
internal/router: bulk metadata API implemention

rogpeppe:
We also fix all paths so that they start with / after splitting. This makes them compatible with the usual way that paths are specified in URLs.

### DIFF
--- a/internal/blobstore/blobstore.go
+++ b/internal/blobstore/blobstore.go
@@ -9,11 +9,11 @@ import (
 	"io"
 	"strconv"
 
+	"github.com/juju/blobstore"
+	"github.com/juju/errors"
 	"labix.org/v2/mgo"
 
-	"github.com/juju/blobstore"
 	"github.com/juju/charmstore/internal/multihash"
-	"github.com/juju/errors"
 )
 
 type ReadSeekCloser interface {
--- a/internal/blobstore/blobstore.go
+++ b/internal/blobstore/blobstore.go
@@ -136,6 +136,13 @@ func (s *Store) Put(r io.Reader, size int64, hash string, proof *ContentChalleng
 	}, nil
 }
 
+// PutUnchallenged stream the content from the given reader into blob
+// storage. The content should have the given size and hash. In this case
+// a challenge is never returned and a proof is not required.
+func (s *Store) PutUnchallenged(r io.Reader, size int64, hash string) error {
+	return s.mstore.PutForEnvironment("", hash, r, size)
+}
+
 // Open opens the blob with the given hash.
 func (s *Store) Open(hashSum string) (ReadSeekCloser, int64, error) {
 	r, length, err := s.mstore.GetForEnvironment("", hashSum)
--- a/internal/charmstore/store_test.go
+++ b/internal/charmstore/store_test.go
@@ -61,6 +61,7 @@ func (s *StoreSuite) TestAddBundle(c *gc.C) {
 	var doc mongodoc.Entity
 	err = store.DB.Entities().FindId("cs:bundle/wordpress-simple-42").One(&doc)
 	c.Assert(err, gc.IsNil)
+	// TODO sort doc.BundleCharms to prevent the test sporadically failing.
 	c.Assert(doc, jc.DeepEquals, mongodoc.Entity{
 		URL:          (*params.CharmURL)(url),
 		BaseURL:      mustParseURL("cs:wordpress-simple"),
--- a/internal/charmstore/store_test.go
+++ b/internal/charmstore/store_test.go
@@ -4,6 +4,11 @@
 package charmstore
 
 import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
 	"sort"
 
 	jc "github.com/juju/testing/checkers"
@@ -12,6 +17,7 @@ import (
 	"labix.org/v2/mgo"
 	gc "launchpad.net/gocheck"
 
+	"github.com/juju/charmstore/internal/blobstore"
 	"github.com/juju/charmstore/internal/mongodoc"
 	"github.com/juju/charmstore/internal/storetesting"
 	"github.com/juju/charmstore/params"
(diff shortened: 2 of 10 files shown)


internal/charmstore/store_test.go
```diff
@@ -61,6 +61,7 @@ func (s *StoreSuite) TestAddBundle(c *gc.C) {
 	var doc mongodoc.Entity
 	err = store.DB.Entities().FindId("cs:bundle/wordpress-simple-42").One(&doc)
 	c.Assert(err, gc.IsNil)
+	// TODO sort doc.BundleCharms to prevent the test sporadically failing.
```
urosj: Also need to change order of charms in doc.BundleCharms so that they reflect being sorted.

internal/blobstore/blobstore.go
```diff
@@ -9,11 +9,11 @@ import (
 	"io"
 	"strconv"
 
+	"github.com/juju/blobstore"
+	"github.com/juju/errors"
 	"labix.org/v2/mgo"
 
-	"github.com/juju/blobstore"
 	"github.com/juju/charmstore/internal/multihash"
-	"github.com/juju/errors"
```
frankban: Thanks for this clean up.

internal/router/router.go
```diff
@@ -115,9 +116,7 @@ func (r *Router) serveIds(w http.ResponseWriter, req *http.Request) error {
 		return err
 	}
 	// TODO(rog) can we really just always ignore a trailing slash ?
```
frankban: IIUC this TODO no longer applies.

internal/router/router.go
```diff
@@ -137,7 +136,8 @@ func (r *Router) serveIds(w http.ResponseWriter, req *http.Request) error {
 		return ErrNotFound
 	}
 	req.URL.Path = path
-	resp, err := r.serveMeta(url, w, req)
+	// TODO remove ResponseWriter argument from function passed to JSONHandler
```
frankban: I am not sure I fully understand this comment. Could you clarify? Also, please add a final .

internal/router/router.go
```diff
@@ -137,7 +136,8 @@ func (r *Router) serveIds(w http.ResponseWriter, req *http.Request) error {
...
+	resp, err := r.serveMeta(url, req)
```
urosj, line 143: Although writer is never used in "serveMeta", it does break the common "w, r" api with all other "serve*" functions. Not a problem if we are sure, that the code will never use writers.
Also the comment is not fully understood - why not just change the code if it is required?

internal/router/router.go
```diff
@@ -195,6 +196,37 @@ func (r *Router) metaNames() []string {
 	return names
 }
 
+// serveBulkMeta serves the "bulk" metadata retrieval endpoint
+// that can return information on several ids at once.
+//
+// GET meta/$endpoint?id=$id0[&id=$id1...][$otherflags]
+// http://tinyurl.com/kdrly9f
+func (r *Router) serveBulkMeta(w http.ResponseWriter, req *http.Request) (interface{}, error) {
+	// TODO get the metadata concurrently for each id
```
frankban: // TODO get the metadata concurrently for each id.

internal/router/router.go
rogpeppe: The comment is redundant - I actually made this change. I'll delete the comment.

rogpeppe: Ha, no I didn't. :-)

internal/router/router_test.go
```diff
@@ -265,21 +265,161 @@ var routerTests = []struct {
...
-				"CharmURL": "cs:precise/wordpress-42",
-				"Path":     "",
-				"Flags":    nil,
+			"test": metaHandlerTestResp{
+				CharmURL: "cs:precise/wordpress-42",
 			},
 		},
 	},
+}, {
+	about:  "meta/any, includes with additional path elements",
```
frankban, line 282: Nice tests!

internal/router/router.go
rogpeppe, line 143: I didn't make the change because it seemed like additional unnecessary code churn in a PR that was already a bit too big.

Breaking the commonality is actually a good thing in this case (IMHO) because it would be wrong for the JSON handler function to use the ResponseWriter. If it did, it could break things (for example, if the client set the response code, the JSON handler could not set the encoding type of the response). If a handler really needs the ResponseWriter (for example, because it needs to set some custom headers in the response), it can always avoid using HandleJSON, which is only a trivial wrapper after all.

By omitting the ResponseWriter from HandleJSON, we make it clear that the output of the handler is implied by the returned values (interface{}, error).

frankban:
:+1: thank you!

internal/router/router.go
```diff
@@ -318,17 +351,22 @@ func (r *Router) getter(id interface{}, val interface{}, fields ...string) error
 // after path[i:] and the start of the next
 // element.
 //
-// For example, splitPath("foo/bar/bzr", 4) returns ("bar", 8).
+// For example, splitPath("/foo/bar/bzr", 4) returns ("bar", 7).
```
urosj: splitPathTest does not assert returned "nextIndex" ... that 7 in the comment is fishy - please add nextIndex tests

internal/router/router.go
```diff
@@ -345,13 +383,20 @@ func splitId(path string) (url *charm.URL, rest string, err error) {
...
-	ref, series, err := charm.ParseReference(urlStr)
+	url, err = parseURL(urlStr)
 	if err != nil {
 		return nil, "", err
 	}
-	url = &charm.URL{
+	return url, path[i:], nil
+}
+
+func parseURL(urlStr string) (*charm.URL, error) {
```
urosj, line 396: Will this have to change if we agree with core about the proposal about the syntax of charm.URL and charm.ReferenceURL?

internal/router/router.go
rogpeppe: The 7 in the comment is indeed fishy, but we do have tests for nextIndex (phrased to be slightly more readable by testing against path[nextIndex:] rather than nextIndex itself)

internal/router/router.go
rogpeppe, line 396: yup.

internal/router/router.go
urosj, line 143: Ok.

urosj:
LGMT, but would also add content test for bulk meta before merging.

internal/router/router.go
rogpeppe: done

internal/router/router.go
rogpeppe: moved comment to HandleJSON itself and clarified.

internal/router/router.go
rogpeppe: Fixed comment.

rogpeppe:
:shipit:

jujugui:

PR #39, opened 2014-07-29
internal/router: support for efficient include retrieval

rogpeppe:
We drop the ItemGetter logic - it was more indirect
and inefficient than necessary - and introduce the
notion of a BulkIncludeHandler which knows how
to group itself with other like-minded handlers
to make a single shared request.

This has the nice side effect that the RegisterCollection
magic is no longer needed, and the shared request
can be anything we like. The router no longer needs
mgo as a dependency.

### DIFF
--- a/internal/router/fieldinclude.go
+++ b/internal/router/fieldinclude.go
@@ -11,7 +11,7 @@ import (
 type FieldQueryFunc func(id *charm.URL, selector map[string]int) (interface{}, error)
 
 // A FieldHandlerFunc returns some data from the given document. The
-// document will have been returned from an earlier call totp the
+// document will have been returned from an earlier call to the
 // associated QueryFunc.
 type FieldHandlerFunc func(doc interface{}, id *charm.URL, path string, flags url.Values) (interface{}, error)
 
--- a/internal/router/fieldinclude.go
+++ b/internal/router/fieldinclude.go
@@ -1,0 +1,77 @@
+package router
+
+import (
+	"net/url"
+
+	"gopkg.in/juju/charm.v2"
+)
+
+// A FieldQueryFunc is used to retrieve a metadata document for the given URL,
+// selecting only those fields specified in keys of the given selector.
+type FieldQueryFunc func(id *charm.URL, selector map[string]int) (interface{}, error)
+
+// A FieldHandlerFunc returns some data from the given document. The
+// document will have been returned from an earlier call totp the
+// associated QueryFunc.
+type FieldHandlerFunc func(doc interface{}, id *charm.URL, path string, flags url.Values) (interface{}, error)
+
+// FieldIncludeHandler returns a BulkIncludeHandler that will perform
+// only a single database query for several requests. The given key is
+// used to group together similar FieldIncludeHandlers (the same query
+// should be generated for a given key). The given query is used to
+// retrieve the document from the database, and the given handle
+// function is used to actually retrieve the metadata after the query.
+//
+// The fields specify which fields are required by the given handler.
+// The fields passed to the query will be the union of all fields found
+// in all the handlers in the bulk request.
+func FieldIncludeHandler(key interface{}, q FieldQueryFunc, fields []string, handle FieldHandlerFunc) BulkIncludeHandler {
+	return &fieldIncludeHandler{
+		key:    key,
+		query:  q,
+		fields: fields,
+		handle: handle,
+	}
+}
+
+type fieldIncludeHandler struct {
+	key    interface{}
+	query  FieldQueryFunc
+	fields []string
+	handle FieldHandlerFunc
+}
+
+func (h *fieldIncludeHandler) Key() interface{} {
+	return h.key
+}
+
+func (h *fieldIncludeHandler) Handle(hs []BulkIncludeHandler, id *charm.URL, paths []string, flags url.Values) ([]interface{}, error) {
+	funcs := make([]FieldHandlerFunc, len(hs))
+	selector := make(map[string]int)
+	// Extract the handler functions and union all the fields.
+	for i, h := range hs {
+		h := h.(*fieldIncludeHandler)
+		funcs[i] = h.handle
+		for _, field := range h.fields {
+			selector[field] = 1
+		}
+	}
+	// Make the single query.
+	doc, err := h.query(id, selector)
+	if err != nil {
+		return nil, err
+	}
+	// Call all the handlers with the resulting query document
+	results := make([]interface{}, len(hs))
+	for i, f := range funcs {
+		var err error
+		results[i], err = f(doc, id, paths[i], flags)
+		if err != nil {
+			// TODO correlate error with handler (perhaps return
+			// an error that identifies the slice position of the handler that
+			// failed
+			return nil, err
+		}
+	}
+	return results, nil
+}
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -9,13 +9,9 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"reflect"
 	"strings"
-	"sync"
 
 	charm "gopkg.in/juju/charm.v2"
-	"labix.org/v2/mgo"
-	"labix.org/v2/mgo/bson"
 
 	"github.com/juju/charmstore/params"
 )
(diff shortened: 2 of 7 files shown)


jrwren:
Looks good.

internal/v4/api.go
```diff
@@ -79,6 +82,35 @@ func (h *handler) resolveURL(url *charm.URL) error {
 	return ResolveURL(h.store, url)
 }
 
+type entityHandlerFunc func(entity *mongodoc.Entity, id *charm.URL, path string, flags url.Values) (interface{}, error)
+
+// entityHandler returns a handler that calls f with a *mongodoc.Entity that
+// contains at least the given fields.
+func (h *handler) entityHandler(f entityHandlerFunc, fields ...string) router.BulkIncludeHandler {
+	type entityHandlerKey struct{}
+	return router.FieldIncludeHandler(
```
frankban, line 95: I had to look at this for five minutes in order to remember what each argument represents.
Perhaps, something like:
key :=  entityHandlerKey{}
handlerFunc := func(doc interface...)
would help.

urosj:
:+1:

internal/v4/api.go
rogpeppe, line 95: Yes, I sympathise with that. I toyed with the idea of a struct argument here.

I'll factor out the handler function as you suggest - I think it's pretty
obvious that entityHandlerKey is intended to be a key.

internal/router/fieldinclude.go
```diff
@@ -0,0 +1,77 @@
...
+		for _, field := range h.fields {
+			selector[field] = 1
+		}
+	}
+	// Make the single query.
+	doc, err := h.query(id, selector)
+	if err != nil {
+		return nil, err
+	}
+	// Call all the handlers with the resulting query document
```
frankban, line 64: Please complete the sentence.

internal/router/fieldinclude.go
```diff
@@ -0,0 +1,77 @@
...
+// FieldIncludeHandler returns a BulkIncludeHandler that will perform
+// only a single database query for several requests. The given key is
+// used to group together similar FieldIncludeHandlers (the same query
+// should be generated for a given key). The given query is used to
+// retrieve the document from the database, and the given handle
+// function is used to actually retrieve the metadata after the query.
+//
+// The fields specify which fields are required by the given handler.
+// The fields passed to the query will be the union of all fields found
+// in all the handlers in the bulk request.
```
frankban, line 27: This is where a usage example would really help understanding the specific implementation.
Otherwise the logic does not seem to me clear by just looking at the doc (u.e. without looking at the code).
Perhaps it's me, but I was able to understand how to use this and in general what's going on only starting from client code in v4.

internal/router/fieldinclude.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	// Call all the handlers with the resulting query document
+	results := make([]interface{}, len(hs))
+	for i, f := range funcs {
+		var err error
+		results[i], err = f(doc, id, paths[i], flags)
+		if err != nil {
+			// TODO correlate error with handler (perhaps return
+			// an error that identifies the slice position of the handler that
+			// failed
```
frankban, line 72: Please close the parenthesis and complete the sentence.

internal/router/router.go
```diff
@@ -30,12 +26,33 @@ var knownSeries = map[string]bool{
...
+	//
+	// All of the handlers' Keys will be equal to the receiving handler's
+	// Key.
+	//
+	// Each item in paths holds the remaining metadata path
+	// for the handler in the corresponding position
+	// in hs after the prefix in Handlers.Meta has been stripped,
+	// and flags holds all the url query values.
+	//
+	// TODO(rog) document indexed errors
```
frankban, line 53: Complete the sentence here too.

internal/router/router.go
```diff
@@ -230,124 +249,43 @@ func (r *Router) serveBulkMeta(w http.ResponseWriter, req *http.Request) (interf
...
-	collections     = make(map[reflect.Type]string)
-)
-
-// RegisterCollection registers the type of docType as document type for
-// a given collection, so documents of that type can be retrieved using
-// ItemGetter's GetItem method.
-//
-// The document type must be a pointer to a struct or map type suitable
-// for unmarshalling with mgo/bson.
-func RegisterCollection(collectionName string, docType interface{}) {
```
frankban, line 244: Nice to see this deleted.

internal/router/router.go
```diff
@@ -230,124 +249,43 @@ func (r *Router) serveBulkMeta(w http.ResponseWriter, req *http.Request) (interf
...
 func (r *Router) GetMetadata(id *charm.URL, includes []string) (map[string]interface{}, error) {
-	if len(includes) == 0 {
-		return nil, nil
-	}
-	// TODO(rog) optimize this to avoid making a db round trip
-	// for every piece of metadata and to cache previously
-	// fetched results.
-	getter := getterFunc(r.getter)
-	results := make(map[string]interface{})
+	groups := make(map[interface{}][]BulkIncludeHandler)
```
frankban, line 255: This code here is nice but very dense.
It would be nice to add some comments, e.g. about the difference between the handler key and the include keys, and why we use g[0] for each group to collect all the results.

internal/router/router.go
```diff
@@ -230,124 +249,43 @@ func (r *Router) serveBulkMeta(w http.ResponseWriter, req *http.Request) (interf
...
+			_, paths[i] = handlerKey(include)
+		}
+		groupResults, err := g[0].Handle(g, id, paths, nil)
 		if err != nil {
-			// TODO(rog) If error is "inappropriate metadata", then
-			// just omit the result.
-			return nil, fmt.Errorf("cannot get meta/%q: %v", include, err)
+			// TODO(rog) if it's a BulkError, attach
+			// the original include path to error (the BulkError
+			// should contain the index of the failed one)
```
frankban, line 278: Please complete the sentence.

internal/router/router.go
```diff
@@ -230,124 +249,43 @@ func (r *Router) serveBulkMeta(w http.ResponseWriter, req *http.Request) (interf
...
-	collection, err := collectionForValue(val)
-	if err != nil {
-		return err
-	}
-	selector := make(bson.D, len(fields))
-	for i, field := range fields {
-		selector[i] = bson.DocElem{
-			Name:  field,
-			Value: 1,
+		for i, result := range groupResults {
```
frankban, line 281: Nice.

internal/router/router_test.go
```diff
@@ -446,16 +461,16 @@ func noResolveURL(*charm.URL) error {
...
-
 	for i, test := range routerTests {
 		c.Logf("test %d: %s", i, test.about)
 		resolve := noResolveURL
 		if test.resolveURL != nil {
 			resolve = test.resolveURL
 		}
-		router := New(db, &test.handlers, resolve)
+		router := New(&test.handlers, resolve)
+		queryCount = 0
```
frankban, line 471: What do you think about a comment here explaining that queryCount is atomically increased by the handlers (fieldSelectHandler)?

internal/router/singleinclude.go
```diff
@@ -0,0 +1,34 @@
...
+
+var _ BulkIncludeHandler = SingleIncludeHandler(nil)
+
+// SingleIncludeHandler implements BulkMetaHander for a non-batching
+// metadata retrieval function.
+type SingleIncludeHandler func(id *charm.URL, path string, flags url.Values) (interface{}, error)
+
+// Key implements BulkMetadataHander.Key
+func (h SingleIncludeHandler) Key() interface{} {
+	type singleMetaHandlerKey int
```
frankban, line 17: So, this key will be different for each instance right? If so, I'd add a comment to clarify the intent.

internal/router/singleinclude.go
```diff
@@ -0,0 +1,34 @@
...
+	return singleMetaHandlerKey(0)
+}
+
+// Handle implements BulkMetadataHander.Handle
```
frankban, line 21: Final dot.

internal/router/singleinclude.go
```diff
@@ -0,0 +1,34 @@
...
+func (h SingleIncludeHandler) Handle(hs []BulkIncludeHandler, id *charm.URL, paths []string, flags url.Values) ([]interface{}, error) {
+	results := make([]interface{}, len(hs))
+	for i, h := range hs {
+		h := h.(SingleIncludeHandler)
+		result, err := h(id, paths[i], flags)
+		if err != nil {
+			// TODO(rog) include index of failed handler
```
frankban, line 28: Please complete the sentence.

frankban:
:+1: with some comments, FWIW.

Most of the comments are just requests for comments, and that's because I have the impression that, while the resulting logic is really nice from the API implemeter perspective (e.g. `h.entityHandler(h.metaCharmMetadata, "charmmeta"),`), this branch introduces more inversion of control, that can be confusing without documenting the framework accurately.

Thank you!

rogpeppe:
:shipit:

jujugui:

PR #45, opened 2014-07-30
internal/v4: refactor tests

rogpeppe:
We create a table with an entry for each kind of metadata so
that it's easy to add new entries and new tests that
test the metadata in arbitrary configurations without
writing reams of duplicated test code.

We also fix router so that it works correctly with
typed nil data, which avoids an if statement in
every metadata handler.

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"reflect"
 	"sort"
 	"strings"
 
@@ -210,12 +211,25 @@ func (r *Router) serveMeta(id *charm.URL, req *http.Request) (interface{}, error
 			return nil, err
 		}
 		result := results[0]
-		if result == nil {
+		if isNull(result) {
 			return nil, ErrDataNotFound
 		}
 		return results[0], nil
 	}
 	return nil, ErrNotFound
+}
+
+// isNull reports whether the given value will encode to
+// a null JSON value.
+func isNull(val interface{}) bool {
+	if val == nil {
+		return true
+	}
+	v := reflect.ValueOf(val)
+	if kind := v.Kind(); kind != reflect.Map && kind != reflect.Ptr && kind != reflect.Slice {
+		return false
+	}
+	return v.IsNil()
 }
 
 func (r *Router) metaNames() []string {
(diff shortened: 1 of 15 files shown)


jujugui:
please test this

bac:
ok to test

internal/v4/api_test.go
```diff
@@ -40,49 +46,123 @@ func (s *APISuite) SetUpTest(c *gc.C) {
...
-	c.Assert(err, gc.IsNil)
-	return url, wordpress
+type metaEndpoint struct {
+	name       string
+	exclusive  int
+	bundleOnly bool
+	get        func(*charmstore.Store, *charm.URL) (interface{}, error)
 }
 
-func (s *APISuite) addBundle(c *gc.C, bundleName string, curl string) (*charm.URL, charm.Bundle) {
```
urosj, line 52: Note: AddCharm and AddBundle are not really removed, just "moved".

rogpeppe:
ok to test

urosj:
:+1:

internal/v4/api_test.go
```diff
@@ -40,49 +46,123 @@ func (s *APISuite) SetUpTest(c *gc.C) {
...
-}
+const (
+	charmOnly = iota + 1
+	bundleOnly
+)
+
+var metaEndpoints = []metaEndpoint{{
+	name:      "charm-config",
+	exclusive: charmOnly,
+	get:       entityFieldGetter("CharmConfig"),
```
frankban, line 70: I wonder if, in order to make the cases more explicit, we can just put the returned value here without fetching it dynamically with entityFieldGetter/entityGetter. Otherwise it seems those functions are non-trivial enough to deserve their own tests. What do you think?

internal/router/router.go
```diff
@@ -218,6 +219,19 @@ func (r *Router) serveMeta(id *charm.URL, req *http.Request) (interface{}, error
 	return nil, ErrNotFound
 }
 
+// isNull reports whether the given value will encode to
+// a null JSON value.
+func isNull(val interface{}) bool {
```
frankban, line 224: Nice!

frankban:
:+1: thank you!

internal/v4/api_test.go
urosj, line 70: I was thinking the same, but then concluded that this way is actually more descriptive as we "see" the mapping function.

rogpeppe, line 70: We can't just put the value here, because it varies according to the charm URL.

I can add some tests for the various endpoint get functions, but we do at least
currently check that they return non-nil (avoiding the most obvious pitfall).

I would prefer to avoid hard-coding the charm metadata from the charm/testing/repo
charms in this file.

jrwren:
lgtm

bac:
ok to test

frankban:
:+1: thanks

rogpeppe:
:shipit:

jujugui:

PR #50, opened 2014-07-31
charmstore/router: generate http.StatusNotFound appropriately

rogpeppe:
This generates messages (and error codes) as expected by the API specification.


jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/41/

jrwren:
:+1:

internal/router/router.go
```diff
@@ -262,7 +258,7 @@ func (r *Router) serveBulkMeta(w http.ResponseWriter, req *http.Request) (interf
 			return nil, errgo.Mask(err)
```
urosj, line 277: why not errgo.Mask(err, errgo.Any)?

internal/router/router.go
```diff
@@ -209,15 +205,15 @@ func (r *Router) serveMeta(id *charm.URL, req *http.Request) (interface{}, error
 	if handler := r.handlers.Meta[key]; handler != nil {
 		results, err := handler.Handle([]BulkIncludeHandler{handler}, id, []string{path}, req.Form)
 		if err != nil {
-			return nil, errgo.Mask(err)
+			return nil, errgo.Mask(err, errgo.Any)
 		}
 		result := results[0]
 		if isNull(result) {
-			return nil, ErrDataNotFound
+			return nil, params.ErrMetadataNotFound
```
urosj, line 231: Nice! Let's make errors descriptive - good point.

internal/router/router_test.go
```diff
@@ -230,8 +263,9 @@ var routerTests = []struct {
 		},
 	},
 	urlStr:     "http://example.com/precise/wordpress-42/meta/foo",
-	expectCode: http.StatusInternalServerError,
+	expectCode: http.StatusNotFound,
 	expectBody: params.Error{
+		Code:    params.ErrMetadataNotFound,
```
urosj, line 282: This tests are getting harder to read (might be just the hour of day), We have two same urlStrings, but different codes. The test with the constMetaHandler(nil) returns ErrMetadataNotFound is clear, however, it is not "obvious on first sight" why previous test returns ErrNotFound instead (as no handler is provided).

internal/router/router_test.go
```diff
@@ -138,6 +140,36 @@ var routerTests = []struct {
...
+	about: "id handler that returns some other kind of coded error",
+	handlers: Handlers{
+		Id: map[string]IdHandler{
+			"foo": func(charmId *charm.URL, w http.ResponseWriter, req *http.Request) error {
+				return errgo.WithCausef(nil, params.ErrorCode("foo"), "a message")
+			},
+		},
+	},
+	urlStr:     "http://example.com/~joe/precise/wordpress-34/foo",
+	expectCode: http.StatusInternalServerError,
```
urosj, line 167: So the only time now that we return StatusInternalServerError is when errgo.Cause(err) occurs, otherwise it's *NotFound err? 
:+1:

internal/v4/api.go
```diff
@@ -105,6 +107,9 @@ func (h *handler) entityQuery(id *charm.URL, selector map[string]int) (interface
 		Find(bson.D{{"_id", id}}).
 		Select(selector).
 		One(&val)
+	if err == mgo.ErrNotFound {
+		return nil, params.ErrNotFound
+	}
 	if err != nil {
 		return nil, errgo.Mask(err)
```
urosj, line 115: errgo.Mask(err, errgo.Any)?

internal/router/router.go
```diff
@@ -151,19 +146,20 @@ func (r *Router) serveIds(w http.ResponseWriter, req *http.Request) error {
 	}
```
urosj, line 161: there are three errgo.Mask(err) calls in lines before the comment ... should they be errgo.Mask(err, errgo.Any)?

rogpeppe, line 161: The first two, I don't think so. The third, yes, you're right - we should allow the plugged-in resolveURL function to set arbitrary status codes. Will change and add a test.

internal/router/router.go
rogpeppe, line 277: Because we don't want parseURL setting status codes. By omitting errgo.Any, we can see at a glance which are the possible places that could be returning an error with a Cause (the only kind of error that can end up setting an HTTP status code)

internal/router/router_test.go
rogpeppe, line 167: Not quite. Any error at all will generate StatusInternalServerError unless it has a Cause of params.ErrNotFound or params.ErrMetadataNotFound. We will probably add more http status codes in the future.

internal/router/router_test.go
rogpeppe, line 282: It's precisely that distinction we are currently making - between "not found" (you're trying to address something that's genuinely not there) and "metadata not found" (we know what you're trying to address, but it has no content). The former would probably generate an error in a client; the latter is expected and probably would not.

I will add comments.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/42/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/43/

urosj:
:+1: Thanks.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/47/

jrwren:
still LGTM

internal/router/router.go
```diff
@@ -18,6 +18,13 @@ import (
 	"github.com/juju/charmstore/params"
 )
 
+// Implementation node on error handling:
```
frankban: Implementation _note_

internal/router/util.go
```diff
@@ -45,16 +45,19 @@ func HandleJSON(handle func(http.ResponseWriter, *http.Request) (interface{}, er
...
 // HTTP status.
 func WriteError(w http.ResponseWriter, err error) {
-	// TODO(rog) determine appropriate http status
-	// from error rather than always using 500.
 	errResp := &params.Error{
 		Message: err.Error(),
 	}
-	if err, ok := errgo.Cause(err).(params.ErrorCoder); ok {
-		errResp.Code = err.ErrorCode()
+	if err, ok := errgo.Cause(err).(params.ErrorCode); ok {
```
frankban, line 51: Cool.

internal/router/util.go
```diff
@@ -45,16 +45,19 @@ func HandleJSON(handle func(http.ResponseWriter, *http.Request) (interface{}, er
...
+		errResp.Code = err
+	}
+	status := http.StatusInternalServerError
+	switch errResp.Code {
+	case params.ErrNotFound, params.ErrMetadataNotFound:
+		status = http.StatusNotFound
```
frankban, line 57: So in the future we can have something like:
case params.ErrForbidden:
    status = http.StatusForbidden
For when we need to return a 403 e.g. for unauthenticated requests.
Correct?

params/error.go
```diff
@@ -0,0 +1,53 @@
+package params
+
+import (
+	"fmt"
+)
+
+// ErrorCode holds the class of an error in machine-readable format.
+type ErrorCode string
```
frankban, line 8: Just curious: why defining ErrorCode from a string and not just using an error (or a errgo.Err)?

frankban:
:+1: with some comments/questions.
Thank you!

internal/router/util.go
rogpeppe, line 57: Yup.

params/error.go
rogpeppe, line 8: Because these codes need to travel across the network. If they were just errors, they wouldn't be comparable for equality after unmarshaling.

rogpeppe:
:shipit:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/48/

jujugui:

PR #52, opened 2014-08-04
internal/v4: implement stats endpoint

rogpeppe:
internal/v4/stats.go and internal/v4/stats_test.go are directly derived from
code in _old, although simplified to return only JSON-formatted results.

This required a fair number of fixes to other code, which
unfortunately aren't that easy to pick apart now, including the
following:
- internal/charmstore: make stats counter tests less hacky.

Because the internal/v4 tests are outside charmstore, delving
directly into the database to tweak the time stamps seemed unwarranted,
so added IncCounterAtTime and changed internal/v4 tests and existing
internal/charmstore tests to use it.
- internal/charmstore: move mongojs condition into storetesting.

This avoids duplicating the logic to determine whether tests requiring
mongodb embedded javascript can be run.
- internal/router: universal ParseForm.

This clears up responsibility for who should call ParseForm.
- internal/router: strip prefix from global handlers

This makes the global handlers (including /stats/counters) consistent
with other handlers in router.Handlers.
- internal/router: add ServeMux, NotFoundHandler

The original /stats/counter tests tested http status of paths which previously
returned an http.StatusNotFound status but with a non-JSON body.
The above two handlers make the not-found pages consistent
with the rest of the API.
- params: add ErrForbidden, ErrBadRequest

The original /stats/counter tests tested for statuses which our
current code was incapable of generating. The above two
errors make it capable.

### DIFF
--- a/internal/charmstore/stats.go
+++ b/internal/charmstore/stats.go
@@ -17,6 +17,15 @@ import (
 	"github.com/juju/charmstore/params"
 )
 
+// Note that changing the StatsGranularity constant
+// will not change the stats time granularity - it
+// is defined for external code clarity.
+
+// StatsGranularity holds the time granularity of statistics
+// gathering. IncCounter calls within this duration
+// may be aggregated.
+const StatsGranularity = time.Minute
+
 // The stats mechanism uses the following MongoDB collections:
 //
 //     juju.stat.counters - Counters for statistics
@@ -148,6 +157,14 @@ func timeToStamp(t time.Time) int32 {
 
 // IncCounter increases by one the counter associated with the composed key.
 func (s *Store) IncCounter(key []string) error {
+	return s.IncCounterAtTime(key, time.Now())
+}
+
+// IncCounter increases by one the counter associated with the composed key,
+// associating it with the given time, which should be time.Now.
+// This method is exposed for testing purposes only - production
+// code should always call IncCounter.
+func (s *Store) IncCounterAtTime(key []string, t time.Time) error {
 	db := s.DB.Copy()
 	defer db.Close()
 
@@ -156,9 +173,8 @@ func (s *Store) IncCounter(key []string) error {
 		return err
 	}
 
-	t := time.Now().UTC()
 	// Round to the start of the minute so we get one document per minute at most.
-	t = t.Add(-time.Duration(t.Second()) * time.Second)
+	t = t.UTC().Add(-time.Duration(t.Second()) * time.Second)
 	counters := db.StatCounters()
 	_, err = counters.Upsert(bson.D{{"k", skey}, {"t", timeToStamp(t)}}, bson.D{{"$inc", bson.D{{"c", 1}}}})
 	return err
@@ -238,7 +254,11 @@ func (s *Store) Counters(req *CounterRequest) ([]Counter, error) {
 	searchKey, err := s.statsKey(db, req.Key, false)
 	if errgo.Cause(err) == params.ErrNotFound {
 		if !req.List {
-			return []Counter{{Key: req.Key, Prefix: req.Prefix, Count: 0}}, nil
+			return []Counter{{
+				Key:    req.Key,
+				Prefix: req.Prefix,
+				Count:  0,
+			}}, nil
 		}
 		return nil, nil
 	}
--- a/internal/charmstore/stats_test.go
+++ b/internal/charmstore/stats_test.go
@@ -4,14 +4,11 @@
 package charmstore_test
 
 import (
-	"flag"
 	"fmt"
-	"os"
 	"strconv"
 	"sync"
 	"time"
 
-	jujutesting "github.com/juju/testing"
 	"gopkg.in/mgo.v2/bson"
 	gc "launchpad.net/gocheck"
 
@@ -24,21 +21,7 @@ type StatsSuite struct {
 	store *charmstore.Store
 }
 
-var noTestMongoJs *bool = flag.Bool("notest-mongojs", false, "Disable MongoDB tests that require javascript")
-
 var _ = gc.Suite(&StatsSuite{})
-
-func (s *StatsSuite) SetUpSuite(c *gc.C) {
-	s.IsolatedMgoSuite.SetUpSuite(c)
-	if os.Getenv("JUJU_NOTEST_MONGOJS") == "1" || jujutesting.MgoServer.WithoutV8 {
-		c.Log("Tests requiring MongoDB Javascript will be skipped")
-		*noTestMongoJs = true
-	}
-}
-
-func (s *StatsSuite) TearDownSuite(c *gc.C) {
-	s.IsolatedMgoSuite.TearDownSuite(c)
-}
 
 func (s *StatsSuite) SetUpTest(c *gc.C) {
 	s.IsolatedMgoSuite.SetUpTest(c)
@@ -52,7 +35,7 @@ func (s *StatsSuite) TearDownTest(c *gc.C) {
 }
 
 func (s *StatsSuite) TestSumCounters(c *gc.C) {
-	if *noTestMongoJs {
+	if !storetesting.MongoJSEnabled() {
 		c.Skip("MongoDB javascript not available")
 	}
 
@@ -129,7 +112,7 @@ func (s *StatsSuite) TestSumCounters(c *gc.C) {
 }
 
 func (s *StatsSuite) TestCountersReadOnlySum(c *gc.C) {
-	if *noTestMongoJs {
+	if !storetesting.MongoJSEnabled() {
 		c.Skip("MongoDB javascript not available")
 	}
 
@@ -145,7 +128,7 @@ func (s *StatsSuite) TestCountersReadOnlySum(c *gc.C) {
 }
 
 func (s *StatsSuite) TestCountersTokenCaching(c *gc.C) {
-	if *noTestMongoJs {
+	if !storetesting.MongoJSEnabled() {
 		c.Skip("MongoDB javascript not available")
 	}
 
@@ -204,7 +187,7 @@ func (s *StatsSuite) TestCountersTokenCaching(c *gc.C) {
 }
 
 func (s *StatsSuite) TestCounterTokenUniqueness(c *gc.C) {
-	if *noTestMongoJs {
+	if !storetesting.MongoJSEnabled() {
 		c.Skip("MongoDB javascript not available")
 	}
 
@@ -229,7 +212,7 @@ func (s *StatsSuite) TestCounterTokenUniqueness(c *gc.C) {
 }
 
 func (s *StatsSuite) TestListCounters(c *gc.C) {
-	if *noTestMongoJs {
+	if !storetesting.MongoJSEnabled() {
 		c.Skip("MongoDB javascript not available")
 	}
 
@@ -293,7 +276,7 @@ func (s *StatsSuite) TestListCounters(c *gc.C) {
 }
 
 func (s *StatsSuite) TestListCountersBy(c *gc.C) {
-	if *noTestMongoJs {
+	if !storetesting.MongoJSEnabled() {
 		c.Skip("MongoDB javascript not available")
 	}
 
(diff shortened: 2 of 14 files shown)


internal/router/router_test.go
```diff
@@ -44,13 +44,46 @@ var routerTests = []struct {
...
-				return &Foo{"hello"}, nil
+				return ReqInfo{
+					Path: req.URL.Path,
+					Form: req.Form,
+				}, nil
 			}),
 		},
 	},
-	urlStr:     "http://example.com/foo",
+	urlStr:     "http://0.1.2.3/foo",
```
jrwren, line 54: consider using a valid test or documentation IP address range. 198.18/15 is reserved for test. 192.0.2.0/24 is reserved for documentation. See rfc5737 and rfc2544 for more.

rogpeppe, line 54: Unfortunately, most machines' router tables do not special-case either of the above two ranges, so they don't help to guard against accidental network access from tests. 0.1.2.3, on the other hand is (at least on Linux) picked up immediately by the network stack as an invalid IP address, and generates an error without touching the network.

That's why we use 0.1.2.3 (or similar) as a testing IP address in juju-core too.

jrwren:
LGTM

internal/charmstore/stats_test.go
```diff
@@ -319,17 +302,14 @@ func (s *StatsSuite) TestListCountersBy(c *gc.C) {
...
 
-		// Hack time so counters are assigned to 2012-05-<day>
-		filter := bson.M{"t": bson.M{"$gt": charmstore.TimeToStamp(time.Date(2013, time.January, 1, 0, 0, 0, 0, time.UTC))}}
-		stamp := charmstore.TimeToStamp(day(inc.day))
-		stamp += int32(i) * 60 // Make every entry unique.
-		err = counters.Update(filter, bson.D{{"$set", bson.D{{"t", stamp}}}})
-		c.Check(err, gc.IsNil)
+		err := s.store.IncCounterAtTime(inc.key, t)
+		c.Assert(err, gc.IsNil)
 	}
```
urosj, line 313: Getting rid of hardcoded stuff, nice ;)

internal/charmstore/stats.go
```diff
@@ -148,6 +157,14 @@ func timeToStamp(t time.Time) int32 {
 
 // IncCounter increases by one the counter associated with the composed key.
 func (s *Store) IncCounter(key []string) error {
+	return s.IncCounterAtTime(key, time.Now())
+}
+
+// IncCounter increases by one the counter associated with the composed key,
```
frankban, line 163: // IncCounterAtTime increases...

internal/charmstore/stats.go
```diff
@@ -17,6 +17,15 @@ import (
 	"github.com/juju/charmstore/params"
 )
 
+// Note that changing the StatsGranularity constant
```
frankban, line 20: It seems to me more natural to have this note after the actual docstring below.

rogpeppe, line 20: I didn't want it to show up in the godoc output, as it's an implementation detail.
We could put it after the constant if you'd find that better.

internal/router/router.go
```diff
@@ -118,10 +125,15 @@ func New(handlers *Handlers, resolveURL func(url *charm.Reference) error) *Route
...
 		resolveURL: resolveURL,
 	}
-	mux := http.NewServeMux()
+	mux := NewServeMux()
 	mux.Handle("/meta/", http.StripPrefix("/meta", HandleJSON(r.serveBulkMeta)))
 	for path, handler := range r.handlers.Global {
-		mux.Handle("/"+path, handler)
+		path = "/" + path
+		prefix := path
+		if strings.HasSuffix(prefix, "/") {
```
frankban, line 133: If I am not missing something, we can replace this condition with just `prefix = strings.TrimRight(prefix, "/")`.

internal/router/router.go
```diff
@@ -118,10 +125,15 @@ func New(handlers *Handlers, resolveURL func(url *charm.Reference) error) *Route
...
+			prefix = prefix[0 : len(prefix)-1]
+		}
+		mux.Handle(path, http.StripPrefix(prefix, handler))
```
frankban, line 136: Ok, we remove the prefix here.

internal/router/util.go
```diff
@@ -78,3 +82,34 @@ func WriteJSON(w http.ResponseWriter, code int, val interface{}) error {
...
+type ServeMux struct {
+	*http.ServeMux
+}
+
+func (mux *ServeMux) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if req.RequestURI == "*" {
+		mux.ServeMux.ServeHTTP(w, req)
+		return
+	}
+	h, pat := mux.Handler(req)
```
frankban: Can we call this "pattern"? "pat" in this context can be confusing IMHO.

internal/storetesting/flag.go
```diff
@@ -0,0 +1,22 @@
+package storetesting
```
frankban, line 4: Missing file headers.

internal/storetesting/flag.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"flag"
+	"os"
+
+	jujutesting "github.com/juju/testing"
+)
+
+var noTestMongoJs *bool = flag.Bool("notest-mongojs", false, "Disable MongoDB tests that require javascript")
```
frankban: "Disable MongoDB tests that require JavaScript"

internal/storetesting/flag.go
```diff
@@ -0,0 +1,22 @@
...
+
+func init() {
+	if os.Getenv("JUJU_NOTEST_MONGOJS") == "1" || jujutesting.MgoServer.WithoutV8 {
+		*noTestMongoJs = true
+	}
+}
+
+// MongoJSEnabled reports whether testing code should run tests
+// that rely on Javascript inside MongoDB.
```
frankban: JavaScript [shrug]

internal/storetesting/http.go
```diff
@@ -44,7 +44,7 @@ func AssertJSONCall(
 
 	var gotBodyVal interface{}
 	err = json.Unmarshal(rec.Body.Bytes(), &gotBodyVal)
-	c.Assert(err, gc.IsNil)
+	c.Assert(err, gc.IsNil, gc.Commentf("json body: %q", rec.Body.Bytes()))
```
frankban, line 47: Didn't know you can attach comments to tests. Nice.

internal/v4/api.go
```diff
@@ -30,7 +30,8 @@ func New(store *charmstore.Store) http.Handler {
 	}
 	h.Router = router.New(&router.Handlers{
 		Global: map[string]http.Handler{
-			"stats/counter":      http.HandlerFunc(h.serveStatsCounter),
+			"stats/counter/":     router.HandleJSON(h.serveStatsCounter),
+			"stats/":             router.NotFoundHandler(),
```
frankban, line 34: Do we need an explicit NotFoundHandler for "stats/"?
I supposed that if there's no match then a 404 is returned automatically.

internal/v4/api.go
```diff
@@ -151,12 +156,6 @@ func preferredURL(url0, url1 *charm.Reference) bool {
 
 var errNotImplemented = errgo.Newf("method not implemented")
 
-// GET stats/counter/key[:key]...?[by=unit]&start=date][&stop=date][&list=1]
```
frankban, line 154: Good to see the stats logic moved into its own file.

internal/v4/api_test.go
```diff
@@ -38,13 +39,20 @@ var _ = gc.Suite(&APISuite{})
...
 	store, err := charmstore.NewStore(db)
 	c.Assert(err, gc.IsNil)
-	s.store = store
 	srv, err := charmstore.NewServer(db, map[string]charmstore.NewAPIHandler{"v4": v4.New})
 	c.Assert(err, gc.IsNil)
-	s.srv = srv
+	return srv, store
+}
+
+func storeURL(path string) string {
```
frankban, line 54: getStoreURL or makeStoreURL? I usually prefer functions to be verbs: is this a common practice also in the Go world?

internal/v4/stats_test.go
```diff
@@ -0,0 +1,422 @@
+// Copyright 2012 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
```
frankban: This should be LGPLv3.

internal/v4/api.go
rogpeppe, line 34: If there's no match, we fall back to trying to parse "stats" as a charm id.

internal/v4/api_test.go
rogpeppe, line 54: It's not uncommon, particularly for pure functions. I think of it a bit like a mathematical function, say sine or log. You wouldn't want to say math.NewSin(x) rather than math.Sin(x). The fact that it's statically typed should make it unambiguous, and I think it reads ok.

internal/v4/stats.go
```diff
@@ -0,0 +1,98 @@
...
+	var by charmstore.CounterRequestBy
+	switch v := r.Form.Get("by"); v {
+	case "":
+		by = charmstore.ByAll
+	case "day":
+		by = charmstore.ByDay
+	case "week":
+		by = charmstore.ByWeek
+	default:
+		return nil, errgo.Newf("invalid 'by' value: %q", v)
```
frankban: I am not sure what happens here, but I suppose a 500 is returned. Shouldn't this return a BadRequest instead?

internal/v4/stats.go
```diff
@@ -0,0 +1,98 @@
...
+	req := charmstore.CounterRequest{
+		Key:  strings.Split(base, ":"),
+		List: r.Form.Get("list") == "1",
+		By:   by,
+	}
+	if v := r.Form.Get("start"); v != "" {
+		var err error
+		req.Start, err = time.Parse("2006-01-02", v)
+		if err != nil {
+			return nil, errgo.Newf("invalid 'start' value: %q", v)
```
frankban: Ditto.

internal/v4/stats_test.go
```diff
@@ -0,0 +1,422 @@
...
+	expected := map[string]int64{
+		"a:b":   2,
+		"a:b:*": 0,
+		"a:*":   3,
+		"a":     1,
+		"a:b:c": 0,
+	}
+
+	for counter, n := range expected {
+		c.Logf("test %q", counter)
```
frankban, line 138: We could also log the expected key here. [shrug]

internal/v4/stats_test.go
```diff
@@ -0,0 +1,422 @@
...
+	}}
+
+	for i, test := range tests {
+		c.Logf("test %d: %s", i, test.key)
+		url := storeURL("stats/counter/" + test.key + "?list=1")
+		storetesting.AssertJSONCall(c, s.srv, "GET", url, "", http.StatusOK, test.result)
+	}
+}
+
+// counterEpoch is a copy of charmstore.counterEpoch.
```
frankban: Perhaps it's worth of adding those to an export_test? If not, when do you think it's worth of doing that, and when it's better to just copy the definitions?

internal/v4/stats_test.go
```diff
@@ -0,0 +1,422 @@
...
+}
+
+func (s *StatsSuite) TestServerStatsStatus(c *gc.C) {
+	tests := []struct {
+		path    string
+		status  int
+		message string
+		code    params.ErrorCode
+	}{{
+		path:    "stats/counter/",
```
frankban, line 59: ... and here I'd add also the BadRequest cases.

frankban:
:+1: with a few comments.
Thank you!

internal/router/util.go
rogpeppe: done.

internal/storetesting/flag.go
rogpeppe, line 4: done.

internal/storetesting/flag.go
rogpeppe: done (and other occurrences of "javascript" too)

internal/v4/api.go
frankban, line 34: OIC, good then.

internal/storetesting/flag.go
rogpeppe: done.

internal/v4/stats.go
rogpeppe: done.

internal/v4/stats_test.go
rogpeppe: done.

internal/v4/stats_test.go
rogpeppe, line 59: done.

internal/v4/stats_test.go
rogpeppe, line 138: I think the counter is sufficient identification. If the test fails, it will print the expected value too.

internal/v4/stats_test.go
rogpeppe: Good catch. They're hangovers from when the tests used internal details. Deleted.

rogpeppe:
:shipit:

jujugui:

internal/charmstore/stats.go
rogpeppe, line 163: Indeed. Missed this before shipit. Will fix in later branch.

internal/router/router.go
rogpeppe, line 133: Yes. Missed this before shipit. Adding to later branch.

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/55/

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/59/

jujugui:
Build failed: Attempt to land pull request failed
        build url: http://ci.jujugui.org:8080/job/charmstore-merge/42

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/62/

PR #56, opened 2014-08-06
internal/router: do not always resolve ids

rogpeppe:
When there's a PUT or a POST, we might actually be
creating a new entity, so defer id resolution
in that case so the specific handler can
deal with it.

For POSTs and PUTs to metadata, the id must already
exist, because otherwise there could not be
metadata about it, so we always resolve in that
case.

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -167,15 +167,22 @@ func (r *Router) serveIds(w http.ResponseWriter, req *http.Request) error {
 	if err != nil {
 		return errgo.Mask(err)
 	}
-	if err := r.resolveURL(url); err != nil {
-		// Note: preserve error cause from resolveURL.
-		return errgo.Mask(err, errgo.Any)
-	}
 	key, path := handlerKey(path)
 	if key == "" {
 		return params.ErrNotFound
 	}
-	if handler, ok := r.handlers.Id[key]; ok {
+	handler := r.handlers.Id[key]
+	if handler == nil || idHandlerNeedsResolveURL(req) {
+		// If it's not an id handler, it's a meta endpoint, so
+		// we always want a resolved URL. Otherwise we leave the
+		// URL unresolved for cases where the id may validly not
+		// exist (for example when uploading a new charm).
+		if err := r.resolveURL(url); err != nil {
+			// Note: preserve error cause from resolveURL.
+			return errgo.Mask(err, errgo.Any)
+		}
+	}
+	if handler != nil {
 		req.URL.Path = path
 		err := handler(url, w, req)
 		// Note: preserve error cause from handlers.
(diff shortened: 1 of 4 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/67/

internal/v4/api_test.go
```diff
@@ -281,15 +281,14 @@ func (s *APISuite) TestBulkMeta(c *gc.C) {
...
-			},
-		})
+	storetesting.AssertJSONCall(c, storetesting.JSONCallParams{
+		Handler: s.srv,
+		URL:     storeURL("meta/charm-metadata?id=precise/wordpress-23&id=precise/mysql-10"),
+		ExpectBody: map[string]*charm.Meta{
+			"precise/wordpress-23": wordpress.Meta(),
+			"precise/mysql-10":     mysql.Meta(),
+		},
+	})
```
urosj, line 291: Why this change if the patch deals with get/put resolve issues?

internal/v4/api_test.go
```diff
@@ -299,27 +298,26 @@ func (s *APISuite) TestBulkMetaAny(c *gc.C) {
...
-			Handler: s.srv,
-			URL:     storeURL("meta/any?include=charm-metadata&include=charm-config&id=precise/wordpress-23&id=precise/mysql-10"),
-			ExpectBody: map[string]params.MetaAnyResponse{
-				"precise/wordpress-23": {
-					Id: wordpressURL,
-					Meta: map[string]interface{}{
-						"charm-config":   wordpress.Config(),
-						"charm-metadata": wordpress.Meta(),
-					},
+	storetesting.AssertJSONCall(c, storetesting.JSONCallParams{
```
urosj, line 301: Same here. Even if it's just formatting, I would suggest that next time we do not make the changes that are not directly related to the suggested PR. And create another PR for them.

internal/router/router.go
```diff
@@ -167,15 +167,22 @@ func (r *Router) serveIds(w http.ResponseWriter, req *http.Request) error {
...
-	if handler, ok := r.handlers.Id[key]; ok {
+	handler := r.handlers.Id[key]
+	if handler == nil || idHandlerNeedsResolveURL(req) {
+		// If it's not an id handler, it's a meta endpoint, so
+		// we always want a resolved URL. Otherwise we leave the
+		// URL unresolved for cases where the id may validly not
+		// exist (for example when uploading a new charm).
+		if err := r.resolveURL(url); err != nil {
+			// Note: preserve error cause from resolveURL.
+			return errgo.Mask(err, errgo.Any)
```
urosj, line 182: What do we do with it in the API?

internal/v4/api_test.go
rogpeppe, line 291: It's a trivial formatting change, a relic of a previous branch.

internal/v4/api_test.go
rogpeppe, line 301: OK, will do.

jrwren:
LGTM

rogpeppe:
```
[10:59:46] <urulama> rogpeppe1: LGTM, but go through comments, please
```

rogpeppe:
:shipit:

jujugui:

PR #62, opened 2014-08-07
internal/v4: implement archive uploading

rogpeppe:
We don't yet verify bundles against charms in the store.

This required some refactoring of the testing code;
the most important change is AssertJSONCall actually
makes a real HTTP call through a socket, as that
allows us to correctly test the ContentLength logic.
It also gives us greater assurance that the logic
will work over a real HTTP connection. The overhead
seems to be negligible.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -82,10 +82,11 @@ func (s *Store) putArchive(archive blobstore.ReadSeekCloser) (string, int64, err
 	return blobHash, size, nil
 }
 
-// UploadCharm uploads the given charm to the blob store.
-// This method is provided for tests only. Normal clients
-// should upload charms through the API.
-func (s *Store) UploadCharm(url *charm.Reference, ch charm.Charm) error {
+// AddCharmWithArchive is like AddCharm but
+// also adds the charm archive to the blob store.
+// This method is provided principally so that
+// tests can easily create content in the store.
+func (s *Store) AddCharmWithArchive(url *charm.Reference, ch charm.Charm) error {
 	blobHash, size, err := s.uploadCharmOrBundle(ch)
 	if err != nil {
 		return errgo.Mask(err)
@@ -93,10 +94,11 @@ func (s *Store) UploadCharm(url *charm.Reference, ch charm.Charm) error {
 	return s.AddCharm(url, ch, blobHash, size)
 }
 
-// UploadBundle uploads the given bundle to the blob store.
-// This method is provided for tests only. Normal clients
-// should upload bundles through the API.
-func (s *Store) UploadBundle(url *charm.Reference, b charm.Bundle) error {
+// AddBundleWithArchive is like AddBundle but
+// also adds the charm archive to the blob store.
+// This method is provided principally so that
+// tests can easily create content in the store.
+func (s *Store) AddBundleWithArchive(url *charm.Reference, b charm.Bundle) error {
 	blobHash, size, err := s.uploadCharmOrBundle(b)
 	if err != nil {
 		return errgo.Mask(err)
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/juju/charmstore/internal/blobstore"
 	"github.com/juju/charmstore/internal/mongodoc"
+	"github.com/juju/charmstore/params"
 )
 
 // Store represents the underlying charm and blob data stores.
@@ -81,32 +82,56 @@ func (s *Store) putArchive(archive blobstore.ReadSeekCloser) (string, int64, err
 	return blobHash, size, nil
 }
 
-// AddCharm adds a charm to the blob store and to the entities collection
-// associated with the given URL.
-func (s *Store) AddCharm(url *charm.Reference, c charm.Charm) error {
-	// Insert the charm archive into the blob store.
+// UploadCharm uploads the given charm to the blob store.
+// This method is provided for tests only. Normal clients
+// should upload charms through the API.
+func (s *Store) UploadCharm(url *charm.Reference, ch charm.Charm) error {
+	blobHash, size, err := s.uploadCharmOrBundle(ch)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	return s.AddCharm(url, ch, blobHash, size)
+}
+
+// UploadBundle uploads the given bundle to the blob store.
+// This method is provided for tests only. Normal clients
+// should upload bundles through the API.
+func (s *Store) UploadBundle(url *charm.Reference, b charm.Bundle) error {
+	blobHash, size, err := s.uploadCharmOrBundle(b)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	return s.AddBundle(url, b, blobHash, size)
+}
+
+func (s *Store) uploadCharmOrBundle(c interface{}) (blobHash string, size int64, err error) {
 	archive, err := getArchive(c)
 	if err != nil {
-		return errgo.Mask(err)
+		return "", 0, errgo.Mask(err)
 	}
 	defer archive.Close()
-	blobHash, size, err := s.putArchive(archive)
-	if err != nil {
-		return errgo.Mask(err)
-	}
+	return s.putArchive(archive)
+}
 
+// AddCharm adds a charm to the blob store and to the entities collection
+// associated with the given URL.
+func (s *Store) AddCharm(url *charm.Reference, c charm.Charm, blobHash string, blobSize int64) error {
 	// Add charm metadata to the entities collection.
-	return s.DB.Entities().Insert(&mongodoc.Entity{
+	err := s.DB.Entities().Insert(&mongodoc.Entity{
 		URL:                     url,
 		BaseURL:                 baseURL(url),
 		BlobHash:                blobHash,
-		Size:                    size,
+		Size:                    blobSize,
 		CharmMeta:               c.Meta(),
 		CharmConfig:             c.Config(),
 		CharmActions:            c.Actions(),
 		CharmProvidedInterfaces: interfacesForRelations(c.Meta().Provides),
 		CharmRequiredInterfaces: interfacesForRelations(c.Meta().Requires),
 	})
+	if mgo.IsDup(err) {
+		return params.ErrDuplicateUpload
+	}
+	return errgo.Mask(err)
 }
 
 // ExpandURL returns all the URLs that the given URL may refer to.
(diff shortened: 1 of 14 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/81/

internal/router/router_test.go
```diff
@@ -702,9 +703,15 @@ func (s *RouterSuite) TestRouterMethodsThatPassThroughUnresolvedId(c *gc.C) {
 		"POST":   false,
 		"PUT":    false,
 		"GET":    true,
-		"HEAD":   true,
```
rogpeppe, line 705: Note that this failed because the real http logic (as opposed to when we were
faking it with httptest.ResponseRecorder) does not send back the body
when doing a HEAD request. Given that there's no actual logic
in the code that is HEAD specific, it seems reasonable to drop the this case.
An alternative would be to refactor the test to make the handler
store the id in a local variable.

internal/v4/archive.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (h *handler) servePostArchive(id *charm.Reference, w http.ResponseWriter, req *http.Request) (resp *params.ArchivePostResponse, err error) {
+	// Validate the request parameters.
+	if id.Series == "" {
+		return nil, badRequestf(nil, "series not specified")
+	}
+	if id.Revision != -1 {
+		return nil, badRequestf(nil, "revision specified, but should not be specified")
+	}
+	hash := req.Form.Get("hash")
```
jrwren, line 66: The spec names this sha256 and says its not required. "The sha256 flag may specify the hash of the uploaded archive in hexadecimal format. If the same content has already been uploaded, the response will return immediately without reading the entire body."

I might be reading the spec wrong. If it is required can we change may to must in the spec?

internal/v4/archive.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	defer r.Close()
+	defer func() {
+		if err != nil {
+			h.store.BlobStore.Remove(hash)
+			// TODO(rog) log if remove fails.
+		}
+	}()
+
+	// Create the entry for the entity in charm store.
```
jrwren, line 93: This comment looks like it applies to the next line, but that is not what the next line is doing.

internal/v4/archive.go
rogpeppe, line 66: Good catch. We aren't using the sha256 name because the hash is not actually an sha256 hash.
It's a concatenation of sha256 and md5 hashes, because that's what the github.com/juju/blobstore
currently uses. I'm not fully satisfied what we're going to end up with, so we've left it as just "hash"
for the time being.

Currently, we do require the hash, so I think it should probably be "must" in the spec.

internal/v4/archive.go
```diff
@@ -0,0 +1,151 @@
...
+	return &params.ArchivePostResponse{
+		Id: id,
+	}, nil
+}
+
+type readerAtSeeker struct {
+	r io.ReadSeeker
+}
+
+func (r *readerAtSeeker) ReadAt(buf []byte, p int64) (int, error) {
```
jrwren, line 135: This looks unused. Am I missing where it is used?

internal/v4/archive.go
```diff
@@ -0,0 +1,151 @@
...
+		if err := h.store.AddCharm(id, ch, hash, req.ContentLength); err != nil {
+			return nil, errgo.Mask(err, errgo.Is(params.ErrDuplicateUpload))
+		}
+	}
+	return &params.ArchivePostResponse{
+		Id: id,
+	}, nil
+}
+
+type readerAtSeeker struct {
```
jrwren, line 131: Unnecessary if ReadAt isn't used.

internal/v4/archive.go
frankban, line 93: It applies to the next code block, like all the other comments in this method.

internal/v4/archive.go
frankban, line 131: It is necessary because charm.Read(Charm|Bundle)ArchiveFromReader requires an io.ReaderAt.

internal/v4/archive.go
frankban, line 135: As mentioned, this is defined to implement the io.ReaderAt interface.

jrwren:
LGTM

internal/charmstore/store.go
```diff
@@ -81,32 +82,56 @@ func (s *Store) putArchive(archive blobstore.ReadSeekCloser) (string, int64, err
...
 
-// AddCharm adds a charm to the blob store and to the entities collection
-// associated with the given URL.
-func (s *Store) AddCharm(url *charm.Reference, c charm.Charm) error {
-	// Insert the charm archive into the blob store.
-	archive, err := getArchive(c)
+// UploadCharm uploads the given charm to the blob store.
+// This method is provided for tests only. Normal clients
+// should upload charms through the API.
+func (s *Store) UploadCharm(url *charm.Reference, ch charm.Charm) error {
```
urosj: Should we indicate in the name, that it is just for tests? I'm also wondering if we need it in Store at all if it is just "internal".

internal/charmstore/store.go
```diff
@@ -81,32 +82,56 @@ func (s *Store) putArchive(archive blobstore.ReadSeekCloser) (string, int64, err
...
 	}
-	defer archive.Close()
-	blobHash, size, err := s.putArchive(archive)
+	return s.AddCharm(url, ch, blobHash, size)
+}
+
+// UploadBundle uploads the given bundle to the blob store.
+// This method is provided for tests only. Normal clients
+// should upload bundles through the API.
+func (s *Store) UploadBundle(url *charm.Reference, b charm.Bundle) error {
```
urosj: Same comment as with UploadCharm.

internal/charmstore/store.go
rogpeppe: I'm not sure it matters that much. Everything in Store is internal to the charmstore repo
anyway. UploadCharm is useful for tests outside of the internal charmstore package,
where we don't necessarily have an API available. We could lose the "for tests only"
comment if you like - it's actually a perfectly valid method, but just one that we
don't happen to need in production code.

internal/charmstore/store.go
rogpeppe: ditto :-)

internal/v4/archive_test.go
```diff
@@ -0,0 +1,413 @@
...
+
+func (s *ArchiveSuite) SetUpTest(c *gc.C) {
+	s.IsolatedMgoSuite.SetUpTest(c)
+	s.srv, s.store = newServer(c, s.Session)
+}
+
+func (s *ArchiveSuite) TestArchiveGet(c *gc.C) {
+	wordpress := charmtesting.Charms.CharmArchive(c.MkDir(), "wordpress")
+	url := mustParseReference("cs:precise/wordpress-23")
+	err := s.store.UploadCharm(url, wordpress)
```
urosj: Please correct the call from store.UploadCharm ...

internal/charmstore/store.go
urosj: Renaming this to AddCharmWithArchive as agreed.

internal/charmstore/store.go
urosj: Renaming it to AddBundleWithArchive as agreed.

internal/router/router_test.go
urosj, line 705: Maybe instead of dropping it out from code, we might leave it in, as a comment, and documented with the given explanation.

internal/storetesting/http.go
```diff
@@ -76,18 +83,32 @@ func AssertJSONCall(c *gc.C, p JSONCallParams) {
...
-	req.Header.Set("Content-Type", bodyContentType)
-	rec := httptest.NewRecorder()
-	handler.ServeHTTP(rec, req)
-	return rec
+	resp, err := http.DefaultClient.Do(req)
+	c.Assert(err, gc.IsNil)
+	defer resp.Body.Close()
+
+	// TODO(rog) don't return a ResponseRecorder because we're not actually
+	// using httptest.NewRecorder ?
```
urosj, line 106: But we do use ResponseRecorder type?

rogpeppe, line 106: Yes, but it's arguably not right to use it because we're not creating it the usual way.
It's not a big issue, but is probably worth fixing at some point.

internal/v4/archive.go
```diff
@@ -0,0 +1,151 @@
...
+		resp, err := h.servePostArchive(charmId, w, req)
+		if err != nil {
+			return err
+		}
+		return router.WriteJSON(w, http.StatusOK, resp)
+	case "GET":
+	}
+	var entity mongodoc.Entity
+	if err := h.store.DB.Entities().
+		FindId(charmId).
```
urosj, line 40: implementation question: FindId works on charm.Reference properly? Are all fields of the reference available (it is not charm.URL)?

rogpeppe, line 40: Yes, it's fine (this code hasn't changed in this PR, just moved from api.go).

A charm.Reference is just like a charm.URL except that it is not guaranteed
to have a Series. My point of view on this is that it's not really worth representing
that distinction in the type system and so we're using charm.Reference throughout,
which makes other things more straightforward.

internal/v4/archive.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, errgo.Notef(err, "cannot get next revision for id")
+	}
+	id.Revision = rev
+	readerAt := &readerAtSeeker{r}
+	if id.Series == "bundle" {
+		b, err := charm.ReadBundleArchiveFromReader(readerAt, req.ContentLength)
+		if err != nil {
+			return nil, errgo.Notef(err, "cannot read bundle archive")
+		}
+		if err := b.Data().Verify(func(string) error { return nil }); err != nil {
```
urosj: I had to go deep into budnledata code to understand that part of { return nil } ... This will have to change into "VerifyWithCharms" at some point.

internal/v4/archive.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (h *handler) nextRevisionForId(id *charm.Reference) (int, error) {
+	id1 := *id
+	id1.Revision = -1
+	err := ResolveURL(h.store, &id1)
+	if err == nil {
+		return id1.Revision + 1, nil
+	}
+	if errgo.Cause(err) != params.ErrNotFound {
+		return 0, errgo.Notef(err, "cannot resolve id")
```
urosj, line 150: What about returning -1 here?

internal/charmstore/store.go
rogpeppe: done

internal/charmstore/store.go
rogpeppe: done.

internal/v4/archive.go
frankban: +1
This can be made more readable like:
verifyConstraints := func(string) error { return nil }
// TODO: provide a real constrains verifier.
if err := b.Data().Verify(verifyConstraints); err != nil {
...

internal/router/router_test.go
rogpeppe, line 705: done

internal/v4/archive.go
frankban, line 150: I guess it's idiomatic in go to return the zero value for the given type in the case the function returns an error. If there is an error, the value is ignored anyway.

internal/v4/archive_test.go
rogpeppe: done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/82/

internal/v4/archive.go
rogpeppe, line 150: > I guess it's idiomatic in go to return the zero value for the given type
> in the case the function returns an error. If there is an error, the value is ignored anyway.

+1

internal/v4/archive.go
rogpeppe: You should not have needed to go deep into the bundledata code to
understand that - it's part of the documented contract of the function.
Perhaps the wording of that should be improved (I thought that
"The verifyConstraints function is called to verify any constraints
that are found." was fairly self-evident, but I guess not).

I agree with frankban's suggestion regardless, so I've done that.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/83/

internal/v4/archive.go
urosj, line 150: Ok. Tnx ... will use that from now on.

internal/v4/archive.go
rogpeppe, line 93: I added a blank line after each comment to try to make the scope of the
comments clearer.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/84/

urosj:
:+1: great work, guys

rogpeppe:
:shipit:

jujugui:

PR #64, opened 2014-08-08
store: implement serveArchiveFile API.

frankban:


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/87/

internal/v4/archive.go
```diff
@@ -128,6 +125,48 @@ func verifyConstraints(s string) error {
...
+		}
+		defer content.Close()
+		// Send the response to the client.
+		ctype := mime.TypeByExtension(filepath.Ext(filePath))
+		if ctype != "" {
+			w.Header().Set("Content-Type", ctype)
+		}
+		w.Header().Set("Content-Length", strconv.FormatInt(fileInfo.Size(), 10))
+		w.WriteHeader(http.StatusOK)
+		io.Copy(w, content)
```
jrwren, line 164: Should return value be checked here comparing to content length? Should err be logged if there is one?

jrwren:
:+1:

internal/v4/archive.go
```diff
@@ -152,8 +191,20 @@ func (h *handler) nextRevisionForId(id *charm.Reference) (int, error) {
 	return 0, nil
 }
 
-// GET id/archive/…
-// http://tinyurl.com/lampm24
-func (h *handler) serveArchiveFile(charmId *charm.Reference, w http.ResponseWriter, req *http.Request) error {
-	return errNotImplemented
+func (h *handler) openBlobById(id *charm.Reference) (blobstore.ReadSeekCloser, int64, error) {
```
rogpeppe: just openBlob, perhaps? I'm not sure how else you might open one except by id.

internal/v4/archive_test.go
```diff
@@ -368,6 +370,98 @@ func (s *ArchiveSuite) assertUpload(c *gc.C, url *charm.Reference, fileName stri
...
+	// Retrieve the expected bytes.
+	var expectBytes []byte
+	for _, file := range zipFile.File {
+		if file.Name == filePath {
+			r, err := file.Open()
+			c.Assert(err, gc.IsNil)
+			defer r.Close()
+			expectBytes, err = ioutil.ReadAll(r)
+			c.Assert(err, gc.IsNil)
+		}
```
rogpeppe, line 449: break?

internal/v4/archive.go
```diff
@@ -128,6 +125,48 @@ func verifyConstraints(s string) error {
...
+		return err
+	}
+	defer r.Close()
+	zipReader, err := zip.NewReader(&readerAtSeeker{r}, size)
+	if err != nil {
+		return errgo.Notef(err, "cannot read archive data for %s", id)
+	}
+
+	// Retrieve the requested file from the zip archive.
+	filePath := strings.TrimPrefix(req.URL.String(), "/")
```
rogpeppe: This seems a bit odd. Doesn't req.URL.String() start with http:// ?
Shouldn't it be req.URL.Path ? I'm not sure quite how it's working right now.
Hmm, looks like the URL doesn't have a method. Still better
to  use req.URL.Path though, I think.

Also, perhaps path.Clean the URL path?

internal/v4/archive.go
```diff
@@ -128,6 +125,48 @@ func verifyConstraints(s string) error {
...
+	for _, file := range zipReader.File {
+		if path.Clean(file.Name) != filePath {
+			continue
+		}
```
rogpeppe, line 146: Do we want to TrimPrefix("./") perhaps? I remember seeing some logic
that did that once, and it makes some sense.

internal/v4/archive.go
rogpeppe, line 164: I think this is OK as is. The content can be truncated for a few reasons (e.g. client going down),
and I'm not sure that logging that message is likely to be that useful.

rogpeppe:
LGTM with a few minor suggestion. Nice, thanks.

internal/v4/archive.go
frankban: By hash, by entity? [shrug]
I don't have a strong opinion, so I guess I'll change this as you suggested.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/93/

internal/v4/archive_test.go
frankban, line 449: Oh, good catch, fixed.

internal/v4/archive.go
frankban: It doesn't, but I agree with you it's better to use URL.Path here.
Done.

internal/v4/archive.go
frankban, line 146: path.Clean does that automatically.

frankban:
Thanks for the reviews!
:shipit:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/94/

jujugui:

PR #65, opened 2014-08-08
Fix the meta archive-size response.

frankban:
Now the size is wrapped in a struct as required by the spec.

### DIFF
--- a/internal/v4/api_test.go
+++ b/internal/v4/api_test.go
@@ -106,14 +106,13 @@ var metaEndpoints = []metaEndpoint{{
 		c.Assert(data.(*charm.Actions).ActionSpecs["snapshot"].Description, gc.Equals, "Take a snapshot of the database.")
 	},
 }, {
-	name:            "archive-size", // Charm size.
-	get:             entityGetter(entitySizeGetter),
+	name: "archive-size",
+	get: entityGetter(func(entity *mongodoc.Entity) interface{} {
+		return &params.ArchiveSizeResponse{
+			Size: entity.Size,
+		}
+	}),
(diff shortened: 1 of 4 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/89/

jrwren:
:+1:

internal/v4/api_test.go
```diff
@@ -107,14 +107,14 @@ var metaEndpoints = []metaEndpoint{{
 	},
 }, {
 	name:            "archive-size", // Charm size.
-	get:             entityFieldGetter("Size"),
+	get:             entityGetter(entitySizeGetter),
 	checkURL:        "cs:precise/wordpress-23",
-	assertCheckData: sizeChecker,
+	assertCheckData: entitySizeChecker,
 }, {
 	name:            "archive-size", // Bundle size.
```
rogpeppe: I don't think we need this entry. The endpoint is the same whether the
id is a charm or a bundle. We check both charms and bundles against
this endpoint in the tests.

Once we delete the unneeded endpoint, I'd fold entitySizeGetter
back into the assertCheckData field as a literal function.

rogpeppe:
LGTM with the above change, thanks.

internal/v4/api_test.go
frankban: Done.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/90/

PR #70, opened 2014-08-13
Bundle validation: verify with charms.

frankban:
Validate the bundle with respect to charms in the store.


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/107/

internal/v4/archive.go
```diff
@@ -208,3 +213,52 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+		}
+		if err = h.resolveURL(url); err != nil {
+			if errgo.Cause(err) == params.ErrNotFound {
+				// Ignore this error too, for the same reasons
+				// described above.
+				continue
+			}
+			return nil, err
+		}
+		r, size, err := h.openBlob(url)
```
rogpeppe: I don't think it's necessary to open the blob. All the information is available in the
Entity document. I'd define a type, say entityCharm, which implements the
methods required by the charm.Charm interface. 

e.g.
    type entityCharm mongodoc.Entity
    func (e *entityCharm) Meta() *charm.Meta {
        return e.CharmMeta
    }
    ... etc

Then you can just retrieve all the entities from the entities collection with the
appropriate ids (entities.Find(bson.D{{"_id", bson.D{{"$in", ids}}}}).All(&found)
might work ok) put them into a map and return them as is.

internal/v4/archive.go
```diff
@@ -208,3 +213,52 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+		}
+		charms[id] = ch
+	}
+	return charms, nil
+}
+
+// verificationError returns an error whose string representation includes all
+// the verification error messages stored in err.
+// Note that err must be a *charm.VerificationError.
+func verificationError(err error) error {
```
rogpeppe, line 284: I'm not _entirely_ sure we want to return all the error messages here. Or rather, if we
do return all the error messages, it would be good if they were in a form that was
easy for a client to split into separate messages for display. I'm not sure that
we guarantee that the error messages in VerificationError.Errors do not contain
semicolons.

My first thought was that we could just return the single error, and leave the
multiple errors to be printed by a client-side tool, run independently of the
charm store.

My second thought is that we could pack all the errors into JSON and
have that as part of the returned error message.

i.e. the returned error message would look like:

```
bundle verification error: ["charm \"wrdpress\" not found","another error etc"]
```

Then it's easy for a client to split them out again.

What do you think?

internal/v4/archive.go
```diff
@@ -208,3 +213,52 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+
+func (h *handler) bundleCharms(ids []string) (map[string]charm.Charm, error) {
+	charms := make(map[string]charm.Charm, len(ids))
+	for _, id := range ids {
+		url, err := charm.ParseReference(id)
+		if err != nil {
+			// Ignore this error. This will be caught in the bundle
+			// verification process and will be returned to the user
+			// along with other bundle errors.
+			continue
```
urosj, line 229: verified in the bundleData.VerifyWithCharms?

internal/v4/archive.go
```diff
@@ -208,3 +213,52 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
 	}
 	return r, size, nil
 }
+
+func (h *handler) bundleCharms(ids []string) (map[string]charm.Charm, error) {
```
rogpeppe, line 218: // bundleCharms retrieves charms with the given ids from the store and returns them
// in a form suitable for passing to charm.BundleData.VerifyWithCharms.

?

rogpeppe:
Excellent direction, thanks, but with some suggested simplifications above.

internal/v4/archive.go
urosj, line 218: Does it make sense to do change this into: 
func (h *handler) bundleCharms(ids []string) (map[string]charm.Charm, error) {
  return bundleCharmsSequentialGet(ids) {
} 
and renaming current func into:
func (h *handler) bundleCharmsSequentialGet(ids []string) (map[string]charm.Charm, error) {
so that:
a) it is clear that it is sequential
b) it is easy to add concurrent check

internal/v4/archive.go

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/108/

internal/v4/archive.go
frankban, line 284: As discussed, I went with your second suggestion, thanks.

internal/v4/archive.go
frankban, line 229: Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/109/

internal/v4/archive.go
```diff
@@ -208,3 +214,76 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+}
+
+func (e *entityCharm) Revision() int {
+	return e.URL.Revision
+}
+
+func (h *handler) bundleCharms(ids []string) (map[string]charm.Charm, error) {
+	urls := make([]*charm.Reference, len(ids))
+	urlIdmap := make(map[charm.Reference]string, len(ids))
+	for i, id := range ids {
```
rogpeppe: // TODO resolve ids concurrently

?

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/110/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/111/

internal/v4/archive.go
```diff
@@ -208,3 +214,76 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+func (e *entityCharm) Actions() *charm.Actions {
+	return e.CharmActions
+}
+
+func (e *entityCharm) Revision() int {
+	return e.URL.Revision
+}
+
+func (h *handler) bundleCharms(ids []string) (map[string]charm.Charm, error) {
+	urls := make([]*charm.Reference, len(ids))
```
rogpeppe: This doesn't seem quite right. If URL resolving fails, we can be left
with empty slots here.

I'd suggest:
    urls := make([]*charm.Reference, 0, len(ids))

and then append to urls after parsing each one.

It would be good to have some better test coverage in this area. Perhaps
a unit test of bundleCharms itself?

internal/v4/archive.go
```diff
@@ -208,3 +214,76 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+	urlIdmap := make(map[charm.Reference]string, len(ids))
```
rogpeppe: I'm not sure about charm.Reference here. How about using *charm.Reference; that
way you can associate the exact *charm.Reference stored in the urls slice
with the string that created it.

Alternatively (and slightly more efficient) you could just have a slice
to go alongside the urls slice:

  idKeys := make([]string, 0, len(ids))

then append the original form of each id to it when you
append to urls.

On balance, I think I prefer the latter.

internal/v4/archive.go
```diff
@@ -208,3 +214,76 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+		urlIdmap[*url] = id
+	}
+	var entities []mongodoc.Entity
+	if err := h.store.DB.Entities().
+		Find(bson.D{{"_id", bson.D{{"$in", urls}}}}).
+		All(&entities); err != nil {
+		return nil, err
+	}
+	charms := make(map[string]charm.Charm, len(entities))
+	for _, entity := range entities {
```
rogpeppe: I think I'd avoid allocating a new entityCharm for each element here by doing:

```
for i := range entities {
     charms[idKeys[i]] = entityCharm(&entities[i])
}
```

(assuming the change to use idKeys suggested above)

internal/v4/archive.go
```diff
@@ -208,3 +214,76 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+		charms[id] = &ch
+	}
+	return charms, nil
+}
+
+// verificationError returns an error whose string representation includes all
+// the verification error messages stored in err.
+// Note that err must be a *charm.VerificationError.
+func verificationError(err error) error {
+	verr := err.(*charm.VerificationError)
```
rogpeppe: I think I'd prefer to code defensively here, and do:

```
verr, ok := err.(*charm.VerificationError)
if !ok {
    return err
}
```

rogpeppe:
Nicer, thanks. A few further suggestions above.

internal/v4/archive.go

jrwren:
:+1:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/112/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/113/

internal/v4/archive.go
```diff
@@ -208,3 +214,82 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
 	}
 	return r, size, nil
 }
+
+// entityCharm implements charm.Charm.
```
rogpeppe: perhaps move this (and the methods) after bundleCharms?

internal/v4/archive.go
```diff
@@ -208,3 +214,82 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+		idKeys = append(idKeys, id)
+	}
+	var entities []mongodoc.Entity
+	if err := h.store.DB.Entities().
+		Find(bson.D{{"_id", bson.D{{"$in", urls}}}}).
+		All(&entities); err != nil {
+		return nil, err
+	}
+	entityCharms := make(map[charm.Reference]entityCharm, len(entities))
+	for _, entity := range entities {
```
rogpeppe: slightly better, i think:

```
for i := range entities {
    entityCharms[*entity.URL] = (*entityCharm)(&entities[i])
}
```

then entityCharms can be map[charm.Reference] charm.Charm.
This will generate less garbage.

internal/v4/archive.go
```diff
@@ -208,3 +214,82 @@ func (h *handler) openBlob(id *charm.Reference) (blobstore.ReadSeekCloser, int64
...
+		}
+	}
+	return charms, nil
+}
+
+// verificationError returns an error whose string representation is a list of
+// all the verification error messages stored in err, in JSON format.
+// Note that err must be a *charm.VerificationError.
+func verificationError(err error) error {
+	verr := err.(*charm.VerificationError)
```
rogpeppe: I still think this should be defensively coded.

internal/v4/archive_test.go
```diff
@@ -485,6 +501,98 @@ func (s *ArchiveSuite) assertArchiveFileContents(c *gc.C, zipFile *zip.ReadClose
...
+	wordpress := charmtesting.Charms.CharmArchive(c.MkDir(), "wordpress")
+	err = s.store.AddCharmWithArchive(
+		mustParseReference("cs:utopic/wordpress-47"), wordpress)
+	c.Assert(err, gc.IsNil)
+
+	// Retrieve the bundleCharms method.
+	handler := v4.New(s.store)
+	bundleCharms := v4.BundleCharms(handler)
+
+	tests := []struct {
```
rogpeppe, line 523: Nice tests, thanks.

rogpeppe:
LGTM with a few minor suggestions, thanks!

internal/v4/archive.go
frankban: Missed this suggestion before, darn...

internal/v4/archive.go

internal/v4/archive.go

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/114/

frankban:
:shipit: thank you for the great reviews!

jujugui:

PR #71, opened 2014-08-13
use blob names instead of hashes

rogpeppe:
We can't use content addressing with juju blob store,
as discussed ad nauseam online.


jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/116/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/117/

jrwren:
LGTM

internal/charmstore/store_test.go
```diff
@@ -56,9 +56,13 @@ func (s *StoreSuite) checkAddCharm(c *gc.C, ch charm.Charm) {
...
 
+	blobName := doc.BlobName
+	c.Assert(blobName, gc.Matches, "[0-9a-z]+")
+	doc.BlobName = ""
 	c.Assert(doc, jc.DeepEquals, mongodoc.Entity{
-		URL:                     url,
-		BaseURL:                 mustParseReference("cs:wordpress"),
+		URL:     url,
+		BaseURL: mustParseReference("cs:wordpress"),
+
```
frankban, line 65: Is this new line intentional?

internal/charmstore/store_test.go
```diff
@@ -56,9 +56,13 @@ func (s *StoreSuite) checkAddCharm(c *gc.C, ch charm.Charm) {
 
 	doc.UploadTime = time.Time{}
 
+	blobName := doc.BlobName
```
frankban, line 59: What do you think about a comment here explaining the blob name is set as a random string?

internal/mongodoc/doc.go
```diff
@@ -18,8 +18,16 @@ type Entity struct {
...
-	Size     int64
+	// BlobHash holds the hash checksum of the blob, in hexadecimal format,
+	// as created by blobstore.NewHash.
+	BlobHash string
+
+	// Size holds the size of the archive blob.
+	// TODO(rog) rename this to BlobSize.
+	Size int64
+
+	// BlobName holds the  name that the archive blob is given in the blob store.
```
frankban, line 29: Double space between "the" and "name".

internal/v4/api.go
```diff
@@ -92,20 +92,54 @@ func (h *handler) resolveURL(url *charm.Reference) error {
 type entityHandlerFunc func(entity *mongodoc.Entity, id *charm.Reference, path string, method string, flags url.Values) (interface{}, error)
 
 // entityHandler returns a handler that calls f with a *mongodoc.Entity that
-// contains at least the given fields.
+// contains at least the given fields. It allows only GET requests.
```
frankban, line 95: The changes from here seem to try to fix another issue you didn't describe in the MP description.
No problem if this branch is really intended to solve two issues.

internal/v4/api.go
```diff
@@ -92,20 +92,54 @@ func (h *handler) resolveURL(url *charm.Reference) error {
...
 func (h *handler) entityHandler(f entityHandlerFunc, fields ...string) router.BulkIncludeHandler {
+	return h.entityHandler0(f, false, fields...)
+}
+
+// entityHandler returns a handler that calls f with a *mongodoc.Entity that
```
frankban, line 100: entityHandlerWithPut ...

internal/v4/api.go
```diff
@@ -92,20 +92,54 @@ func (h *handler) resolveURL(url *charm.Reference) error {
 type entityHandlerFunc func(entity *mongodoc.Entity, id *charm.Reference, path string, method string, flags url.Values) (interface{}, error)
 
 // entityHandler returns a handler that calls f with a *mongodoc.Entity that
-// contains at least the given fields.
+// contains at least the given fields. It allows only GET requests.
 func (h *handler) entityHandler(f entityHandlerFunc, fields ...string) router.BulkIncludeHandler {
+	return h.entityHandler0(f, false, fields...)
```
frankban, line 97: I wonder if we can found a name better than entityHandler0.
It does not seem to really describe what it does.

internal/v4/api.go
```diff
@@ -92,20 +92,54 @@ func (h *handler) resolveURL(url *charm.Reference) error {
...
+		),
+	}
+}
+
+type entityHandler struct {
+	allowPut bool
+	router.BulkIncludeHandler
+}
+
+func (h *entityHandler) Handle(hs []router.BulkIncludeHandler, id *charm.Reference, paths []string, method string, flags url.Values) ([]interface{}, error) {
```
frankban, line 131: I don't see how this solves the problem of an endpoint wanting to access other attributes of the request not passed by the BulkIncludeHandler framework.

AFAICT this just adds a check for allowed methods, and the special case for PUT feels like too special at first sight (but perhaps I am wrong).

My impression is that we are adding code that will call us, instead of letting endpoint handlers decide what checks to perform, and how to modify their own arguments explicitly.
E.g. it would be good to have a generic:
func allowedMethods(handler http.Handler, methods ...string) http.Handler
or just
func allowedMethods(handler func(w http.ResponseWriter, r *http.Request), methods ...string) func(w http.ResponseWriter, r *http.Request)

As another example, for the expand-id endpoint we would need the unresolved Id, but (if I am not missing something) at the time the handler is called the id has been already resolved by the framework.
We would be forced to add another special case to the framework, and that doesn't feel optimal to me.

To be clear, this is more a request for discussion (and agreement on the direction we are following) than a request for actual changes. What do you think?

internal/v4/archive.go
```diff
@@ -72,29 +72,33 @@ func (h *handler) servePostArchive(id *charm.Reference, w http.ResponseWriter, r
...
 		if err != nil {
-			h.store.BlobStore.Remove(hash)
+			h.store.BlobStore.Remove(name)
 			// TODO(rog) log if remove fails.
 		}
 	}()
 
-	// Create the entry for the entity in charm store.
-
+	// Assign a new revision to the id, so that we know what name
```
frankban, line 92: I am not sure I understand this comment.

frankban:
This branch is nice Roger, :+1: 
Thanks for handling this transition to using names as blobstore paths.
I have a couple of suggestions and some questions.
Have you planned to implement a test exercising the refcount bug we solved with this branch?
I know we have a test that intermittently failed, but it would be nice to have a specific case for the refcount problem, if possible.

internal/v4/api.go
rogpeppe, line 95: This is mixed up from another branch. Will fix and re-push.

internal/v4/api.go
rogpeppe, line 131: This discussion should happen in a different code review :-)

internal/v4/archive.go
rogpeppe, line 92: It was a hangover from a previous iteration.

rogpeppe:
Superceded by https://github.com/juju/charmstore/pull/73

PR #75, opened 2014-08-18
Implement the charm-related API.

frankban:
Also move MgoTestPackage suite registration out from api_test to its
own package_test.go.

### DIFF
--- a/internal/v4/package_test.go
+++ b/internal/v4/package_test.go
@@ -1,0 +1,14 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package v4_test
+
+import (
+	"testing"
+
+	jujutesting "github.com/juju/testing"
+)
+
+func TestPackage(t *testing.T) {
+	jujutesting.MgoTestPackage(t, nil)
+}
--- a/internal/v4/relations.go
+++ b/internal/v4/relations.go
@@ -1,0 +1,134 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package v4
+
+import (
+	"net/url"
+
+	"github.com/juju/errgo"
+	"gopkg.in/juju/charm.v3"
+	"gopkg.in/mgo.v2/bson"
+
+	"github.com/juju/charmstore/internal/mongodoc"
+	"github.com/juju/charmstore/params"
+)
+
+// GET id/meta/charm-related[?include=meta[&include=meta…]]
+// http://tinyurl.com/q7vdmzl
+func (h *handler) metaCharmRelated(entity *mongodoc.Entity, id *charm.Reference, path, method string, flags url.Values) (interface{}, error) {
+	// If the charm does not define any relation we can just return without
+	// hitting the db.
+	if len(entity.CharmProvidedInterfaces)+len(entity.CharmRequiredInterfaces) == 0 {
+		return &params.RelatedResponse{}, nil
+	}
+
+	// Build the query to retrieve the related entities.
+	query := bson.M{
+		"$or": []bson.M{
+			bson.M{"charmrequiredinterfaces": bson.M{
+				"$elemMatch": bson.M{"$in": entity.CharmProvidedInterfaces},
+			}},
+			bson.M{"charmprovidedinterfaces": bson.M{
+				"$elemMatch": bson.M{"$in": entity.CharmRequiredInterfaces},
+			}},
+		},
+	}
+	fields := bson.D{
+		{"_id", 1},
+		{"charmrequiredinterfaces", 1},
+		{"charmprovidedinterfaces", 1},
+	}
+
+	// Retrieve the entities from the database.
+	var entities []mongodoc.Entity
+	if err := h.store.DB.Entities().Find(query).Select(fields).All(&entities); err != nil {
+		return nil, errgo.Notef(err, "cannot retrieve the related charms")
+	}
+
+	// If no entities are found there is no need for further processing the
+	// results.
+	if len(entities) == 0 {
+		return &params.RelatedResponse{}, nil
+	}
+
+	// Build the results, by grouping entities based on their relations' roles
+	// and interfaces.
+	includes := flags["include"]
+	requires, err := h.getRelatedCharmsResponse(entity.CharmProvidedInterfaces, entities, func(e mongodoc.Entity) []string {
+		return e.CharmRequiredInterfaces
+	}, includes)
+	if err != nil {
+		return nil, errgo.Notef(err, "cannot retrieve the charm requires")
+	}
+	provides, err := h.getRelatedCharmsResponse(entity.CharmRequiredInterfaces, entities, func(e mongodoc.Entity) []string {
+		return e.CharmProvidedInterfaces
+	}, includes)
+	if err != nil {
+		return nil, errgo.Notef(err, "cannot retrieve the charm provides")
+	}
+
+	// Return the response.
+	return &params.RelatedResponse{
+		Requires: requires,
+		Provides: provides,
+	}, nil
+}
+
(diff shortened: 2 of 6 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/125/

internal/v4/package_test.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"testing"
+
+	jujutesting "github.com/juju/testing"
+)
+
+func TestPackage(t *testing.T) {
+	jujutesting.MgoTestPackage(t, nil)
+}
```
urosj, line 14: Nice separation, makes api_tests.go cleaner. Thanks.

internal/v4/relations.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/juju/charmstore/internal/mongodoc"
+	"github.com/juju/charmstore/params"
+)
+
+// GET id/meta/charm-related[?include=meta[&include=meta…]]
+// http://tinyurl.com/q7vdmzl
+func (h *handler) metaCharmRelated(entity *mongodoc.Entity, id *charm.Reference, path, method string, flags url.Values) (interface{}, error) {
+	// If the charm does not define any relation we can just return without
+	// hitting the db.
+	if len(entity.CharmProvidedInterfaces)+len(entity.CharmRequiredInterfaces) == 0 {
```
urosj, line 22: or just simply len(entity.CharmProvidedInterfaces) == 0 && len(entity.CharmRequiredInterfaces) == 0?

internal/v4/relations.go
```diff
@@ -0,0 +1,134 @@
...
+		return e.CharmRequiredInterfaces
+	}, includes)
+	if err != nil {
+		return nil, errgo.Notef(err, "cannot retrieve the charm requires")
+	}
+	provides, err := h.getRelatedCharmsResponse(entity.CharmRequiredInterfaces, entities, func(e mongodoc.Entity) []string {
+		return e.CharmProvidedInterfaces
+	}, includes)
+	if err != nil {
+		return nil, errgo.Notef(err, "cannot retrieve the charm provides")
```
urosj, line 68: I'm never sure what errgo func to use here :S

internal/v4/relations.go
bac, line 22: I like Francesco's original.  Would like to see spaces around the + though.  Or is that not a go thing?

internal/v4/relations_test.go
```diff
@@ -0,0 +1,341 @@
...
+				Name:      "reverseproxy",
+				Role:      "requirer",
+				Interface: "http",
+			},
+		},
+	},
+}
+
+var metaCharmRelatedTests = []struct {
+	// About the test.
```
bac: Maybe "Description of the test." ?

bac:
:+1:

internal/v4/relations_test.go
frankban: Done.

internal/v4/relations.go
frankban, line 22: It is the result of gofmt.

internal/v4/relations.go
```diff
@@ -0,0 +1,134 @@
...
+// charms. For instance:
+//   map[string][]params.MetaAnyResponse{
+//       "http": []params.MetaAnyResponse{
+//           {Id: "cs:utopic/django-42", Meta: ...},
+//           {Id: "cs:trusty/wordpress-47", Meta: ...},
+//       },
+//       "memcache": []params.MetaAnyResponse{
+//           {Id: "cs:utopic/memcached-0", Meta: ...},
+//       },
+//   }
```
urosj, line 90: +1 for detailed comments!

internal/v4/relations.go
```diff
@@ -0,0 +1,134 @@
...
+	getInterfaces entityRelatedInterfacesGetter,
+	includes []string) ([]params.MetaAnyResponse, error) {
+	// Build a list of responses including entities which are related
+	// to the given interface.
+	responses := make([]params.MetaAnyResponse, 0, len(entities))
+	for _, entity := range entities {
+		for _, entityIface := range getInterfaces(entity) {
+			if entityIface == iface {
+				// Retrieve the requested metadata for the entity.
+				meta, err := h.GetMetadata(entity.URL, includes)
```
urosj, line 121: Maybe skipping a call if includes are empty? (if not done in getRelatedCharmsResponse)

internal/v4/relations.go
```diff
@@ -0,0 +1,134 @@
...
+//       },
+//   }
+func (h *handler) getRelatedCharmsResponse(
+	ifaces []string,
+	entities []mongodoc.Entity,
+	getInterfaces entityRelatedInterfacesGetter,
+	includes []string) (map[string][]params.MetaAnyResponse, error) {
+	results := make(map[string][]params.MetaAnyResponse, len(ifaces))
+	for _, iface := range ifaces {
+		responses, err := h.getRelatedIfaceResponses(iface, entities, getInterfaces, includes)
```
urosj, line 98: Do we need responses if includes is empty?

internal/v4/relations.go
frankban, line 121: Not sure. GetMetadata correctly returns an empty map if there are no includes, so the condition here is not strictly required, and would involve another nesting level in the code. It seems to me clearer without the condition. What do you think?

internal/v4/relations.go
frankban, line 98: Yes, because we still need the related charm id.

urosj, line 98: Even if we already know iface?

internal/v4/relations.go
urosj, line 121: Agree on being clearer without additional nesting. GetMetadata does return an empty map, however, it is far from an NOP function when includes is empty. But we might leave it just as it is for now, it's not a big deal.

internal/v4/relations.go
frankban, line 98: For each iface we need to retrieve all the entities providing (or requiring) that iface, and include their id in the result.
The underlying decision here is that, rather than executing one query for each iface, we retrieve all the entities in a single query. After that we need to group the ids (and optionally meta) of those entities by iface programmatically. That's what getRelatedIfaceResponses does.

jrwren:
:+1:

internal/v4/relations_test.go
```diff
@@ -0,0 +1,341 @@
...
+	about:       "includes",
+	charms:      metaCharmRelatedCharms,
+	id:          "precise/nfs-1",
+	querystring: "?include=archive-size&include=charm-metadata",
+	expectBody: params.RelatedResponse{
+		Requires: map[string][]params.MetaAnyResponse{
+			"mount": []params.MetaAnyResponse{{
+				Id: mustParseReference("utopic/wordpress-0"),
+				Meta: map[string]interface{}{
+					"archive-size": params.ArchiveSizeResponse{Size: 42},
```
urosj, line 240: Maybe too fragile. Change in a utopic/wordpress-0 description will brake this test. I would suggest adding extra-info/featured set as true.

urosj:
:+1: Thank you.
(We might think about that change in the tests of not using archive-size but something else - extra-info/featured seems fine).

One more thing: Although Charm.Reference is used, it is expected, that the name is already resolved by the point reference is passed as a parameter.

internal/v4/relations_test.go
frankban, line 240: The archive size is something we pass for those tests. It's not calculated, we just pass 42.
This test mostly checks that the requested info are present in meta.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/126/

frankban:
Thank for the reviews!
:shipit:

jujugui:

PR #78, opened 2014-08-20
Implement the bundles-containing API endpoint.

frankban:


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/133/

internal/charmstore/store.go
```diff
@@ -228,6 +228,9 @@ func bundleCharms(data *charm.BundleData) ([]*charm.Reference, error) {
 			return nil, errgo.Mask(err)
 		}
 		urlMap[url.String()] = url
+		// Also add the corresponding base URL.
+		base := baseURL(url)
+		urlMap[base.String()] = base
```
rogpeppe, line 233: Or just:

   urlMap[base.String() = baseURL(url)

?

internal/v4/relations.go
```diff
@@ -132,3 +132,94 @@ func (h *handler) getRelatedIfaceResponses(
 	}
 	return responses, nil
 }
+
+// GET id/meta/bundles-containing[?include=meta[&include=meta…]][&any-series=1][&any-revision=1]
+// http://tinyurl.com/oqc386r
+func (h *handler) metaBundlesContaining(entity *mongodoc.Entity, id *charm.Reference, path, method string, flags url.Values) (interface{}, error) {
+	anySeries := flags.Get("any-series") == "1"
```
rogpeppe: I think we should probably adhere to the letter of the API spec here,
and reject values that aren't "1", "0" or empty. Perhaps add a little
helper function for that?

internal/v4/relations.go
```diff
@@ -132,3 +132,94 @@ func (h *handler) getRelatedIfaceResponses(
...
+	searchId := *id
+	if anySeries || anyRevision {
+		searchId.Revision = -1
+		searchId.Series = ""
+	}
+
+	// Retrieve the bundles containing the resulting charm id.
+	var entities []mongodoc.Entity
+	if err := h.store.DB.Entities().Find(bson.D{{
+		"bundlecharms", bson.D{{
```
rogpeppe: I don't think you need the $in here.

```
Find(bson.D{{"bundlecharms", &searchId}})
```

should do the trick.

internal/v4/relations.go
urosj: +1 on validation

internal/v4/relations_test.go
```diff
@@ -339,3 +341,239 @@ func (ch *relationTestingCharm) Revision() int {
...
+			"bundle-metadata": metaBundlesContainingBundles["bundle/wordpress-complex-1"].Data(),
+		},
+	}, {
+		Id: mustParseReference("bundle/useless-0"),
+		Meta: map[string]interface{}{
+			"archive-size":    params.ArchiveSizeResponse{Size: 42},
+			"bundle-metadata": metaBundlesContainingBundles["bundle/useless-0"].Data(),
+		},
+	}},
+}}
```
urosj, line 527: What happens when id is "wordpress" or "precise/wordpress"? We must make sure to catch "improper" usage as well.

internal/v4/relations.go
```diff
@@ -132,3 +132,94 @@ func (h *handler) getRelatedIfaceResponses(
...
+			"$in", []*charm.Reference{&searchId},
+		}},
+	}}).Select(bson.D{{"_id", 1}, {"bundlecharms", 1}}).All(&entities); err != nil {
+		return nil, errgo.Notef(err, "cannot retrieve the related bundles")
+	}
+
+	// Further filter the entities if required.
+	if anySeries != anyRevision {
+		filterId := *id
+		var transform func(*charm.Reference) *charm.Reference
```
rogpeppe: I'm finding this logic a little hard to understand.
How about removing one level of abstraction and just implementing
the predicate function inline here?

```
match := func(e *mongodoc.Entity) bool {
    for _, cid := range entity.BundleCharms {
        if cid.Name == id.Name &&
            cid.User == id.User &&
            (anySeries || cid.Series == id.Series) &&
            (anyRevision || cid.Revision == id.Revision) {
            return true
        }
    }
    return false
}
entities = filterEntities(entities, match)
```

internal/v4/relations.go
```diff
@@ -132,3 +132,94 @@ func (h *handler) getRelatedIfaceResponses(
...
+	}
+	return response, nil
+}
+
+// entitiesFilterer functions can be passed as predicate to filterEntities in
+// order to filter entities.
+type entitiesFilterer func(*mongodoc.Entity) bool
+
+// filterEntities filters the given entities based on the given predicate.
+func filterEntities(entities []mongodoc.Entity, predicate entitiesFilterer) []mongodoc.Entity {
```
rogpeppe: I think I'd lose the entitiesFilterer type - I personally feel that func(*mongodoc.Entity) bool is
more descriptive than the entitiesFilterer name.

It would also be fairly trivial to make this just update the slice in place, a fairly
standard idiom:

```
j := 0
for i := range entities {
    entity := &entities[i]
    if predicate(entity) {
        entities[j] = *entity
        j++
    }
 }
 return entities[0:j]
```

internal/v4/relations_test.go
```diff
@@ -4,9 +4,11 @@
 package v4_test
```
rogpeppe, line 4: Please add bundles-containing to the existing metadata endpoints in api_test.go.

internal/v4/relations_test.go
```diff
@@ -339,3 +341,239 @@ func (ch *relationTestingCharm) Revision() int {
 	// relevant.
 	return 0
 }
+
+func (s *RelationsSuite) addBundles(c *gc.C, bundles map[string]charm.Bundle) {
```
rogpeppe: Can't we use the existing addBundle method? (It's currently defined on APISuite). Perhaps
just turn it into a function. Oops, scratch that. I see why you can't use that. I still
think it might be better in singular form though. addBundleWithoutArchive?

internal/v4/relations_test.go
```diff
@@ -339,3 +341,239 @@ func (ch *relationTestingCharm) Revision() int {
...
+	blobSize := int64(42)
```
rogpeppe: Why not just 0?

internal/v4/relations_test.go
```diff
@@ -339,3 +341,239 @@ func (ch *relationTestingCharm) Revision() int {
...
+	for id, b := range bundles {
+		url := mustParseReference(id)
+		// The blob related info are not used in these tests.
+		// The charm-bundle relations are retrieved from the entities
+		// collection, without accessing the blob store.
+		err := s.store.AddBundle(url, b, "blobName", "blobHash", blobSize)
```
rogpeppe: I think I'd be tempted to use a real blob hash here.
If we use 0 as the blob size, then fmt.Sprintf("%x", blobstore.NewHash().Sum(nil))
would do.

internal/v4/relations_test.go
```diff
@@ -339,3 +341,239 @@ func (ch *relationTestingCharm) Revision() int {
...
+	querystring: "?any-series=1",
+	expectBody: []*params.MetaAnyResponse{{
+		Id: mustParseReference("bundle/wordpress-simple-0"),
+	}, {
+		Id: mustParseReference("bundle/wordpress-complex-1"),
+	}},
+}, {
+	about:       "invalid any series",
+	id:          "utopic/mysql-0",
+	querystring: "?any-series=true",
```
rogpeppe, line 456: This should result in an error, I think.

internal/v4/relations_test.go
```diff
@@ -339,3 +341,239 @@ func (ch *relationTestingCharm) Revision() int {
...
+		// Clean up the charm entity in the store.
+		err = s.store.DB.Entities().Remove(bson.D{{"_id", url}})
+		c.Assert(err, gc.IsNil)
+	}
+}
+
+func (s *RelationsSuite) TestMetaBundlesContainingIncludeError(c *gc.C) {
+	// Add a charm.
+	charmUrl := mustParseReference("cs:utopic/django-0")
+	err := s.store.AddCharm(charmUrl, &relationTestingCharm{}, "blobName", "blobHash", int64(47))
```
rogpeppe: Again, please use 0 size and a valid blob hash.

urosj:
LGTM, with some small remarks (validation and "flat" names).

internal/v4/relations_test.go
rogpeppe, line 527: Actually that usage is not improper, but we should definitely have some test cases for it.

rogpeppe:
Very nice stuff, thanks. LGTM modulo the above comments and suggestions.

internal/charmstore/store.go
frankban, line 233: I don't follow this, we need to define base to be able to do `base.String()`.

internal/v4/relations.go
frankban: Did not remember we wrote that in the spec. Good catch, will do.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/134/

internal/v4/relations.go
frankban: Nice, I'll try it.

internal/v4/relations_test.go
frankban, line 527: The id resolution is done by the framework, which is supposed to be already tested.
If you want me to test those here too, that's ok, I agree on adding more tests.

internal/v4/relations.go
frankban: Ok.

internal/v4/relations.go
frankban: I am not sure about this change: I like filter functions with predicates; I don't like (and I feel like they are less functional) mutating the input list in place. Moreover, the code takes more time to be understood (at least by me), especially compared to the simplicity of the original implementation. I don't think the efficiency gain here justifies a more complex code and the possible bugs introduced by in place modifications.
If you don't want me really hard to do this change, I'd leave the code as is.

internal/v4/relations_test.go
frankban, line 4: Could you please be more specific about the reason to do that, given that we have dedicated tests for the endpoint? I am sure I am missing something.

internal/v4/relations_test.go
frankban: This is just an helper in the suite. I've found a map[string]charm.Bundle to be a comfortable data structure to express a set of bundles, and I just wanted to be able to use that directly.
If a more general need comes up we can always implement what you suggest later. What do you think?

internal/v4/relations_test.go
frankban: I wanted something less ambiguous than a zero-value for that type.

internal/v4/relations_test.go
frankban: As stated in the comment, those args do not need to be realistic in this context.
Do they? Moreover a size of 42 is more convincing when testing include=archive-size.

internal/v4/relations_test.go
frankban, line 456: Ok.

internal/v4/relations_test.go
frankban: See above.

frankban:
Thanks for the reviews!

internal/charmstore/store.go
rogpeppe, line 233: Good point. Ignore me :-)

internal/v4/relations.go
rogpeppe: Fair enough.

internal/v4/relations.go
frankban: Done.

internal/v4/relations.go
frankban: Done.

internal/v4/relations_test.go
rogpeppe, line 4: The metadata endpoints in api_test.go make sure that all the metadata endpoint work correctly with the bulk endpoints. All the known metadata endpoints should be represented there.

internal/v4/relations_test.go
frankban, line 527: Done.

internal/v4/relations_test.go
rogpeppe: If it's not going to be used, then it doesn't matter. As it is, it's deliberately wrong, which seems a bit odd. I'd prefer us to use well formed data, even if it's not that important in the current implementation. Perhaps a future implementation might panic with an malformed hash. How about defining a couple of globals:

```
var fakeBlobSize, fakeBlobHash = func() (int64, string) {
     b := []byte("fake content")
     h := blobstore.NewHash()
     h.Write(b)
     return int64(len(b)), fmt.Sprintf("%x", h.Sum(nil))
}()
```

and then using those in the tests?

internal/v4/relations_test.go
rogpeppe, line 527: It's worth having at least one test, I think.

internal/v4/relations.go
frankban: Done.

internal/v4/relations_test.go
frankban, line 456: Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/135/

internal/v4/relations_test.go
rogpeppe: I think map[string]charm.Bundle is a fine data structure to express
a set of bundles, but there are only two calls to addBundles currently - one
uses the map, the other doesn't. I'd incline towards keeping the primitive
simple and using a loop in the former case. It makes it easy to
add more arguments to addBundle if we want to.

internal/v4/relations_test.go
frankban, line 4: OIC, good suggestion!
Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/137/

jrwren:
:+1:

internal/v4/relations_test.go
frankban: Done.

internal/v4/relations_test.go
frankban: Good stuff. Fixed.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/141/

internal/v4/api.go
```diff
@@ -160,6 +160,18 @@ func preferredURL(url0, url1 *charm.Reference) bool {
...
+// stringToBool returns false if value is "0" or empty, true if value is "1",
+// an error otherwise. This can be useful to convert url.Values data.
+func stringToBool(value string) (bool, error) {
+	switch value {
+	case "0", "":
+		return false, nil
+	case "1":
+		return true, nil
+	}
+	return false, errgo.Newf(`value must be either "0", "1" or empty, but %q passed`, value)
```
rogpeppe:
```
errgo.Newf(`unexpected bool value %q (must be "0" or "1")`, value)
```

?

(probably not worth mentioning that it can be empty - that's just to allow
it to be omitted)

internal/v4/api_test.go
```diff
@@ -57,11 +57,13 @@ func storeURL(path string) string {
 type metaEndpointExpectedValueGetter func(*charmstore.Store, *charm.Reference) (interface{}, error)
 
 type metaEndpoint struct {
-	name       string
-	exclusive  int
-	bundleOnly bool
-	get        metaEndpointExpectedValueGetter
-
+	// The name of the meta endpoint.
```
rogpeppe: Thanks for adding these comments. I'd be tempted to stick to the usual style,
with white space between the entries. Here are some suggestions
for slightly different phrasing:

```
// name names the meta endpoint.
name string

// exclusive specifies whether the endpoint is
// valid for charms only (charmOnly), bundles only (bundleOnly)
// or to both (zero).
exclusive int

// get returns the expected data for the endpoint.
get metaEndpointExpectedValueGetter
```

internal/v4/api_test.go
```diff
@@ -147,6 +149,36 @@ var metaEndpoints = []metaEndpoint{{
...
+		// The charms we use for those tests are not related each other.
+		// Charm relations are independently tested in relations_test.go.
+		if url.Series == "bundle" {
+			return nil, nil
+		}
+		return &params.RelatedResponse{}, nil
+	},
+	checkURL: "cs:precise/wordpress-23",
+	assertCheckData: func(c *gc.C, data interface{}) {
+		_ = data.(*params.RelatedResponse)
```
rogpeppe:
```
c.Assert(data, gc.FitsTypeOf, (*params.RelatedResponse)(nil))
```

?

I'm assuming the (non-EndpointGet) tests fail if you change the
get function to return nil always, or &params.RelatedResponse{} always.

If you haven't checked that, please do, just for peace of mind.

internal/v4/api.go
```diff
@@ -160,6 +160,18 @@ func preferredURL(url0, url1 *charm.Reference) bool {
 	return ltsReleases[url0.Series]
 }
 
+// stringToBool returns false if value is "0" or empty, true if value is "1",
```
rogpeppe: parseBool would be a slightly more conventional name for this.
Perhaps follow the wording from strconv.ParseBool?

// parseBool returns the boolean value represented by the string.
// It accepts "1" or "0". Any other value returns an error.

internal/v4/relations.go
```diff
@@ -132,3 +136,79 @@ func (h *handler) getRelatedIfaceResponses(
...
+		}
+		response = append(response, &params.MetaAnyResponse{
+			Id:   e.URL,
+			Meta: meta,
+		})
+	}
+	return response, nil
+}
+
+// filterEntities filters the given entities based on the given predicate.
```
rogpeppe: I'd prefer if this said _how_ it filters the entities.

```
// filterEntities returns a slice containing all the entities
// for which the given predicate returns true.
```

?

rogpeppe:
A few more suggestions, but still LGTM.

internal/v4/api.go
frankban: Done.

internal/v4/api.go
frankban: Done.

internal/v4/api_test.go
frankban: Done.

internal/v4/api_test.go
frankban: Yeah I already checked that.
Added the data assertion.

internal/v4/relations.go
frankban: Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/142/

frankban:
:shipit:

jujugui:

PR #81, opened 2014-08-26
Add basic HTTP authentication to archive POST/DELETE.

frankban:
The credentials can be stored in the charmd configuration file.
Abstract out the authentication logic in a decorator.
Created a card for the DoRequest refactoring.


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/157/

internal/charmstore/server_test.go
```diff
@@ -71,6 +81,23 @@ func (s *ServerSuite) TestNewServerWithVersions(c *gc.C) {
 	assertServesVersion(c, h, "version3")
 }
 
+func (s *ServerSuite) TestNewServerWithConfig(c *gc.C) {
```
jrwren, line 79: What is the goal of exporting this?

cmd/charmd/main.go
```diff
@@ -55,8 +61,30 @@ func serve() error {
...
-	MongoURL string `yaml:"mongo-url"`
-	APIAddr  string `yaml:"api-addr"`
+	MongoURL     string `yaml:"mongo-url"`
+	APIAddr      string `yaml:"api-addr"`
+	AuthUsername string `yaml:"auth-username"`
+	AuthPassword string `yaml:"auth-password"`
+}
+
+func (c *config) validate() error {
+	missing := make([]string, 0)
```
rogpeppe: var missing []string

internal/charmstore/server_test.go
frankban, line 79: I am not sure I understand the question, could you please clarify?

cmd/charmd/main.go
```diff
@@ -55,8 +61,30 @@ func serve() error {
...
+		missing = append(missing, "api-addr")
+	}
+	if c.AuthUsername == "" {
+		missing = append(missing, "auth-username")
+	}
+	if c.AuthPassword == "" {
+		missing = append(missing, "auth-password")
+	}
+	if len(missing) != 0 {
+		return fmt.Errorf("missing %s in config file", strings.Join(missing, ", "))
```
rogpeppe: "missing fields %s ..."
?

internal/charmstore/server.go
```diff
@@ -13,16 +13,18 @@ import (
 	"gopkg.in/mgo.v2"
 
 	"github.com/juju/charmstore/internal/router"
+	"github.com/juju/charmstore/params"
 )
 
 // NewAPIHandler returns a new API handler that
 // uses the given Store.
-type NewAPIHandler func(*Store) http.Handler
+type NewAPIHandler func(*Store, *params.HandlerConfig) http.Handler
```
rogpeppe: I'm not entirely sure about passing the HandlerConfig into the handlers. It seems a little bit
like a boundary violation. I'd prefer to define a type that specifically enumerates the
dependencies that API handlers can have (handlers should not depend on the mongo URL
or the API address, for example).

Perhaps name it HandlerParams?

internal/charmstore/server_test.go
rogpeppe, line 79: If it's not exported, it won't be run as a test.

internal/storetesting/http.go
```diff
@@ -84,7 +90,9 @@ func AssertJSONCall(c *gc.C, p JSONCallParams) {
 
 // DoRequest invokes a request on the given handler with the given
 // method, URL, body, content length and headers.
-func DoRequest(c *gc.C, handler http.Handler, method string, urlStr string, body io.Reader, contentLength int64, header map[string][]string) *httptest.ResponseRecorder {
+func DoRequest(c *gc.C, handler http.Handler, method string, urlStr string, body io.Reader, contentLength int64, header map[string][]string, username, password string) *httptest.ResponseRecorder {
+	// TODO frankban: this function has too many arguments.
+	//   Use something like RequestParams.
```
rogpeppe, line 95: +1
I nearly did this anyway, recently. DoRequestParams would be the conventional name.

internal/charmstore/server.go
frankban: IIUC this is exactly what the code does. HandlerConfig is not the same structure used to deserialize the charmd configuration, and only includes stuff relevant for handlers (e.g. the mondo URL and the API address is not there).

internal/v4/auth.go
```diff
@@ -0,0 +1,65 @@
...
+	"strings"
+
+	"github.com/juju/errgo"
+	"gopkg.in/juju/charm.v3"
+
+	"github.com/juju/charmstore/internal/router"
+	"github.com/juju/charmstore/params"
+)
+
+const basicRealm = "CharmStore v4"
```
rogpeppe: Do we really want a space here?

internal/charmstore/server_test.go
jrwren, line 79: I should stop doing reviews before enough coffee in the morning. Thank for reminding me, @rogpeppe

jrwren:
:+1:

internal/v4/auth.go
```diff
@@ -0,0 +1,65 @@
...
+
+// authRequired decorates the given function so that an "unathorized"
+// response is returned if the HTTP basic auth in the request fails.
+// For the time being, authentication is done by simply checking that the
+// provided user/password match the superuser credentials stored in the
+// API handler.
+func (h *handler) authRequired(f router.IdHandler) router.IdHandler {
```
rogpeppe: I'm not sure there's a great advantage in phrasing this as a decorator - we mostly use decorators
for methods where we can generate the handler in advance, but we're not actually doing
that in any cases here.

I'd be inclined to add:

```
func (h *handler) authenticate(req *http.Request) error
```

and call it directly.

internal/v4/archive.go
```diff
@@ -39,13 +39,9 @@ func (h *handler) serveArchive(id *charm.Reference, w http.ResponseWriter, req *
 		// TODO(rog) params.ErrMethodNotAllowed
 		return errgo.Newf("method not allowed")
 	case "DELETE":
-		return h.serveDeleteArchive(id, w, req)
+		return h.authRequired(h.serveDeleteArchive)(id, w, req)
```
rogpeppe: As I've said later on in this review, I think it may be just as clear to do:

```
 if err := h.authenticate(req); err != nil {
     return err
 }
 return h.serveDeleteArchive(id, w, req)
```

and similarly for servePostArchive below. Slightly more
direct code, and you can leave the return type of servePostArchive
unchanged.

internal/v4/auth.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// parseCredentials parses the given request and returns the HTTP basic auth
+// credentials included in its header.
+func parseCredentials(req *http.Request) (username, password string, err error) {
+	parts := strings.Fields(req.Header.Get("Authorization"))
+	if len(parts) != 2 || parts[0] != "Basic" {
+		return "", "", errgo.New("invalid or missing HTTP auth header")
+	}
+	challenge, err := base64.StdEncoding.DecodeString(parts[1])
```
rogpeppe, line 41:
```
// Challenge is a base64-encoded "tag:pass" string.
// See RFC 2617, Section 2.
```

?

internal/charmstore/server.go
rogpeppe: Ah, thanks, you're right. It probably should not be defined inside params then, but inside internal/charmstore, as that's where the various handler types are defined.

server.go
```diff
@@ -33,9 +34,10 @@ func Versions() []string {
 	return vs
 }
 
-// NewServer returns a new handler that handles
-// charm store requests and stores its data in the given database.
-func NewServer(db *mgo.Database, serveVersions ...string) (http.Handler, error) {
+// NewServer returns a new handler that handles charm store requests and stores
+// its data in the given database. Each version handler is created providing
```
rogpeppe:
```
// NewServer returns a new handler that handles charm store requests and stores
// its data in the given database. The handler will serve the specified versions of
// the API using the given configuration.
```

I _think_ it's better have a separate type for the config that's specific to the top
level package, rather than putting HandlerConfig in params - params is
mostly about parameters to API calls rather than parameters to functions.

I'm open to arguments the other way though.

server_test.go
```diff
@@ -23,18 +23,27 @@ func TestPackage(t *testing.T) {
...
 type ServerSuite struct {
 	storetesting.IsolatedMgoSuite
+	config *params.HandlerConfig
 }
 
 var _ = gc.Suite(&ServerSuite{})
 
+func (s *ServerSuite) SetUpSuite(c *gc.C) {
+	s.IsolatedMgoSuite.SetUpSuite(c)
+	s.config = &params.HandlerConfig{
```
rogpeppe: Or just as a global variable?

internal/v4/archive_test.go
```diff
@@ -731,6 +755,64 @@ func (s *ArchiveSuite) TestDeleteCounters(c *gc.C) {
 	checkCounterSum(c, s.store, key, false, 1)
 }
 
+func (s *ArchiveSuite) TestDeleteAuthErrors(c *gc.C) {
+	checkAuthErrors(c, s.srv, "DELETE", "utopic/django-42/archive")
+}
+
```
rogpeppe, line 761: Perhaps the other call to checkAuthErrors (POST) could move here, so they're both together?

rogpeppe:
LGTM modulo some comments and suggestions above.

cmd/charmd/main.go
frankban: Done.

cmd/charmd/main.go
frankban: Done.

internal/charmstore/server.go
frankban: Not sure if I understand this. This is something we pass to an exposed API (NewServer).
Why do we want HandlerConfig to be internal?

internal/storetesting/http.go
frankban, line 95: Great!

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/160/

internal/v4/auth.go
frankban: I don't think it's a problem, but removed anyway.

internal/v4/auth.go
frankban: Well the way I see decorators is exactly for things like this. We don't want to repeat the same logic again and again and so we abstract out the generic part in a decorator form.
More below.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/161/

internal/v4/archive.go
frankban: It's more imperative but also more error prone. For instance, we need to remember to set the WWW-Authenticate header each time.
Moreover, I see value in having the same signature in the different serve*Archive methods.
I don't think the functional approach is less readable, the decorator itself is straightforward.
What do you think?

internal/v4/auth.go
frankban, line 41: Done.

server.go
frankban: Updated the docstring, thanks.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/162/

server.go
frankban: Re HandlerConfig, I am not sure I understood what you are suggesting. Using two different structures for NewServer and the underlying real handlers? About the placement, while it's true that HandlerConfig is not an HTTP API endpoint parameters, it's also true that it's still an external API parameter, not just a function argument. But I am sure I am missing your real point here.

server_test.go
frankban: Done.

internal/v4/archive_test.go
frankban, line 761: I grouped those by method (POST vs DELETE).
No problem in putting those together instead.
Fixed.

frankban:
Thanks for the reviews, replied to some of the comments.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/163/

internal/v4/archive.go
rogpeppe: Ah, I'd missed the WWW-Authenticate side-effect. But that's trivially dealt with
by passing w to h.authenticate.

I still think the non-callback approach is somewhat cleaner in this instance.
Having h.authenticate as a standalone method means that it's applicable
whatever the type of the handler (we might want authentication on methods
that don't refer to a charm/bundle id, for example).

Although I like higher order functions where they add real value,
they do make for slightly more obfuscated code - it we're just
saving an if statement, as in this case, I'm not sure it's worth it.

internal/charmstore/server.go
rogpeppe: I'm not sure that the type we pass into the exposed API should be the
same as that we use internally. I see why you've made it part of the
params package (because internal/charmstore can't depend on
charmstore itself), but perhaps it's better just to define the internal
and external parameter types separately. They may well acquire
different fields from each other in the future.

Also, perhaps better to pass around by value rather than by pointer,
so there's no ambiguity about whether it's required or not.

internal/v4/archive.go

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/164/

internal/charmstore/server.go
frankban: Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/165/

frankban:
:shipit:

jujugui:

PR #84, opened 2014-08-28
internal/router: implement PUT for meta endpoints

rogpeppe:
This also adds the notion of a multi-error, adding a new field to the Error
type.

### DIFF
--- a/internal/router/fieldinclude.go
+++ b/internal/router/fieldinclude.go
@@ -24,7 +24,7 @@ func (u *FieldUpdater) UpdateField(fieldName string, val interface{}) {
 }
 
 // A FieldUpdateFunc is used to update a metadata document for the
-// given URL. For each field in fields, it should set that field to
+// given id. For each field in fields, it should set that field to
 // its corresponding value in the metadata document.
 type FieldUpdateFunc func(id *charm.Reference, fields map[string]interface{}) error
 
@@ -33,8 +33,8 @@ type FieldUpdateFunc func(id *charm.Reference, fields map[string]interface{}) er
 // associated QueryFunc.
 type FieldGetFunc func(doc interface{}, id *charm.Reference, path string, flags url.Values) (interface{}, error)
 
-// FieldPutFunc sets values in updateOp corresponding to fields to be set
-// in the metadata document for the given URL. The path holds the metadata path
+// FieldPutFunc sets using the given FieldUpdater corresponding to fields to be set
+// in the metadata document for the given id. The path holds the metadata path
 // after the initial prefix has been removed.
 type FieldPutFunc func(id *charm.Reference, path string, val *json.RawMessage, updater *FieldUpdater) error
 
@@ -44,22 +44,25 @@ type FieldIncludeHandlerParams struct {
 	// (the same query should be generated for any given key).
 	Key interface{}
 
-	// Query is used to retrieve the document from the database
-	// The fields passed to the query will be the union of all fields found
-	// in all the handlers in the bulk request.
+	// Query is used to retrieve the document from the database for
+	// GET requests. The fields passed to the query will be the
+	// union of all fields found in all the handlers in the bulk
+	// request.
 	Query FieldQueryFunc
 
 	// Fields specifies which fields are required by the given handler.
 	Fields []string
 
-	// Handle actually retrieves the data from the document.
+	// Handle actually returns the data from the document retrieved
+	// by Query, for GET requests.
 	HandleGet FieldGetFunc
 
 	// HandlePut generates update operations for a PUT
 	// operation.
 	HandlePut FieldPutFunc
 
-	// Update is used to update the document in the database.
+	// Update is used to update the document in the database for
+	// PUT requests.
 	Update FieldUpdateFunc
 }
 
--- a/internal/router/fieldinclude.go
+++ b/internal/router/fieldinclude.go
@@ -1,6 +1,7 @@
 package router
 
 import (
+	"encoding/json"
 	"net/url"
 
 	"github.com/juju/errgo"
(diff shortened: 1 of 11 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/169/

internal/router/fieldinclude.go
```diff
@@ -11,56 +12,125 @@ import (
...
+}
+
+// UpdateField requests that the provided field is updated with
+// the given value.
+func (u *FieldUpdater) UpdateField(fieldName string, val interface{}) {
+	u.fields[fieldName] = val
+}
+
+// A FieldUpdateFunc is used to update a metadata document for the
+// given URL. For each field in fields, it should set that field to
```
frankban: given charm or bundle URL?

internal/router/fieldinclude.go
```diff
@@ -11,56 +12,125 @@ import (
...
+// its corresponding value in the metadata document.
+type FieldUpdateFunc func(id *charm.Reference, fields map[string]interface{}) error
+
+// A FieldGetFunc returns some data from the given document. The
 // document will have been returned from an earlier call to the
 // associated QueryFunc.
-type FieldHandlerFunc func(doc interface{}, id *charm.Reference, path string, method string, flags url.Values) (interface{}, error)
+type FieldGetFunc func(doc interface{}, id *charm.Reference, path string, flags url.Values) (interface{}, error)
+
+// FieldPutFunc sets values in updateOp corresponding to fields to be set
```
frankban: // FieldPutFunc sets values using the given FieldUpdater?
I am not sure about the meaning of updateOp here.

internal/router/fieldinclude.go
```diff
@@ -11,56 +12,125 @@ import (
...
+// in the metadata document for the given URL. The path holds the metadata path
+// after the initial prefix has been removed.
+type FieldPutFunc func(id *charm.Reference, path string, val *json.RawMessage, updater *FieldUpdater) error
+
+// FieldIncludeHandlerParams specifies the parameters for NewFieldIncludeHandler.
```
frankban, line 41: Nice to have params for this now.

internal/router/fieldinclude.go
```diff
@@ -11,56 +12,125 @@ import (
...
+type FieldIncludeHandlerParams struct {
+	// Key is used to group together similar FieldIncludeHandlers
+	// (the same query should be generated for any given key).
+	Key interface{}
+
+	// Query is used to retrieve the document from the database
+	// The fields passed to the query will be the union of all fields found
+	// in all the handlers in the bulk request.
+	Query FieldQueryFunc
```
frankban, line 51: I suppose Query and Fields are only used as part of the GET machinery, not the PUT one.
If that's the case, perhaps we should clarify that here.

internal/router/fieldinclude.go
```diff
@@ -11,56 +12,125 @@ import (
...
-	query  FieldQueryFunc
-	fields []string
-	handle FieldHandlerFunc
+func (h *fieldIncludeHandler) Key() interface{} {
+	return h.p.Key
 }
 
-func (h *fieldIncludeHandler) Key() interface{} {
-	return h.key
+func (h *fieldIncludeHandler) HandlePut(hs []BulkIncludeHandler, id *charm.Reference, paths []string, values []*json.RawMessage) []error {
```
frankban, line 86: Github code review is particularly bad for long lines :-/

internal/router/fieldinclude.go
```diff
@@ -11,56 +12,125 @@ import (
...
+	updater := &FieldUpdater{
+		fields: make(map[string]interface{}),
+	}
+	var errs []error
+	errCount := 0
+	setError := func(i int, err error) {
+		if errs == nil {
```
frankban, line 93: So in this way we create a slice of nil errors with the given length I presume.
Any reason to not doing this in advance? I guess the reason is that the contract for this function is to always return an empty slice if there is no error. I am ok with that (even if maybe "a slice of error results" is more easy to explain), but I'd clarify that in the BulkIncludeHandler interface decsription.

internal/router/fieldinclude.go
rogpeppe: Changed URL to id. The fact that it's a charm.Reference should make it clear that it's
a charm/bundle id.

internal/router/router.go
```diff
@@ -50,7 +52,7 @@ type BulkIncludeHandler interface {
 	// Please do not return NaN. That would be silly, OK?
 	Key() interface{}
 
-	// Handle returns the results of invoking all the given handlers
```
frankban, line 53: In the Key() docstring above, we should repace 
// together in preparation for a call to Handle.
with
// together in preparation for a call to Handle(Get|Put).

So bad github does not allow to put a comment in a non-modified line :rage:

internal/router/fieldinclude.go
rogpeppe: It's a relic from the past. Changed.

internal/router/router.go
```diff
@@ -63,7 +65,18 @@ type BulkIncludeHandler interface {
...
+
+	// HandlePut invokes a PUT request on all the given handlers on
+	// the given charm or bundle id. If there is an error, the
+	// returned errors slice should contain one element for each element
+	// in paths. The error for handler hs[i] should be returned in errors[i].
+	//
+	// Each item in paths holds the remaining metadata path
+	// for the handler in the corresponding position
+	// in hs after the prefix in Handlers.Meta has been stripped,
+	// and flags holds all the url query values.
```
frankban, line 79: the URL query values?

internal/router/fieldinclude.go
rogpeppe, line 51: done.

internal/router/router.go
```diff
@@ -223,7 +230,23 @@ func handlerKey(path string) (key, rest string) {
...
+	case "GET", "HEAD":
+		resp, err := r.serveMetaGet(id, req)
+		if err != nil {
+			// Note: preserve error causes from meta handlers.
+			return errgo.Mask(err, errgo.Any)
+		}
+		WriteJSON(w, http.StatusOK, resp)
+		return nil
+	case "PUT":
+		return r.serveMetaPut(id, req)
```
frankban, line 247: Are we sure we never need a JSON response for PUT requests?
If so, could you add a comment?

internal/router/fieldinclude.go
rogpeppe, line 93: The reason for not doing it in advance is to avoid unnecessary garbage
(and it means that checking whether there is an error is just a simple len check
rather than a scan through the slice.).

I've updated the BulkIncludeHandler docs to make the contract a little clearer.

internal/router/router.go
```diff
@@ -223,7 +230,23 @@ func handlerKey(path string) (key, rest string) {
...
+	}
+	return errgo.Newf("method %q not allowed", req.Method)
```
frankban: Is this a 500? Can we return a method not allowed status (or alternatively add a TODO and a card)?

internal/router/router.go
rogpeppe, line 53: Annoying, isn't it.
Fixed.

internal/router/router.go
rogpeppe, line 79: done

internal/router/router.go
```diff
@@ -280,12 +358,28 @@ func (r *Router) metaNames() []string {
...
+		resp, err := r.serveBulkMetaGet(req)
+		if err != nil {
+			return errgo.Mask(err, errgo.Any)
+		}
+		WriteJSON(w, http.StatusOK, resp)
+		return nil
+	case "PUT":
+		return r.serveBulkMetaPut(req)
+	default:
+		return fmt.Errorf("method not allowed")
```
frankban: Ditto.

internal/router/router.go
rogpeppe, line 247: I'm not sure forever, but it would be easy to change in the future,
and that's certainly the case now. Added a comment.

internal/router/router.go
```diff
@@ -258,6 +281,61 @@ func (r *Router) serveMeta(id *charm.Reference, req *http.Request) (interface{},
...
+		return errgo.WithCausef(nil, params.ErrBadRequest, "unexpected Content-Type %q; expected %q", ct, jsonContentType)
+	}
+	dec := json.NewDecoder(req.Body)
+	if err := dec.Decode(val); err != nil {
+		return errgo.Notef(err, "cannot unmarshal body")
+	}
+	return nil
+}
+
+func (r *Router) serveMetaPut(id *charm.Reference, req *http.Request) error {
```
frankban, line 303: We are starting to have lots of methods with similar names.
Some docstrings or an overview at the beginning of the module could help understanding what's going on when looking at the code in the future.

internal/router/router.go
rogpeppe: Changed so that it returns a 405.

internal/router/router.go
rogpeppe: done also.

internal/router/router_test.go
```diff
@@ -719,7 +711,620 @@ func (s *RouterSuite) TestRouterGet(c *gc.C) {
 	}
 }
 
-func (s *RouterSuite) TestRouterMethodsThatPassThroughUnresolvedId(c *gc.C) {
+var routerPutTests = []struct {
```
frankban, line 714: Very nice tests!

internal/router/router_test.go
```diff
@@ -1168,29 +1756,92 @@ func testIdHandler(charmId *charm.Reference, w http.ResponseWriter, req *http.Re
...
+func (testMetaHandler) HandlePut(hs []BulkIncludeHandler, id *charm.Reference, paths []string, rawValues []*json.RawMessage) []error {
+	// Handlers are provided in arbitrary order,
+	// so we order them (and their associated paths
+	// and values) to enable easier testing.
+	keys := make(sort.StringSlice, len(hs))
+	for i, h := range hs {
+		// Sort by handler primary, path secondary.
+		keys[i] = fmt.Sprintf("%d.%s", int(h.(testMetaHandler)), paths[i])
+	}
+	sort.Sort(groupSort{
```
frankban, line 1814: And here comes the magic...

internal/router/router_test.go
```diff
@@ -1209,22 +1860,91 @@ type fieldSelectQueryInfo struct {
...
+// ordered by their JSON serialization.
+func RecordedCalls() []interface{} {
+	callRecordsMutex.Lock()
+	defer callRecordsMutex.Unlock()
+
+	sort.Sort(callRecords)
+	return callRecords
+}
+
+type byJSON []interface{}
```
frankban, line 1920: Docstring? I guess this sorts elements by their JSON representation in bytes.

internal/router/router.go
rogpeppe, line 303: Added doc strings for all functions.

internal/router/router_test.go
```diff
@@ -1264,3 +2026,33 @@ func mustParseReference(url string) *charm.Reference {
...
+	Swap(i, j int)
+}
+
+type swapFunc func(i, j int)
+
+func (f swapFunc) Swap(i, j int) {
+	f(i, j)
+}
+
+type groupSort struct {
```
frankban, line 2046: So, does this keep in sync the key and the others while sorting?
Docstring?

internal/router/singleinclude.go
```diff
@@ -21,12 +22,12 @@ func (h SingleIncludeHandler) Key() interface{} {
 	return singleMetaHandlerKey(singleMetaHandlerKey{})
 }
 
-// Handle implements BulkMetadataHander.Handle.
-func (h SingleIncludeHandler) Handle(hs []BulkIncludeHandler, id *charm.Reference, paths []string, method string, flags url.Values) ([]interface{}, error) {
+// Handle implements BulkMetadataHander.HandleGet.
```
frankban: HandleGet

internal/router/singleinclude.go
```diff
@@ -35,3 +36,13 @@ func (h SingleIncludeHandler) Handle(hs []BulkIncludeHandler, id *charm.Referenc
 	}
 	return results, nil
 }
+
+var errPutNotImplemented = errgo.New("PUT not implemented")
+
+func (h SingleIncludeHandler) HandlePut(hs []BulkIncludeHandler, id *charm.Reference, paths []string, values []*json.RawMessage) []error {
```
frankban, line 43: // HandlePut implements BulkMetadataHander.HandlePut?

internal/router/util.go
```diff
@@ -5,6 +5,8 @@ package router
 
 import (
 	"encoding/json"
+	"fmt"
+	//	"log"
```
frankban: This comment can be removed.

internal/v4/api.go
```diff
@@ -246,19 +246,22 @@ func badRequestf(underlying error, f string, a ...interface{}) error {
 
 // GET id/meta/charm-metadata
 // http://tinyurl.com/poeoulw
-func (h *handler) metaCharmMetadata(entity *mongodoc.Entity, id *charm.Reference, path string, method string, flags url.Values) (interface{}, error) {
+func (h *handler) metaCharmMetadata(entity *mongodoc.Entity, id *charm.Reference, path string, flags url.Values) (interface{}, error) {
 	return entity.CharmMeta, nil
 }
 
+//func (h *handler) putMetaExtraInfo(id *charm.Reference, path string, value *json.RawMessage, flags url.Values, fieldToSet map[string] interface{}) error {
```
frankban: I think this is left in preparation for your next branch. [shrug]

internal/router/router_test.go
rogpeppe, line 1920: Done

internal/router/router_test.go
rogpeppe, line 2046: yes. done.

internal/router/singleinclude.go
rogpeppe: done.

internal/router/singleinclude.go
rogpeppe, line 43: done.

internal/router/util.go
rogpeppe: done.

frankban:
Wow, finally done! This branch is nice Roger, thank you!
:+1: with some minor requests.

internal/v4/api.go
rogpeppe: removed.

internal/router/fieldinclude.go
```diff
@@ -11,56 +12,125 @@ import (
...
+	// HandlePut generates update operations for a PUT
+	// operation.
+	HandlePut FieldPutFunc
+
+	// Update is used to update the document in the database.
+	Update FieldUpdateFunc
+}
+
+type fieldIncludeHandler struct {
+	p FieldIncludeHandlerParams
```
jrwren, line 70: I prefer to avoid 1 letter names in favor of something longer which helps me identify type, such as handlerParams. I find myself saying "what is p", because it could be anything. In this specific case, perhaps this could be an anonymous field?

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/170/

internal/router/fieldinclude.go
rogpeppe, line 70: I don't like to use anonymous fields where I'm not gaining particular benefit from
them, and "p" seems a reasonable name for the parameters to me - surely
some name is better than none, even if it's short?!

The reason for the short name is that it's used all over, and as there's only one
field, the field name adds weight but not much value. Hopefully the code
is clear as soon as you've looked at this type definition...

internal/router/router_test.go
jrwren, line 714: A 513 line composite literal!

internal/router/router_test.go
```diff
@@ -719,7 +711,620 @@ func (s *RouterSuite) TestRouterGet(c *gc.C) {
...
+}}
+
+func nopUpdate(id *charm.Reference, fields map[string]interface{}) error {
+	return nil
+}
+
+func (s *RouterSuite) TestRouterPut(c *gc.C) {
+	for i, test := range routerPutTests {
+		c.Logf("test %d: %s", i, test.about)
+		ResetRecordedCalls()
```
jrwren, line 1236: Why the use of the global RecordedCalls? I feel this could be scoped to this method alone.

jrwren:
:+1:  My comments are just curiosity questions.

internal/router/fieldinclude.go
jrwren, line 70: I can disagree :)   I prefer no name to a single char name.

Being used all over is all the more reason for it to be descriptive.

It is clear once I've looked at the type definition.

I do feel it is less important given that it is not an exported type.

rogpeppe:
:shipit:

jujugui:

PR #85, opened 2014-08-29
Fix intermittent test failures due to results ordering issues.

frankban:
Make bundles-containing and charm-related API
endpoints return results in a specific order.

### DIFF
--- a/internal/v4/archive.go
+++ b/internal/v4/archive.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"path"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -324,6 +325,7 @@ func verificationError(err error) error {
 	for i, err := range verr.Errors {
 		messages[i] = err.Error()
 	}
+	sort.Strings(messages)
 	encodedMessages, err := json.Marshal(messages)
 	if err != nil {
 		// This should never happen.
--- a/internal/v4/archive.go
+++ b/internal/v4/archive.go
@@ -134,6 +134,7 @@ func (h *handler) servePostArchive(id *charm.Reference, w http.ResponseWriter, r
 			return errgo.Notef(err, "cannot retrieve bundle charms")
 		}
 		if err := bundleData.VerifyWithCharms(verifyConstraints, charms); err != nil {
+			// TODO frankban: use multiError (defined in internal/router).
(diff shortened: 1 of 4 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/172/

internal/v4/archive.go
```diff
@@ -324,6 +325,7 @@ func verificationError(err error) error {
 	for i, err := range verr.Errors {
 		messages[i] = err.Error()
 	}
+	sort.Strings(messages)
```
rogpeppe, line 329: It occurs to me that now we've got multiError, we could probably use it here.
We're probably still need to the sorting to get consistent keys though.

This could probably do with a doc comment anyway.

internal/v4/relations.go
```diff
@@ -187,18 +189,19 @@ func (h *handler) metaBundlesContaining(entity *mongodoc.Entity, id *charm.Refer
 	}
 
 	// Prepare and return the response.
-	response := make([]*params.MetaAnyResponse, 0, len(entities))
```
rogpeppe, line 190: Why the change from pointer to value?

internal/v4/relations.go
```diff
@@ -134,6 +135,7 @@ func (h *handler) getRelatedIfaceResponses(
 			}
 		}
 	}
+	sort.Sort(byId(responses))
```
rogpeppe: I'm not sure we need to sort the responses if the entities are already sorted,
which could probably be done by mongodb server side. Do we mind
that much about numeric revision ordering?

internal/v4/archive.go
frankban, line 329: Done.

internal/v4/relations.go
frankban: Ok, fixed.

internal/v4/relations.go
frankban, line 190: Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/173/

urosj:
:+1:

frankban:
Thanks for the reviews!
:shipit:

jujugui:

PR #86, opened 2014-08-29
Hook up archive upload related stats counters.

frankban:


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/175/

internal/v4/stats.go
```diff
@@ -26,7 +26,10 @@ import (
 //   - kind:trusty:django:42:who -> a specific user charm.
 // The above also applies to bundles (where the series is "bundle").
 func entityStatsKey(url *charm.Reference, kind string) []string {
-	key := []string{kind, url.Series, url.Name, strconv.Itoa(url.Revision)}
+	key := []string{kind, url.Series, url.Name}
+	if url.Revision != -1 {
```
rogpeppe, line 35: I'm not sure about this. As both revision and key are optional, we are left with
a situation where we can't know whether the 4th part of the key is a revision
number or a user. It seems a bit odd that user isn't before the charm name
actually. I wonder what the motivation for that was - it means that ~rogpeppe/precise/wordpress
counts towards global wordpress stats, AFAICS. Let's discuss this.

frankban, line 35: Good point. I modified how stats keys are generated, and it seems that the current approach is able to answer a wider number of questions.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/177/

internal/v4/archive_test.go
```diff
@@ -93,7 +93,7 @@ func (s *ArchiveSuite) TestGetCounters(c *gc.C) {
 	c.Assert(rec.Code, gc.Equals, http.StatusOK)
 
 	// Check that the downloads count for the entity has been updated.
-	key := []string{params.StatsArchiveDownload, "utopic", "mysql", "42"}
+	key := []string{params.StatsArchiveDownload, "utopic", "mysql", "", "42"}
```
urosj: Could we add a test with a user as well?

urosj:
:+1: 
Would add a test that includes a username in the counter key.

jrwren:
:+1:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/180/

internal/v4/archive_test.go
frankban: Done.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

PR #88, opened 2014-09-02
Implement extra-info GET and PUT API endpoints.

frankban:
This includes both extra-info and extra-info with key.
Also improve how not found errors are handled by the router and the endpoints.

This is mostly based on Roger's work. I just fixed the failing tests and other minors.

### DIFF
--- a/internal/v4/api.go
+++ b/internal/v4/api.go
@@ -5,9 +5,11 @@ package v4
 
 import (
 	"archive/zip"
+	"encoding/json"
 	"net/http"
 	"net/url"
 	"sort"
+	"strings"
 
 	"github.com/juju/errgo"
 	"gopkg.in/juju/charm.v3"
@@ -58,11 +60,19 @@ func New(store *charmstore.Store, config charmstore.ServerParams) http.Handler {
 			"manifest":            h.entityHandler(h.metaManifest, "blobname"),
 			"revision-info":       router.SingleIncludeHandler(h.metaRevisionInfo),
 			"stats":               h.entityHandler(h.metaStats),
+			"extra-info": h.puttableEntityHandler(
+				h.metaExtraInfo,
+				h.putMetaExtraInfo,
+				"extrainfo",
+			),
+			"extra-info/": h.puttableEntityHandler(
+				h.metaExtraInfoWithKey,
+				h.putMetaExtraInfoWithKey,
+				"extrainfo",
+			),
 
 			// endpoints not yet implemented - use SingleIncludeHandler for the time being.
-			"color":       router.SingleIncludeHandler(h.metaColor),
-			"extra-info":  router.SingleIncludeHandler(h.metaExtraInfo),
-			"extra-info/": router.SingleIncludeHandler(h.metaExtraInfoWithKey),
+			"color": router.SingleIncludeHandler(h.metaColor),
 		},
 	}, h.resolveURL)
 	return h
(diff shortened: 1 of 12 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/178/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/179/

jrwren:
:+1:

internal/v4/api.go
```diff
@@ -103,9 +113,13 @@ type entityHandlerFunc func(entity *mongodoc.Entity, id *charm.Reference, path s
 // entityHandler returns a handler that calls f with a *mongodoc.Entity that
 // contains at least the given fields. It allows only GET requests.
 func (h *handler) entityHandler(f entityHandlerFunc, fields ...string) router.BulkIncludeHandler {
+	return h.puttableEntityHandler(f, nil, fields...)
+}
+
+func (h *handler) puttableEntityHandler(get entityHandlerFunc, put router.FieldPutFunc, fields ...string) router.BulkIncludeHandler {
```
urosj: The name is misleading as the "puttable" handler deals with GET calls as well. And when we add delete (or even post) calls, it makes even less sense. What about putGetEntityHandlers?

internal/v4/api.go
```diff
@@ -114,9 +128,19 @@ func (h *handler) entityHandler(f entityHandlerFunc, fields ...string) router.Bu
 		Query:     h.entityQuery,
 		Fields:    fields,
 		HandleGet: handleGet,
+		HandlePut: put,
```
urosj: What about handlePut (to be in same line as handleGet above) instead?

internal/v4/api.go
```diff
@@ -373,18 +397,60 @@ func (s entitiesByRevision) Less(i, j int) bool { return s[i].URL.Revision > s[j
...
-type FieldUpdater interface {
-	UpdateField(name string, val interface{})
+func (h *handler) metaExtraInfo(entity *mongodoc.Entity, id *charm.Reference, path string, flags url.Values) (interface{}, error) {
+	// The extra-info is stored in mongo as simple byte
+	// slices, so convert the values to json.RawMessages
+	// so that the client will see the original JSON.
+	m := make(map[string]*json.RawMessage)
+	for key, val := range entity.ExtraInfo {
+		jmsg := json.RawMessage(val)
+		m[key] = &jmsg
```
urosj, line 407: So we can actually add anything to it, like /extra-info/someImage and a PNG with it? The comment is mostly for put.

internal/v4/api.go
frankban: I think the reason here is that we have an entityHandler that is used for GET requests, and a puttableEntityHandler for both GET and PUT. I am not sure about changing the latter without changing the former. I am open to suggestions. putGetEntityHandler does not look so good to me, because it's not generic: as you mentioned, if we introduce delete then we need to change the name of this function all over the code again. Ideas? perhaps s/puttableEntityHandler/entityHandler/ and s/entityHandler/entityHandlerGetOnly/ ?
I don't have a strong opinion BTW, the current also works for me.

internal/v4/api.go
frankban: I am not sure I understood? Are you suggesting to just put HandleGet and HandlePut in the same line? In that case, I think it's idiomatic code style to just put each field in a sperate line.

urosj: Just a rename of that put variable. putFunc sounds more readable to me, as the real put magic happens there. Just "put" makes it so insignificant :)

internal/v4/api.go
urosj: ok, let's leave it with puttable.

internal/v4/api.go
frankban, line 407: Yeah, we can add any values in JSON format.

internal/v4/api.go
frankban: OIC, I misinterpreted your first comment. Sounds good.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/182/

internal/v4/api_test.go
```diff
@@ -784,6 +926,38 @@ func (s *APISuite) addBundle(c *gc.C, bundleName string, curl string) (*charm.Re
...
+		Method:  "PUT",
+		Header: http.Header{
+			"Content-Type": {"application/json"},
+		},
+		Body: bytes.NewReader(body),
+	})
+	c.Assert(rec.Code, gc.Equals, http.StatusOK, gc.Commentf("body: %s", rec.Body.String()))
+	c.Assert(rec.Body.String(), gc.HasLen, 0)
+}
+
```
urosj, line 944: Nice.

urosj:
:+1:  Nice!

frankban:
Thank you!
:shipit:

jujugui:

PR #89, opened 2014-09-02
implement get/put on meta/color

jrwren:
This is just the get/put implementation. The UX algorithm to extract a color on archive post is separate.

### DIFF
--- a/internal/v4/api.go
+++ b/internal/v4/api.go
@@ -57,10 +57,13 @@ func New(store *charmstore.Store, config charmstore.ServerParams) http.Handler {
 			"charm-config":        h.entityHandler(h.metaCharmConfig, "charmconfig"),
 			"charm-metadata":      h.entityHandler(h.metaCharmMetadata, "charmmeta"),
 			"charm-related":       h.entityHandler(h.metaCharmRelated, "charmprovidedinterfaces", "charmrequiredinterfaces"),
-			"color":               h.entityHandler(h.metaColor, "iconbackgroundcolor"),
-			"manifest":            h.entityHandler(h.metaManifest, "blobname"),
-			"revision-info":       router.SingleIncludeHandler(h.metaRevisionInfo),
-			"stats":               h.entityHandler(h.metaStats),
+			"color": h.puttableEntityHandler(
+				h.metaColor,
+				h.putMetaColor,
+				"iconbackgroundcolor"),
+			"manifest":      h.entityHandler(h.metaManifest, "blobname"),
+			"revision-info": router.SingleIncludeHandler(h.metaRevisionInfo),
+			"stats":         h.entityHandler(h.metaStats),
 			"extra-info": h.puttableEntityHandler(
 				h.metaExtraInfo,
 				h.putMetaExtraInfo,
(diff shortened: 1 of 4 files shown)


jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/184/

internal/v4/api.go
```diff
@@ -321,8 +322,19 @@ func (h *handler) metaCharmConfig(entity *mongodoc.Entity, id *charm.Reference,
 
 // GET id/meta/color
 // http://tinyurl.com/o2t3j4p
-func (h *handler) metaColor(id *charm.Reference, path string, flags url.Values) (interface{}, error) {
-	return nil, errNotImplemented
+func (h *handler) metaColor(entity *mongodoc.Entity, id *charm.Reference, path string, flags url.Values) (interface{}, error) {
+	return &params.ColorResponse{entity.IconBackgroundColor}, nil
```
urosj, line 326: Shouldn't we check if the colour is set and if not, return NotFound error? Otherwise an empty string is returned which is not a valid colour.

internal/v4/api_test.go
```diff
@@ -847,6 +847,176 @@ func (s *APISuite) TestMetaStats(c *gc.C) {
...
+
+func (s *APISuite) TestServeMetaColor(c *gc.C) {
+	id, _ := s.addCharm(c, "wordpress", "cs:trusty/wordpress-42")
+	var val mongodoc.Entity
+	s.store.DB.Entities().
+		Find(bson.D{{"_id", id}}).
+		One(&val)
+	val.IconBackgroundColor = "abcdef"
+	s.store.DB.Entities().Update(bson.D{{"_id", id}}, val)
+
```
urosj, line 877: Content type not set?

internal/v4/api_test.go
```diff
@@ -847,6 +847,176 @@ func (s *APISuite) TestMetaStats(c *gc.C) {
...
+	err:   "no matching charm or bundle for cs:precise/no-such-33",
+}, {
+	about:  "partial url uses a default series",
+	url:    "wordpress",
+	expect: "abcdef",
+}, {
+	about:  "fully qualified url",
+	url:    "trusty/wordpress-42",
+	expect: "abcdef",
+}}
```
urosj, line 867: I wouldn't use table tests here. Just simple put/get combinations with assertPut and assertGet.

internal/v4/api_test.go
```diff
@@ -847,6 +847,176 @@ func (s *APISuite) TestMetaStats(c *gc.C) {
...
+		c.Assert(gotBodyVal, jc.DeepEquals, expectBodyVal)
+	}
+}
+
+func (s *APISuite) TestColor(c *gc.C) {
+	id := "precise/wordpress-23"
+	s.addCharm(c, "wordpress", id)
+	s.assertPut(c, id+"/meta/color", map[string]string{"RGB": "123456"})
+	s.assertGet(c, id+"/meta/color", map[string]string{"RGB": "123456"})
+	s.assertPut(c, id+"/meta/any", params.MetaAnyResponse{
```
urosj, line 1008: I would suggest to have only these and simple bad request tests when nothing is set and when json used in PUT to color is not correct.

PR #90, opened 2014-09-03
Add units count extra info to bundles.

frankban:
This is an example of an extra info added at publish time.
Created a card for machines count (in TODO).

### DIFF
--- a/internal/charmstore/extrainfo.go
+++ b/internal/charmstore/extrainfo.go
@@ -1,0 +1,28 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package charmstore
+
+import (
+	"encoding/json"
+
+	"github.com/juju/errgo"
+	"gopkg.in/juju/charm.v3"
+)
+
+func bundleExtraInfo(data *charm.BundleData) (map[string][]byte, error) {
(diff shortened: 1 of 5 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/185/

urosj:
:+1:  Thank you.

internal/charmstore/extrainfo.go
```diff
@@ -0,0 +1,28 @@
...
+package charmstore
+
+import (
+	"encoding/json"
+
+	"github.com/juju/errgo"
+	"gopkg.in/juju/charm.v3"
+)
+
+func bundleExtraInfo(data *charm.BundleData) (map[string][]byte, error) {
```
jrwren, line 13: bundleUnitCounts may be a better name for this function as it doesn't get all the extra info, just the unit counts.

jrwren:
:+1:

internal/charmstore/extrainfo.go
frankban, line 13: But it is supposed to return more data (machines count) in the future, and it actually returns an extra info map rather than a number of units.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

PR #92, opened 2014-09-05
Store the machines count in the bundle extra info.

frankban:


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/194/

jrwren:
:+1:

internal/charmstore/extrainfo_test.go
```diff
@@ -127,6 +128,321 @@ func (s *extraInfoSuite) TestUnitsCount(c *gc.C) {
...
+			"django": {
+				Charm:    "cs:utopic/django-0",
+				NumUnits: 0,
+			},
+			"haproxy": {
+				Charm:    "cs:trusty/haproxy-0",
+				NumUnits: 0,
+			},
+		},
+	},
```
urosj, line 151: add expectMachines here as well?

internal/charmstore/extrainfo_test.go
```diff
@@ -127,6 +128,321 @@ func (s *extraInfoSuite) TestUnitsCount(c *gc.C) {
...
+	data: &charm.BundleData{
+		Services: map[string]*charm.ServiceSpec{
+			"django": {
+				Charm:    "cs:trusty/django-42",
+				NumUnits: 1,
+			},
+			"haproxy": {
+				Charm:    "cs:trusty/haproxy-47",
+				NumUnits: 1,
+				To:       []string{"django/1"},
```
urosj: Is it django/0 or django/1? I also tried the juju --to with such parameters and got invalid --to parameter response.

internal/charmstore/extrainfo_test.go
```diff
@@ -127,6 +128,321 @@ func (s *extraInfoSuite) TestUnitsCount(c *gc.C) {
...
+				Charm:    "cs:utopic/postgres-3",
+				NumUnits: 1,
+				To:       []string{"kvm:2"},
+			},
+		},
+		Machines: map[string]*charm.MachineSpec{
+			"1": nil, "2": nil, "3": nil,
+		},
+	},
+	expectMachines: 2 + 0 + 1,
```
urosj: It's actually 2 + 1 + 0

internal/charmstore/extrainfo_test.go
```diff
@@ -127,6 +128,321 @@ func (s *extraInfoSuite) TestUnitsCount(c *gc.C) {
...
+				Charm:    "cs:trusty/django-42",
+				NumUnits: 1,
+				To:       []string{"1"},
+			},
+		},
+		Machines: map[string]*charm.MachineSpec{
+			"1": nil,
+		},
+	},
+	expectMachines: 1,
```
urosj, line 181: What about adding a test with - NumUnits: 1, to: []string{"1", "2", "3"} ... with wrong expectMachines: 3, but should be 1.

internal/charmstore/extrainfo.go
```diff
@@ -11,10 +11,39 @@ import (
 )
 
 func bundleExtraInfo(data *charm.BundleData) (map[string][]byte, error) {
-	// Calculate the number of units in the bundle.
+	// Calculate the number of units and machines in the bundle.
 	numUnits := 0
+	numMachines := len(data.Machines)
```
urosj, line 16: Number of machines should not be bigger than number of units. See test comment below.

internal/charmstore/extrainfo_test.go
frankban, line 151: This test exercises the number of units, machinesCount has its own test below, and I'd prefer to keep tests separate, for future sanity ;-)

internal/charmstore/extrainfo_test.go
frankban: django/0 is more correct indeed, good catch, changed.
This co-location pattern only work in a bundle deployment context (and right now the only deployer implementation we have is juju-deployer). So this is not supposed to work now, but this needs to work when bundle deployment will be implemented in juju-core.

internal/charmstore/extrainfo_test.go
frankban: Done.

internal/charmstore/extrainfo_test.go
frankban: For more on this, see https://github.com/juju/charm/blob/v3/bundledata.go#L69

internal/charmstore/extrainfo_test.go
frankban, line 181: A bundle like that would not pass verification (and therefore it would not reach this code path).
For this reason, the bundleExtraInfo function safely assumes the bundle is validated, and such a case is not possible (so no need to be tested).
See https://github.com/juju/charm/blob/v3/bundledata.go#L289

internal/charmstore/extrainfo.go
frankban, line 16: As mentioned, bundle validation ensure this is the case.
In this specific line we store the number of declared machines.
Since a bundle cannot declare machines it does not use, it sounds like a good initial value.

internal/charmstore/extrainfo_test.go
urosj, line 181: Great, tnx!

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/195/

urosj:
:+1: Great!

frankban:
:shipit:

jujugui:

PR #95, opened 2014-09-08
Launchpad ingestion to archive endpoint

jrwren:
Wire up launchpad ingestion to charmstore filling out new api/models.

### DIFF
--- a/cmd/charmload/main.go
+++ b/cmd/charmload/main.go
@@ -93,16 +93,7 @@ func load() error {
 		}
 		URLs := []*charm.URL{charmURL}
 		schema, name := charmURL.Schema, charmURL.Name
-		for _, series := range tip.OfficialSeries {
-			nextCharmURL := &charm.URL{
-				Schema:   schema,
-				Name:     name,
-				Revision: -1,
-				Series:   series,
-			}
-			URLs = append(URLs, nextCharmURL)
-			logger.Debugf("added URL %v to URLs list for %v", nextCharmURL, tip.UniqueName)
-		}
+		addPromulgatedCharmURLs(tip.OfficialSeries, schema, name, URLs)
 		err = publishBazaarBranch(*storeURL, *storeUser, URLs, branchURL, tip.Revision)
 		if err != nil {
 			logger.Errorf("publishing branch %v to charmstore: %v", branchURL, err)
@@ -113,6 +104,22 @@ func load() error {
 
 	}
 	return nil
+}
+
+// addPromulgatedCharmURLs adds urls from officialSeries to
+// the URLs slice for the given schema, name.
+// Promulgated charms have OfficialSeries in launchpad.
+func addPromulgatedCharmURLs(officialSeries []string, schema, name string, URLs []*charm.URL) {
+	for _, series := range officialSeries {
+		nextCharmURL := &charm.URL{
+			Schema:   schema,
+			Name:     name,
+			Revision: -1,
+			Series:   series,
+		}
+		URLs = append(URLs, nextCharmURL)
+		logger.Debugf("added URL %v to URLs list for %v", nextCharmURL, URLs[0])
+	}
 }
 
 // uniqueNameURLs returns the branch URL and the charm URL for the
@@ -143,7 +150,6 @@ func uniqueNameURLs(name string) (branchURL string, charmURL *charm.URL, err err
 }
 
 func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, branchURL string, digest string) error {
-
 	// Retrieve the branch with a lightweight checkout, so that it
 	// builds a working tree as cheaply as possible. History
 	// doesn't matter here.
--- a/cmd/charmload/main.go
+++ b/cmd/charmload/main.go
@@ -127,7 +127,7 @@ func uniqueNameURLs(name string) (branchURL string, charmURL *charm.URL, err err
 		branchURL = "lp:" + name
 	}
 	if len(u) < 5 || u[1] != "charms" || u[4] != "trunk" || len(u[0]) == 0 || u[0][0] != '~' {
-		return "", nil, fmt.Errorf("unwanted branch name: %s", name)
+		return "", nil, fmt.Errorf("unsupported branch name: %s", name)
 	}
 	charmURL, err = charm.ParseURL(fmt.Sprintf("cs:%s/%s/%s", u[0], u[2], u[3]))
 	if err != nil {
--- a/cmd/charmload/main.go
+++ b/cmd/charmload/main.go
@@ -35,9 +35,9 @@ func main() {
 func load() error {
 	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	staging := flags.Bool("staging", false, "use the launchpad staging server")
-	storeurl := flags.String("storeurl", "http://localhost:8080/v4/", "the url of the charmstore")
+	storeURL := flags.String("storeurl", "http://localhost:8080/v4/", "the URL of the charmstore")
 	loggingConfig := flags.String("logging-config", "", "specify log levels for modules")
-	showlog := flags.Bool("show-log", false, "if set, write log messages to stderr")
+	showLog := flags.Bool("show-log", false, "if set, write log messages to stderr")
 	storeUser := flags.String("user", "admin:example-passwd", "the colon separated user:password for charmstore")
 	err := flags.Parse(os.Args[1:])
 	if flag.ErrHelp == err {
@@ -50,7 +50,7 @@ func load() error {
 	if *loggingConfig != "" {
 		loggo.ConfigureLoggers(*loggingConfig)
 	}
-	if *showlog {
+	if *showLog {
 		writer := loggo.NewSimpleWriter(os.Stderr, &loggo.DefaultFormatter{})
 		_, err := loggo.ReplaceDefaultWriter(writer)
 		if err != nil {
@@ -76,7 +76,7 @@ func load() error {
 			continue
 		}
 		logger.Tracef("getting uniqueNameURLs for %v", tip.UniqueName)
-		branchurl, charmurl, err := uniqueNameURLs(tip.UniqueName)
+		branchURL, charmURL, err := uniqueNameURLs(tip.UniqueName)
 		if err != nil {
 			logger.Infof("could not get uniqueNameURLs for %v: %v", tip.UniqueName, err)
 			continue
@@ -85,21 +85,21 @@ func load() error {
 			logger.Tracef("skipping %v no revision", tip.UniqueName)
 			continue
 		}
-		urls := []*charm.URL{charmurl}
-		schema, name := charmurl.Schema, charmurl.Name
+		URLs := []*charm.URL{charmURL}
+		schema, name := charmURL.Schema, charmURL.Name
 		for _, series := range tip.OfficialSeries {
-			nextcharmurl := &charm.URL{
+			nextCharmURL := &charm.URL{
 				Schema:   schema,
 				Name:     name,
 				Revision: -1,
 				Series:   series,
 			}
-			urls = append(urls, nextcharmurl)
-			logger.Debugf("added url %v to urls list for %v", nextcharmurl, tip.UniqueName)
+			URLs = append(URLs, nextCharmURL)
+			logger.Debugf("added URL %v to URLs list for %v", nextCharmURL, tip.UniqueName)
 		}
-		err = publishBazaarBranch(*storeurl, *storeUser, urls, branchurl, tip.Revision)
+		err = publishBazaarBranch(*storeURL, *storeUser, URLs, branchURL, tip.Revision)
 		if err != nil {
-			logger.Errorf("publishing branch %v to charmstore: %v", branchurl, err)
+			logger.Errorf("publishing branch %v to charmstore: %v", branchURL, err)
 		}
 		if _, ok := err.(*UnauthorizedError); ok {
 			return err
@@ -118,25 +118,25 @@ func load() error {
 // For testing purposes, if name has a prefix preceding a string in
 // this format, the prefix is stripped out for computing the charm
 // URL, and the unique name is returned unchanged as the branch URL.
-func uniqueNameURLs(name string) (burl string, charmurl *charm.URL, err error) {
+func uniqueNameURLs(name string) (branchURL string, charmURL *charm.URL, err error) {
 	u := strings.Split(name, "/")
 	if len(u) > 5 {
 		u = u[len(u)-5:]
-		burl = name
+		branchURL = name
 	} else {
-		burl = "lp:" + name
+		branchURL = "lp:" + name
 	}
 	if len(u) < 5 || u[1] != "charms" || u[4] != "trunk" || len(u[0]) == 0 || u[0][0] != '~' {
 		return "", nil, fmt.Errorf("unwanted branch name: %s", name)
 	}
-	charmurl, err = charm.ParseURL(fmt.Sprintf("cs:%s/%s/%s", u[0], u[2], u[3]))
+	charmURL, err = charm.ParseURL(fmt.Sprintf("cs:%s/%s/%s", u[0], u[2], u[3]))
 	if err != nil {
 		return "", nil, err
 	}
-	return burl, charmurl, nil
+	return branchURL, charmURL, nil
 }
 
-func publishBazaarBranch(storeurl string, storeUser string, urls []*charm.URL, branchurl string, digest string) error {
+func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, branchURL string, digest string) error {
 
 	var branchDir string
 NewTip:
@@ -151,8 +151,8 @@ NewTip:
 		}
 		defer os.RemoveAll(tempDir)
 		branchDir = filepath.Join(tempDir, "branch")
-		logger.Debugf("running bzr checkout ... %v", branchurl)
-		output, err := exec.Command("bzr", "checkout", "--lightweight", branchurl, branchDir).CombinedOutput()
+		logger.Debugf("running bzr checkout ... %v", branchURL)
+		output, err := exec.Command("bzr", "checkout", "--lightweight", branchURL, branchDir).CombinedOutput()
 		if err != nil {
 			return outputErr(output, err)
 		}
@@ -187,10 +187,10 @@ NewTip:
 			thischarm.ArchiveTo(writer)
 			writer.Close()
 		}()
-		id := urls[0]
-		url := storeurl + id.Path() + "/archive?hash=" + hash1str
-		logger.Infof("posting to %v", url)
-		request, err := http.NewRequest("POST", url, reader)
+		id := URLs[0]
+		URL := storeURL + id.Path() + "/archive?hash=" + hash1str
+		logger.Infof("posting to %v", URL)
+		request, err := http.NewRequest("POST", URL, reader)
 		authhash := base64.StdEncoding.EncodeToString([]byte(storeUser))
 		logger.Tracef("encoded Authorization %v", authhash)
 		request.Header["Authorization"] = []string{"Basic " + authhash}
--- a/cmd/charmload/main.go
+++ b/cmd/charmload/main.go
@@ -174,43 +174,44 @@ func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, b
 		logger.Warningf("tipDigest %v != digest %v", digest, tipDigest)
 	}
 
-	thischarm, err := charm.ReadCharmDir(branchDir)
-	logger.Tracef("read CharmDir from branchDir %v", thischarm, branchDir)
-	if err == nil {
-		reader, writer := io.Pipe()
-		hash1 := sha512.New384()
-		var counter Counter
-		mwriter := io.MultiWriter(hash1, &counter)
-		thischarm.ArchiveTo(mwriter)
-		hash1str := fmt.Sprintf("%x", hash1.Sum(nil))
-		go func() {
-			thischarm.ArchiveTo(writer)
-			writer.Close()
-		}()
-		id := URLs[0]
-		URL := storeURL + id.Path() + "/archive?hash=" + hash1str
-		logger.Infof("posting to %v", URL)
-		request, err := http.NewRequest("POST", URL, reader)
-		authhash := base64.StdEncoding.EncodeToString([]byte(storeUser))
-		logger.Tracef("encoded Authorization %v", authhash)
-		request.Header["Authorization"] = []string{"Basic " + authhash}
-		// go1.2.1 has a bug requiring Content-Type to be sent
-		// since we are posting to a go server which may be running on
-		// 1.2.1, we should send this header
-		// https://code.google.com/p/go/source/detail?r=a768c0592b88
-		request.Header["Content-Type"] = []string{"application/octet-stream"}
-		request.ContentLength = int64(counter)
-		resp, err := http.DefaultClient.Do(request)
-		if resp.StatusCode == http.StatusUnauthorized {
-			logger.Errorf("invalid charmstore credentials")
-			return &UnauthorizedError{}
-		}
-		if err != nil || resp.StatusCode != http.StatusOK {
-			logger.Warningf("error posting:", err, resp.Header)
-			io.Copy(os.Stdout, resp.Body)
-		}
-		logger.Tracef("response: %v", resp)
+	logger.Tracef("read CharmDir from branchDir %v", branchDir)
+	thisCharm, err := charm.ReadCharmDir(branchDir)
+	if err != nil {
+		return err
 	}
+	reader, writer := io.Pipe()
+	hash1 := sha512.New384()
+	var counter Counter
+	mwriter := io.MultiWriter(hash1, &counter)
+	thisCharm.ArchiveTo(mwriter)
+	hash1str := fmt.Sprintf("%x", hash1.Sum(nil))
+	go func() {
+		thisCharm.ArchiveTo(writer)
+		writer.Close()
+	}()
+	id := URLs[0]
+	URL := storeURL + id.Path() + "/archive?hash=" + hash1str
+	logger.Infof("posting to %v", URL)
+	request, err := http.NewRequest("POST", URL, reader)
+	authhash := base64.StdEncoding.EncodeToString([]byte(storeUser))
+	logger.Tracef("encoded Authorization %v", authhash)
+	request.Header["Authorization"] = []string{"Basic " + authhash}
+	// go1.2.1 has a bug requiring Content-Type to be sent
+	// since we are posting to a go server which may be running on
+	// 1.2.1, we should send this header
+	// https://code.google.com/p/go/source/detail?r=a768c0592b88
+	request.Header["Content-Type"] = []string{"application/octet-stream"}
+	request.ContentLength = int64(counter)
+	resp, err := http.DefaultClient.Do(request)
+	if resp.StatusCode == http.StatusUnauthorized {
+		logger.Errorf("invalid charmstore credentials")
+		return &UnauthorizedError{}
+	}
+	if err != nil || resp.StatusCode != http.StatusOK {
+		logger.Warningf("error posting:", err, resp.Header)
+		io.Copy(os.Stdout, resp.Body)
+	}
+	logger.Tracef("response: %v", resp)
 
 	return err
 }
--- a/cmd/charmload/main.go
+++ b/cmd/charmload/main.go
@@ -13,9 +13,11 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
+	"github.com/juju/errgo"
 	"github.com/juju/loggo"
 	"gopkg.in/juju/charm.v3"
 
@@ -42,7 +44,7 @@ func load() error {
 	storeURL := flags.String("storeurl", "http://localhost:8080/v4/", "the URL of the charmstore")
 	loggingConfig := flags.String("logging-config", "", "specify log levels for modules e.g. <root>=TRACE")
 	showLog := flags.Bool("show-log", false, "if set, write log messages to stderr")
-	storeUser := flags.String("user", "admin:example-passwd", "the colon separated user:password for charmstore")
+	storeUser := flags.String("u", "admin:example-passwd", "the colon separated user:password for charmstore")
 	err := flags.Parse(os.Args[1:])
 	if flag.ErrHelp == err {
 		flag.Usage()
@@ -54,9 +56,8 @@ func load() error {
 	if *loggingConfig != "" {
 		loggo.ConfigureLoggers(*loggingConfig)
 	}
-	if *showLog {
-		writer := loggo.NewSimpleWriter(os.Stderr, &loggo.DefaultFormatter{})
-		_, err := loggo.ReplaceDefaultWriter(writer)
+	if !*showLog {
+		_, _, err := loggo.RemoveWriter("default")
 		if err != nil {
 			return err
 		}
@@ -76,6 +77,7 @@ func load() error {
 		return err
 	}
 	for _, tip := range tips {
+		// TODO(jay) need to process bundles as well (there is a card)
 		if !strings.HasSuffix(tip.UniqueName, "/trunk") {
 			continue
 		}
@@ -86,17 +88,17 @@ func load() error {
 			continue
 		}
 		if tip.Revision == "" {
-			logger.Tracef("skipping %v no revision", tip.UniqueName)
+			logger.Errorf("skipping branch %v with no revisions", tip.UniqueName)
 			continue
 		} else {
-			logger.Tracef("found %v with revision %v", tip.UniqueName, tip.Revision)
+			logger.Debugf("found %v with revision %v", tip.UniqueName, tip.Revision)
 		}
 		URLs := []*charm.URL{charmURL}
 		schema, name := charmURL.Schema, charmURL.Name
-		addPromulgatedCharmURLs(tip.OfficialSeries, schema, name, URLs)
+		URLs = addPromulgatedCharmURLs(tip.OfficialSeries, schema, name, URLs)
 		err = publishBazaarBranch(*storeURL, *storeUser, URLs, branchURL, tip.Revision)
 		if err != nil {
-			logger.Errorf("publishing branch %v to charmstore: %v", branchURL, err)
+			logger.Errorf("cannot publish branch %v to charm store: %v", branchURL, err)
 		}
 		if _, ok := err.(*UnauthorizedError); ok {
 			return err
@@ -107,9 +109,9 @@ func load() error {
 }
 
 // addPromulgatedCharmURLs adds urls from officialSeries to
-// the URLs slice for the given schema, name.
+// the URLs slice for the given schema and name.
 // Promulgated charms have OfficialSeries in launchpad.
-func addPromulgatedCharmURLs(officialSeries []string, schema, name string, URLs []*charm.URL) {
+func addPromulgatedCharmURLs(officialSeries []string, schema, name string, URLs []*charm.URL) []*charm.URL {
 	for _, series := range officialSeries {
 		nextCharmURL := &charm.URL{
 			Schema:   schema,
@@ -120,6 +122,7 @@ func addPromulgatedCharmURLs(officialSeries []string, schema, name string, URLs 
 		URLs = append(URLs, nextCharmURL)
 		logger.Debugf("added URL %v to URLs list for %v", nextCharmURL, URLs[0])
 	}
+	return URLs
 }
 
 // uniqueNameURLs returns the branch URL and the charm URL for the
@@ -185,6 +188,7 @@ func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, b
 	mwriter := io.MultiWriter(hash1, &counter)
 	thisCharm.ArchiveTo(mwriter)
 	hash1str := fmt.Sprintf("%x", hash1.Sum(nil))
+	// Asyncronously write the archive while the http requestreads from the Pipe.
 	go func() {
 		thisCharm.ArchiveTo(writer)
 		writer.Close()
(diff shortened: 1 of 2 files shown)


jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/203/

urosj:
Please add launchpad.net/lpad to dependencies.tsv

cmd/charmload/main.go
```diff
@@ -0,0 +1,256 @@
...
+		os.Exit(1)
+	}
+}
+
+func load() error {
+	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	staging := flags.Bool("staging", false, "use the launchpad staging server")
+	storeurl := flags.String("storeurl", "http://localhost:8080/v4/", "the url of the charmstore")
+	loggingConfig := flags.String("logging-config", "", "specify log levels for modules")
+	showlog := flags.Bool("show-log", false, "if set, write log messages to stderr")
```
bac: I find the inconsistent use of camelCase confusing. Is there a pattern I'm not seeing?

cmd/charmload/main.go
```diff
@@ -0,0 +1,256 @@
...
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+}
+
+func load() error {
+	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	staging := flags.Bool("staging", false, "use the launchpad staging server")
+	storeurl := flags.String("storeurl", "http://localhost:8080/v4/", "the url of the charmstore")
+	loggingConfig := flags.String("logging-config", "", "specify log levels for modules")
```
bac: In the usage for logging-config please provide an example or the possible config values. Users should not have to read juju/loggo code to know how to properly write this string.  The problem is made worse since logo.ConfigureLoggers silently rejects improperly formatted/non-sense strings.

Having the user provide the config string is very flexible but demanding.  What about having --trace, --debug, etc., that applies to the 'charmload' module?

Uros indicates adding specific switches has been discussed for a later task.

cmd/charmload/main.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return err
+	}
+	tips, err := charmsDistro.BranchTips(time.Time{})
+	if err != nil {
+		return err
+	}
+	for _, tip := range tips {
+		if !strings.HasSuffix(tip.UniqueName, "/trunk") {
+			continue
```
bac, line 82: Due to this check, only charms are ingested.  Bundles look the same on Launchpad but have a branch named "bundle" instead of "trunk".

Is processing bundles outside the scope of this branch?

cmd/charmload/main.go
```diff
@@ -0,0 +1,256 @@
...
+func uniqueNameURLs(name string) (burl string, charmurl *charm.URL, err error) {
+	u := strings.Split(name, "/")
+	if len(u) > 5 {
+		u = u[len(u)-5:]
+		burl = name
+	} else {
+		burl = "lp:" + name
+	}
+	if len(u) < 5 || u[1] != "charms" || u[4] != "trunk" || len(u[0]) == 0 || u[0][0] != '~' {
+		return "", nil, fmt.Errorf("unwanted branch name: %s", name)
```
bac: Perhaps s/unwanted/unexpected/  in the error message?

cmd/charmload/main.go
```diff
@@ -0,0 +1,256 @@
...
+		// newer revision and published that first, and the digest
+		// parameter provided is in fact an old version that would
+		// overwrite the new version.
+		tipDigest, err := bzrRevisionId(branchDir)
+		if err != nil {
+			return err
+		}
+		if tipDigest != digest {
+			digest = tipDigest
+			goto NewTip
```
bac: How about using a while loop instead of this goto?

bac:
Hi Jay,

Thanks for getting the charmload process going.  

I tried running charmload but consistently got failures from the POST to localhost.

What are your plans for testing this branch?

Finally, please include basic QA instructions for your branch as it a) informs your reviewers on the basics of running the code, especially if there are dependent processes and b) makes for a nice historical record, especially if we ever need to reconstruct QA for an SRU, etc.

cmd/charmload/main.go
jrwren: http://golang.org/doc/effective_go.html#mixed-caps
"the convention in Go is to use MixedCaps or mixedCaps rather than underscores to write multiword names."

Now I see just how inconsistent I was in doing this. I will fix.

jrwren:
@urosj Done - added lpad to dependencies.tsv

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/204/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/205/

bac:
Hi Jay, 

Following your new QA instructions I am able to get the ingestion to work.  After watching a charm get ingested I can visit

http://localhost:8080/v4/~a.rosales/trusty/sugarcrm/meta/charm-metadata

and get results.  Hurrah.

However after ingesting 26 charms I begin seeing failures like:

2014-09-10 08:07:54 TRACE charmload main.go:78 getting uniqueNameURLs for ~zulcss/charms/oneiric/nagios/trunk
2014-09-10 08:07:54 DEBUG charmload main.go:154 running bzr checkout ... lp:~zulcss/charms/oneiric/nagios/trunk
2014-09-10 08:07:54 ERROR charmload main.go:102 publishing branch lp:~zulcss/charms/oneiric/nagios/trunk to charmstore: fork/exec /usr/bin/bzr: cannot allocate memory

This is running on trusty in vmware.  Let's see if that happens on metal.

cmd/charmload/main.go
jrwren: I expect any name, since a branch may be any name. Maybe unsupported?

cmd/charmload/main.go
jrwren, line 82: I did not think it was in scope. It is an important feature. I'll make a card.

cmd/charmload/main.go
jrwren: done

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/213/

cmd/charmload/main.go
jrwren: It shouldn't even be there. I copied code from _old/branch.go and deleted things which did not make sense, turning the goto into a no-op. In fact, after making sure that the code is not necessary, I'm not convinced it ever prevented the race condition which it intended to prevent.

Removing.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/214/

bac:
Jay,

I ran the test with go 1.3.1 overnight. It did better but after ingesting 374 charms it started to fail as before:

2014-09-10 16:54:45 DEBUG charmload main.go:154 running bzr checkout ... lp:~zulcss/charms/oneiric/nagios/trunk
2014-09-10 16:54:45 ERROR charmload main.go:102 publishing branch lp:~zulcss/charms/oneiric/nagios/trunk to charmstore: fork/exec /usr/bin/bzr: cannot allocate memory

This is on a trusty VM with 3G of memory and very little else running (term, emacs, chromium).

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/218/

jrwren:
Brad,

https://github.com/juju/charm/pull/45 landed so I updated dependencies.tsv. Please try again make sure to run make deps before running charm loader.

urosj:
@jrwren I've done QA today and it worked great. 936 charms ingested into CS. Memory usage was stable, CPU usage low. Awesome work!

cmd/charmload/main.go
urosj, line 82: It is out of scope indeed. Changes in bundle yaml file need to be resolved first.

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+	"strings"
+	"time"
+
+	"github.com/juju/loggo"
+	"gopkg.in/juju/charm.v3"
+
+	"launchpad.net/lpad"
+)
+
+var logger = loggo.GetLogger("charmload")
```
urosj: What about if we say charmload_v4?

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+		loggo.ConfigureLoggers(*loggingConfig)
+	}
+	if *showLog {
+		writer := loggo.NewSimpleWriter(os.Stderr, &loggo.DefaultFormatter{})
+		_, err := loggo.ReplaceDefaultWriter(writer)
+		if err != nil {
+			return err
+		}
+	}
+	oauth := &lpad.OAuth{Anonymous: true, Consumer: "juju"}
```
urosj, line 65: I would add these details to the documentation of the function.

cmd/charmload/main.go
urosj, line 82: Add a //Todo(jay) need to process bundles as well
PS: the card is already created

urosj:
:+1: Awesome work, @jrwren

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	charmURL, err = charm.ParseURL(fmt.Sprintf("cs:%s/%s/%s", u[0], u[2], u[3]))
+	if err != nil {
+		return "", nil, err
+	}
+	return branchURL, charmURL, nil
+}
+
+func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, branchURL string, digest string) error {
+
```
frankban: I guess this new line can be removed.

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+			logger.Tracef("skipping %v no revision", tip.UniqueName)
+			continue
+		} else {
+			logger.Tracef("found %v with revision %v", tip.UniqueName, tip.Revision)
+		}
+		URLs := []*charm.URL{charmURL}
+		schema, name := charmURL.Schema, charmURL.Name
+		for _, series := range tip.OfficialSeries {
+			nextCharmURL := &charm.URL{
+				Schema:   schema,
```
frankban: I am sure I am missing something here, but why we don't copy the user for each series?

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+func main() {
+	err := load()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+}
+
+func load() error {
+	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
```
frankban, line 42: // Define and parse the command line arguments?

cmd/charmload/main.go
frankban, line 65: // Log in into Launchpad in order to retrieve the charm branches?

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+// URL, and the unique name is returned unchanged as the branch URL.
+func uniqueNameURLs(name string) (branchURL string, charmURL *charm.URL, err error) {
+	u := strings.Split(name, "/")
+	if len(u) > 5 {
+		u = u[len(u)-5:]
+		branchURL = name
+	} else {
+		branchURL = "lp:" + name
+	}
+	if len(u) < 5 || u[1] != "charms" || u[4] != "trunk" || len(u[0]) == 0 || u[0][0] != '~' {
```
frankban, line 145: Just for my own curiosity: the u[1] != "charms" || u[4] != "trunk" checks should never fail right? And that's because previously we explicitly looped over the charms distro, and excluded non-trunk tip names.

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		if _, ok := err.(*UnauthorizedError); ok {
+			return err
+		}
+
+	}
+	return nil
+}
+
+// uniqueNameURLs returns the branch URL and the charm URL for the
```
frankban, line 128: It would be really nice to have unit tests for this function, and the same for publishBazaarBranch and bzrRevisionId.
This is not a blocker, but I'd be less nervous about future changes to the command if the logic here were more exercized.

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+	tipDigest, err := bzrRevisionId(branchDir)
+	if err != nil {
+		return err
+	}
+	if tipDigest != digest {
+		digest = tipDigest
+		logger.Warningf("tipDigest %v != digest %v", digest, tipDigest)
+	}
+
+	thischarm, err := charm.ReadCharmDir(branchDir)
```
frankban: ch, or thisCharm?

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/222/

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+	logger.Tracef("read CharmDir from branchDir %v", thischarm, branchDir)
```
frankban: Do we need two placeholders for the text here?
Perhaps move this line above the ch definition and avoid using thischarm before checking the error?

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+	if err == nil {
```
frankban: if err != nil {
    return err
}

and then we can dedent the block below?

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	return err
+}
+
+// bzrRevisionId returns the Bazaar revision id for the branch in branchDir.
+func bzrRevisionId(branchDir string) (string, error) {
+	cmd := exec.Command("bzr", "revision-info")
+	cmd.Dir = branchDir
+	stderr := &bytes.Buffer{}
```
frankban, line 282: var stderr bytes.Buffer
cmd.Stderr = &stderr
?

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+	thischarm, err := charm.ReadCharmDir(branchDir)
+	logger.Tracef("read CharmDir from branchDir %v", thischarm, branchDir)
+	if err == nil {
+		reader, writer := io.Pipe()
+		hash1 := sha512.New384()
+		var counter Counter
+		mwriter := io.MultiWriter(hash1, &counter)
+		thischarm.ArchiveTo(mwriter)
+		hash1str := fmt.Sprintf("%x", hash1.Sum(nil))
+		go func() {
```
frankban: I think I understood why we need this goroutine here (together with the io.Pipe above), but making it explicit with a comment would really help future readers.

cmd/charmload/main.go
```diff
@@ -0,0 +1,250 @@
...
+			thischarm.ArchiveTo(writer)
+			writer.Close()
+		}()
+		id := URLs[0]
+		URL := storeURL + id.Path() + "/archive?hash=" + hash1str
+		logger.Infof("posting to %v", URL)
+		request, err := http.NewRequest("POST", URL, reader)
```
frankban: Don't we need to check the possible error here? If not, why (comment)?

frankban:
Thanks for this branch Jay, :+1: with some comments.

cmd/charmload/main.go
jrwren, line 145: Yes. This function was copied from _old/lpad.go and unchanged.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/223/

cmd/charmload/main.go
jrwren, line 128: I agree. If anyone has ideas on how to do that without manually mocking both lpad and a bit of http. Maybe refactoring to isolate those and test what is left?

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
jrwren, line 65: Done.

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
```diff
@@ -0,0 +1,260 @@
...
+			continue
+		}
+		logger.Tracef("getting uniqueNameURLs for %v", tip.UniqueName)
+		branchURL, charmURL, err := uniqueNameURLs(tip.UniqueName)
+		if err != nil {
+			logger.Infof("could not get uniqueNameURLs for %v: %v", tip.UniqueName, err)
+			continue
+		}
+		if tip.Revision == "" {
+			logger.Tracef("skipping %v no revision", tip.UniqueName)
```
rogpeppe: skipping branch %v with no revisions
?

Also, the original code logged this as an error - perhaps it might be worth doing that here.

cmd/charmload/main.go
jrwren: Good point. I extracted this function and documented it. OfficialSeries means the charm is promulgated.

cmd/charmload/main.go
```diff
@@ -0,0 +1,260 @@
...
+			continue
+		} else {
+			logger.Tracef("found %v with revision %v", tip.UniqueName, tip.Revision)
```
rogpeppe: s/Tracef/Debugf/ ?

cmd/charmload/main.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+		URLs := []*charm.URL{charmURL}
+		schema, name := charmURL.Schema, charmURL.Name
+		addPromulgatedCharmURLs(tip.OfficialSeries, schema, name, URLs)
+		err = publishBazaarBranch(*storeURL, *storeUser, URLs, branchURL, tip.Revision)
+		if err != nil {
+			logger.Errorf("publishing branch %v to charmstore: %v", branchURL, err)
```
rogpeppe: s/publishing/cannot publish/
s/charmstore/charm store/
?

cmd/charmload/main.go
frankban, line 128: +1
Also, uniqueNameURLs does not seem to require mocking.

cmd/charmload/main.go
```diff
@@ -0,0 +1,260 @@
...
+		if _, ok := err.(*UnauthorizedError); ok {
+			return err
+		}
+
+	}
+	return nil
+}
+
+// addPromulgatedCharmURLs adds urls from officialSeries to
+// the URLs slice for the given schema, name.
```
rogpeppe: s/,/ and/
?

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
```diff
@@ -0,0 +1,260 @@
...
+// Promulgated charms have OfficialSeries in launchpad.
+func addPromulgatedCharmURLs(officialSeries []string, schema, name string, URLs []*charm.URL) {
```
rogpeppe: I don't see how this can work - it appends to URLs but doesn't return the resulting slice.

cmd/charmload/main.go
jrwren, line 282: Untouched function from _old/branch.go. Fixing regardless. 
Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/224/

cmd/charmload/main.go
jrwren: To much time with C#, I take these things for granted. Can you suggest how to comment it, because I do not know what I would write.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/225/

cmd/charmload/main.go
```diff
@@ -0,0 +1,260 @@
...
+	if tipDigest != digest {
+		digest = tipDigest
+		logger.Warningf("tipDigest %v != digest %v", digest, tipDigest)
+	}
+
+	thischarm, err := charm.ReadCharmDir(branchDir)
+	logger.Tracef("read CharmDir from branchDir %v", thischarm, branchDir)
+	if err == nil {
+		reader, writer := io.Pipe()
+		hash1 := sha512.New384()
```
rogpeppe: s/sha512.New384/blobstore.NewHash/

cmd/charmload/main.go
jrwren: It will only fail if the url isn't able to be parsed. That url could be specified with cmd line option and could be wrong. Thanks, good catch.

Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/226/

cmd/charmload/main.go
```diff
@@ -0,0 +1,260 @@
...
+		var counter Counter
+		mwriter := io.MultiWriter(hash1, &counter)
+		thischarm.ArchiveTo(mwriter)
+		hash1str := fmt.Sprintf("%x", hash1.Sum(nil))
```
rogpeppe: This doesn't seem right to me - we need to calculate the hash _before_ uploading the charm - we're getting an arbitrary  hash. I suggest that we archive to a temporary file before uploading it. I think it might be worth factoring all that out into a separate function.

jrwren:
Moving QA instructions out of PR description
UPDATE

QA instructions:
1. Make sure your dependencies are up to date, run make reps
2. Run mongodb. I use system mongodb
3. Run charmd using go run: go run cmd/charmd/main.go cmd/charmd/config.yaml &
4. Run charmload using go run: go run cmd/charmload/main.go --staging=true --show-log --logging-config='&lt;root&gt;=TRACE'
5. Optionally omit the --staging=true option to use production launchpad.

And, when I follow these instructions on a clean utopic system, I get same errors mentioned by @bac and @urosj so I'm keeping this blocked until I resolve them.  No idea, yet, why it WFM on a different system.

cmd/charmload/main.go
frankban: The code actually calculates the hash before uploading the zip. However, I am still +1 on your suggestion, and with the refactoring you propose we would not require the goroutine and/or the counter type.

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+		return err
+	}
+	reader, writer := io.Pipe()
+	hash1 := sha512.New384()
+	var counter Counter
+	mwriter := io.MultiWriter(hash1, &counter)
+	thisCharm.ArchiveTo(mwriter)
+	hash1str := fmt.Sprintf("%x", hash1.Sum(nil))
+	go func() {
+		thisCharm.ArchiveTo(writer)
```
rogpeppe: There's no particular guarantee that ArchiveTo is deterministic (some other process might be traversing the fs and reading files, changing the file access times, for example). I still think it's probably better to archive to a temporary file (calculating the hash at the same time), then read from that. Then no need for a pipe (or Counter), although it will use more disk space.

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+		writer.Close()
+	}()
+	id := URLs[0]
+	URL := storeURL + id.Path() + "/archive?hash=" + hash1str
+	logger.Infof("posting to %v", URL)
+	request, err := http.NewRequest("POST", URL, reader)
+	if err != nil {
+		return err
```
rogpeppe, line 189: errgo.Notef(err, "cannot make new request")

?

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+	authhash := base64.StdEncoding.EncodeToString([]byte(storeUser))
+	logger.Tracef("encoded Authorization %v", authhash)
+	request.Header["Authorization"] = []string{"Basic " + authhash}
+	// go1.2.1 has a bug requiring Content-Type to be sent
+	// since we are posting to a go server which may be running on
+	// 1.2.1, we should send this header
+	// https://code.google.com/p/go/source/detail?r=a768c0592b88
+	request.Header["Content-Type"] = []string{"application/octet-stream"}
+	request.ContentLength = int64(counter)
+	resp, err := http.DefaultClient.Do(request)
```
rogpeppe, line 252: You should close resp.Body.

Perhaps create a separate function to do a charm store request.
Something like this, perhaps?

```
func CharmStoreDo(req *http.Request, result interface{}) error {
    resp, err := http.DefaultClient.Do(req)
    if err != nil {
          return errgo.Mask(err)
    }
    defer resp.Body.Close()
    dec := json.NewDecoder(resp.Body)
    if resp.StatusCode != http.StatusOk {
         var perr params.Error
         if err := dec.Decode(&perr); err != nil {
              return errgo.Notef(err, "cannot unmarshal error response")
         }
         if err.Message == "" {
               return errgo.New("error response with empty message")
         }
         return &perr
    }
    if err := dec.Decode(result); err != nil {
         return errgo.Notef(err, "cannot unmarshal response")
    }
    return nil
}
```

Then you can do:

```
  var resp params.ArchivePostResponse
  if err := CharmStoreDo(req, &resp); err != nil {
       logger.Errorf("error uploading %s: %v", id, err)
  } else {
       logger.Debugf("uploaded %s", resp.Id)
  }
```

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+	if resp.StatusCode == http.StatusUnauthorized {
+		logger.Errorf("invalid charmstore credentials")
```
rogpeppe: We're returning the error - it will be printed by our caller, so no need to log it here.

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+		return &UnauthorizedError{}
+	}
+	if err != nil || resp.StatusCode != http.StatusOK {
+		logger.Warningf("error posting:", err, resp.Header)
```
rogpeppe: This is an invalid fmt format. (it's missing the %s)

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+	"strings"
+	"time"
+
+	"github.com/juju/loggo"
+	"gopkg.in/juju/charm.v3"
+
+	"launchpad.net/lpad"
+)
+
+var logger = loggo.GetLogger("charmload_v4")
```
rogpeppe, line 27: I think this could probably be just "charmload".
We might consider renaming the command actually, to be something that
indicates it's lp-specific. (lpcharmload ?). That can be left for another PR though.

cmd/charmload/main.go
jrwren: You are right, I was returning it first, then my go-lang newbieness told me its mutable and I changed it. Fixing it.

cmd/charmload/main.go
jrwren: I have intentionally avoided importing blobstore package to utilize only the public http interface.

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+		writer.Close()
+	}()
+	id := URLs[0]
+	URL := storeURL + id.Path() + "/archive?hash=" + hash1str
+	logger.Infof("posting to %v", URL)
+	request, err := http.NewRequest("POST", URL, reader)
+	if err != nil {
+		return err
+	}
+	authhash := base64.StdEncoding.EncodeToString([]byte(storeUser))
```
rogpeppe: How is this a hash?

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+// and get all the Branch Tips in the charms Distro.
+// For each Branch Tip with name ending in /trunk, publish in
+// charmstore
+func load() error {
+	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	staging := flags.Bool("staging", false, "use the launchpad staging server")
+	storeURL := flags.String("storeurl", "http://localhost:8080/v4/", "the URL of the charmstore")
+	loggingConfig := flags.String("logging-config", "", "specify log levels for modules e.g. <root>=TRACE")
+	showLog := flags.Bool("show-log", false, "if set, write log messages to stderr")
+	storeUser := flags.String("user", "admin:example-passwd", "the colon separated user:password for charmstore")
```
rogpeppe: Perhaps go for the same approach here as in the charm config - have both -user and -password options, and default the user value to "admin" ? Another possibility would be to default to reading the same config file that the charm store itself uses, making it trivial to run charmload in the same charm as the charm store.

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	server := lpad.Production
+	if *staging {
+		server = lpad.Staging
+	}
+	if *loggingConfig != "" {
+		loggo.ConfigureLoggers(*loggingConfig)
+	}
+	if *showLog {
+		writer := loggo.NewSimpleWriter(os.Stderr, &loggo.DefaultFormatter{})
```
rogpeppe: Isn't this already the default log writer?

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+func (c *Counter) Write(p []byte) (n int, err error) {
+	size := len(p)
+	*c += Counter(size)
+	return size, nil
+}
+
+type UnauthorizedError struct{}
+
+func (_ *UnauthorizedError) Error() string {
+	return "UnauthorizedError"
```
rogpeppe:
```
return "invalid charmstore credentials"
```

?

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (c *Counter) Write(p []byte) (n int, err error) {
+	size := len(p)
+	*c += Counter(size)
+	return size, nil
+}
+
+type UnauthorizedError struct{}
+
+func (_ *UnauthorizedError) Error() string {
```
rogpeppe: you can leave out the _ here.

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+
+type Counter int
+
+func (c *Counter) Write(p []byte) (n int, err error) {
+	size := len(p)
+	*c += Counter(size)
+	return size, nil
+}
+
+type UnauthorizedError struct{}
```
rogpeppe, line 308: For a single value error, there's no particular need to make a type - just use
a value.

e.g.

```
var errUnauthorized = errgo.New("invalid charm store credentials")
```

Then you can test the error by just checking equality (using errgo.Cause)

cmd/charmload/main.go
jrwren, line 27: I had added _v4 by @urosj request.

cmd/charmload/main.go
```diff
@@ -0,0 +1,263 @@
...
+			logger.Tracef("found %v with revision %v", tip.UniqueName, tip.Revision)
+		}
+		URLs := []*charm.URL{charmURL}
+		schema, name := charmURL.Schema, charmURL.Name
+		addPromulgatedCharmURLs(tip.OfficialSeries, schema, name, URLs)
+		err = publishBazaarBranch(*storeURL, *storeUser, URLs, branchURL, tip.Revision)
+		if err != nil {
+			logger.Errorf("publishing branch %v to charmstore: %v", branchURL, err)
+		}
+		if _, ok := err.(*UnauthorizedError); ok {
```
rogpeppe, line 103: (see later comment)

```
if errgo.Cause(err) == errUnauthorized {
    return errgo.Notef(err, "fatal error uploading charms")
}
```

?

cmd/charmload/main.go
jrwren: I did not feel it important and this takes username password the same as curl -u, which I had been using too. I don't feel strongly about it, I'd like input from someone else.

cmd/charmload/main.go
jrwren: Indeed. I did not realize this.

rogpeppe:
Lots of comments, but mostly minor, and it looks nice in general, thanks. I suggest that the publish code moves to a separate package, with some tests (the old code had some tests, so it seems like we could have some here too).

cmd/charmload/main.go
rogpeppe: Yeah, curl -u is reasonable precedent. How about making this -u too, by analogy?
I still think that taking input from the config file is a nice default, otherwise we'll need to write external code to do that.

cmd/charmload/main.go
jrwren: 50% more IO. Read, Write, Read instead of just Read twice. I had not considered something traversing the fs and changing files. Given that this is branched to a temp director created specifically to do these operations, I do not share them as a concern. If something is tromping all over the filesystem, it could just as easily delete or change the temporary charm file before it gets read in again.

cmd/charmload/main.go
jrwren, line 189: Done.

cmd/charmload/main.go
jrwren: :) its not, at all. Fixing.

cmd/charmload/main.go
jrwren, line 252: IMO, its overkill for an error that I've only observed when using incorrect credentials and I'd have to do more to bubble up the fail fast UnauthorizedError.

cmd/charmload/main.go
jrwren: Removed.

cmd/charmload/main.go
jrwren: Good catch. Done.

cmd/charmload/main.go
jrwren, line 308: Good to know. Is that an errgo feature?

By checking string equality, I feel like I've resorted to magic strings where use of a type checks that I don't make a magic string typo.

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
jrwren: Done.

cmd/charmload/main.go
jrwren: Done.

I agree that the functionality would be nice. I feel that it is easier than it sounds. Which config file? $GOPATH/src/github.com/juju/charmstore/cmd/charmd/config.yaml or /etc/charmd.yaml (as CI runs using that path) Should we take config file path as cmd arg?

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/227/

cmd/charmload/main.go
urosj, line 27: @jrwren sorry about that, I was thinking about different logs for different scenarios, like charmload-errors, or charmload-failed, etc. and somehow made this comment about adding _v4 as postfix. 
@rogpeppe lpcharmload ... it will load bundles as well. lploader or ingestLP maybe?

cmd/charmload/main.go
rogpeppe, line 252: You would not have to do anything more to bubble up the UnauthorizedError - you won't need the http status because the resulting JSON error message will have params.ErrUnauthorized as the error code.

I also think that it will be useful to have this function for the future, as it encapsulates the correct logic for making an arbitrary request to the charm store server.

addPromulgatedCharmURLs could do with some simplification, and this suggestion is trying to help that.

cmd/charmload/main.go
rogpeppe: I would prefer it if there was no chance at all of this going wrong, please. Modifying the file system is one thing, but this code is not obviously correct - it relies on the implementation of ArchiveTo. For example, it is currently the case that the access time of the files is not archived, but it could be, and then this code would probably fail, because calling ArchiveTo would itself modify the access times of the files. Also, yes it will use more IO, but compressing can also be expensive too and temporary files are often in fast storage.

If we archive to a temporary file, we know that it's going to be OK, and I think that assurance is worth something. The code will also be simpler that way.

One other thing: please check the error returned by ArchiveTo.

cmd/charmload/main.go
rogpeppe: Agreed, it could use some discussion. Let's leave that functionality for a subsequent PR.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/232/

cmd/charmload/main.go
```diff
@@ -0,0 +1,312 @@
...
+// and computes the hash needed for posting calls stat on the file to get its
+// size.
+func getPostDataFromCharmDir(charmDir *charm.CharmDir) (tempFile *os.File, hash string, size int64, err error) {
+	sha384 := sha512.New384()
+	tempFile, err = ioutil.TempFile("", "publish-charm")
+	logger.Debugf("writing charm to temporary file %s", tempFile.Name())
+	if err != nil {
+		return tempFile, "", 0, err
+	}
+	err = charmDir.ArchiveTo(tempFile)
```
rogpeppe, line 224: s/tempFile/io.MultiWriter(tempFile, sha384)/

?

(no need to read the file twice)

cmd/charmload/main.go
```diff
@@ -0,0 +1,312 @@
...
+
+// getPostDataFromCharmDir archives the charmDir to a temporary file
+// and computes the hash needed for posting calls stat on the file to get its
+// size.
+func getPostDataFromCharmDir(charmDir *charm.CharmDir) (tempFile *os.File, hash string, size int64, err error) {
+	sha384 := sha512.New384()
+	tempFile, err = ioutil.TempFile("", "publish-charm")
+	logger.Debugf("writing charm to temporary file %s", tempFile.Name())
+	if err != nil {
+		return tempFile, "", 0, err
```
rogpeppe, line 222: s/tempFile/nil/

(not that it makes any difference in practice - it'll be nil anyway in this case)

cmd/charmload/main.go
```diff
@@ -0,0 +1,312 @@
...
+		if err != nil {
+			errgo.Notef(err, "cannot make new request")
+			return err
+		}
+		request.Header["Authorization"] = []string{"Basic " + authBasic}
+		// go1.2.1 has a bug requiring Content-Type to be sent
+		// since we are posting to a go server which may be running on
+		// 1.2.1, we should send this header
+		// https://code.google.com/p/go/source/detail?r=a768c0592b88
+		request.Header["Content-Type"] = []string{"application/octet-stream"}
```
rogpeppe, line 205: I'd lose the comment about the Go bug and set the content type to application/zip,
which is what it is.

also, it can be a little simpler:

```
request.Header.Set("Content-Type", "application/zip")
```

cmd/charmload/main.go
```diff
@@ -0,0 +1,312 @@
...
+	authBasic := base64.StdEncoding.EncodeToString([]byte(storeUser))
+	for _, id := range URLs {
+		URL := storeURL + id.Path() + "/archive?hash=" + hash
+		logger.Infof("posting to %v", URL)
+		request, err := http.NewRequest("POST", URL, tempFile)
+		if err != nil {
+			errgo.Notef(err, "cannot make new request")
+			return err
+		}
+		request.Header["Authorization"] = []string{"Basic " + authBasic}
```
rogpeppe, line 200:
```
request.Header.Set("Authorization", "Basic" + authBasic)
```

but perhaps move this into the doCharmStorePost method?

cmd/charmload/main.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		return tempFile, "", 0, err
+	}
+	size = fileInfo.Size()
+	return tempFile, hash, size, err
+}
+
+// doCharmStorePost performs the request and returns an error or nil
+// if the post request succeeded.
+func doCharmStorePost(request *http.Request) error {
```
rogpeppe, line 251: Perhaps doCharmStoreRequest ? (there's nothing here that makes it necessarily a POST request - it could work for any request)

urosj:
:+1: Agreed to be merged and then completed with a new PR. Thanks all.

jrwren:
:shipit:

jujugui:

PR #100, opened 2014-09-12
internal/legacy: implement charm-info endpoint

rogpeppe:
This requires the SHA256 hash of the content. Rather than clutter the other code,
we calculate this lazily on demand.

### DIFF
--- a/internal/charmstore/store_test.go
+++ b/internal/charmstore/store_test.go
@@ -5,6 +5,7 @@ package charmstore
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -675,6 +676,61 @@ func (s *StoreSuite) TestAddBundleArchive(c *gc.C) {
 	)
 	c.Assert(err, gc.IsNil)
 	s.checkAddBundle(c, bundleArchive)
+}
+
+func (s *StoreSuite) TestOpenBlob(c *gc.C) {
+	charmArchive := testing.Charms.CharmArchive(c.MkDir(), "wordpress")
+
+	store, err := NewStore(s.Session.DB("foo"))
+	c.Assert(err, gc.IsNil)
+	url := mustParseReference("cs:precise/wordpress-23")
+
+	err = store.AddCharmWithArchive(url, charmArchive)
+	c.Assert(err, gc.IsNil)
+
+	r, size, err := store.OpenBlob(url)
+	c.Assert(err, gc.IsNil)
+	defer r.Close()
+
+	f, err := os.Open(charmArchive.Path)
+	c.Assert(err, gc.IsNil)
+	defer f.Close()
+	c.Assert(hashOfReader(c, r), gc.Equals, hashOfReader(c, f))
+
+	info, err := f.Stat()
+	c.Assert(err, gc.IsNil)
+	c.Assert(size, gc.Equals, info.Size())
+}
+
+func (s *StoreSuite) TestBlobName(c *gc.C) {
+	charmArchive := testing.Charms.CharmArchive(c.MkDir(), "wordpress")
+
+	store, err := NewStore(s.Session.DB("foo"))
+	c.Assert(err, gc.IsNil)
+	url := mustParseReference("cs:precise/wordpress-23")
+
+	err = store.AddCharmWithArchive(url, charmArchive)
+	c.Assert(err, gc.IsNil)
+
+	name, err := store.BlobName(url)
+	c.Assert(err, gc.IsNil)
+
+	r, _, err := store.BlobStore.Open(name)
+	c.Assert(err, gc.IsNil)
+	defer r.Close()
+
+	f, err := os.Open(charmArchive.Path)
+	c.Assert(err, gc.IsNil)
+	defer f.Close()
+	c.Assert(hashOfReader(c, r), gc.Equals, hashOfReader(c, f))
+
+}
+
+func hashOfReader(c *gc.C, r io.Reader) string {
+	hash := sha256.New()
+	_, err := io.Copy(hash, r)
+	c.Assert(err, gc.IsNil)
+	return fmt.Sprintf("%x", hash.Sum(nil))
 }
 
 func mustGetSizeAndHash(c interface{}) (int64, string) {
--- a/internal/legacy/api.go
+++ b/internal/legacy/api.go
@@ -4,14 +4,20 @@
 package legacy
 
 import (
+	"crypto/sha256"
 	"fmt"
+	"io"
+	"log"
 	"net/http"
 	"strings"
 
 	"github.com/juju/errgo"
 	"gopkg.in/juju/charm.v3"
+	"gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
 
 	"github.com/juju/charmstore/internal/charmstore"
+	"github.com/juju/charmstore/internal/mongodoc"
 	"github.com/juju/charmstore/internal/router"
 	"github.com/juju/charmstore/internal/v4"
 	"github.com/juju/charmstore/params"
@@ -29,7 +35,7 @@ func NewAPIHandler(store *charmstore.Store, config charmstore.ServerParams) http
 		store: store,
 		mux:   http.NewServeMux(),
 	}
-	h.handle("/charm-info", http.HandlerFunc(h.serveCharmInfo))
+	h.handle("/charm-info", router.HandleJSON(h.serveCharmInfo))
 	h.handle("/charm/", router.HandleErrors(h.serveCharm))
 	return h
 }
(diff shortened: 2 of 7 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/219/

internal/legacy/api.go
```diff
@@ -40,24 +46,98 @@ func (h *Handler) handle(path string, handler http.Handler) {
...
+			// Lazily calulate SHA256 so that we don't burden
+			// non-legacy code with that task.
+			entity.BlobHash256, err = h.updateEntitySHA256(curl)
+		}
+		if err == nil {
+			// TODO(rog) increment charm-info stats counter?
+			c.CanonicalURL = curl.String()
+			c.Sha256 = entity.BlobHash256
+			c.Revision = curl.Revision
+			// TODO(rog) include Digest if it exists in extra-info ?
```
frankban, line 108: If not present already, could you please add a card for this TODO?

internal/charmstore/store_test.go
```diff
@@ -677,6 +678,61 @@ func (s *StoreSuite) TestAddBundleArchive(c *gc.C) {
 	s.checkAddBundle(c, bundleArchive)
 }
 
+func (s *StoreSuite) TestOpenBlob(c *gc.C) {
```
frankban, line 681: Nice tests.

internal/legacy/api.go
```diff
@@ -40,24 +46,98 @@ func (h *Handler) handle(path string, handler http.Handler) {
...
+		curl, err := h.resolveURLStr(url)
+		var entity mongodoc.Entity
+		if err != nil {
+			if errgo.Cause(err) == params.ErrNotFound {
+				err = errNotFound
+			}
+		} else {
+			err = h.store.DB.Entities().FindId(curl).One(&entity)
+			if err == mgo.ErrNotFound {
+				// The old API actually returned "entry not found"
```
frankban, line 92: This comment is right, but looks a bit weird given than right below we return an "entry not found"...

internal/legacy/api.go
```diff
@@ -40,24 +46,98 @@ func (h *Handler) handle(path string, handler http.Handler) {
...
+				// on *any* error, but it seems reasonable to be
+				// a little more descriptive for other errors.
+				err = errNotFound
+			}
+		}
```
frankban, line 97: Perhaps we could append the error and continue in case of errors here, so that below we can avoid checking err twice.

internal/legacy/api.go
```diff
@@ -40,24 +46,98 @@ func (h *Handler) handle(path string, handler http.Handler) {
...
+		if err == nil && entity.BlobHash256 == "" {
+			// Lazily calulate SHA256 so that we don't burden
+			// non-legacy code with that task.
+			entity.BlobHash256, err = h.updateEntitySHA256(curl)
```
frankban, line 101: Oh I see we also have an err here.

internal/legacy/api.go
```diff
@@ -40,24 +46,98 @@ func (h *Handler) handle(path string, handler http.Handler) {
...
+	r, _, err := h.store.OpenBlob(curl)
+	defer r.Close()
+	hash := sha256.New()
+	_, err = io.Copy(hash, r)
+	if err != nil {
+		return "", errgo.Notef(err, "cannot calculate sha256 of archive")
+	}
+	sum256 := fmt.Sprintf("%x", hash.Sum(nil))
+
+	// Update the entry asynchronously because it doesn't matter
```
frankban, line 136: Thanks for this comment.

internal/legacy/api_test.go
```diff
@@ -106,17 +112,187 @@ func (s *APISuite) TestCharmArchiveUnresolvedURL(c *gc.C) {
...
+			Sha256:       test.sha,
+			Revision:     test.revision,
+			Digest:       test.digest,
+		}
+		if test.err != "" {
+			expectInfo.Errors = []string{test.err}
+		}
+		storetesting.AssertJSONCall(c, storetesting.JSONCallParams{
+			Handler:      s.srv,
+			URL:          "/charm-info?charms=" + test.url,
```
frankban, line 175: What do you think about adding a separate test with multiple charms?

internal/legacy/api_test.go
```diff
@@ -106,17 +112,187 @@ func (s *APISuite) TestCharmArchiveUnresolvedURL(c *gc.C) {
...
+			ExpectBody: map[string]charm.InfoResponse{
+				test.url: expectInfo,
+			},
+		})
+	}
+
+	// TODO(rog) check stats events
+}
+
+func fileSHA256(c *gc.C, path string) string {
```
frankban, line 186: mustFileSHA256?

internal/legacy/api_test.go
```diff
@@ -106,17 +112,187 @@ func (s *APISuite) TestCharmArchiveUnresolvedURL(c *gc.C) {
...
+	ch, err := charm.Store.Get(url)
+	c.Assert(err, gc.IsNil)
+	chArchive := ch.(*charm.CharmArchive)
+
+	data, err := ioutil.ReadFile(chArchive.Path)
+	c.Assert(err, gc.IsNil)
+	c.Assert(data, gc.DeepEquals, archiveBytes)
+}
+
+func (s *APISuite) TestCharmPackageCharmInfo(c *gc.C) {
```
frankban, line 216: ... and here it is ;-)

internal/legacy/export_test.go
```diff
@@ -0,0 +1,3 @@
+package legacy
```
frankban, line 4: Copyright headers?

internal/legacy/api_test.go
```diff
@@ -106,17 +112,187 @@ func (s *APISuite) TestCharmArchiveUnresolvedURL(c *gc.C) {
...
+			},
+		},
+	})
+
+	select {
+	case <-updated:
+	case <-time.After(5 * time.Second):
+		c.Fatalf("timed out waiting for update")
+	}
+
```
frankban, line 273: AT this point we should be able to check that the sha256 is also stored in the document.

internal/legacy/api_test.go
```diff
@@ -106,17 +112,187 @@ func (s *APISuite) TestCharmArchiveUnresolvedURL(c *gc.C) {
...
+		ExpectBody: map[string]charm.InfoResponse{
+			wordpressURL.String(): {
+				CanonicalURL: wordpressURL.String(),
+				Sha256:       sum256,
+				Revision:     0,
+			},
+		},
+	})
+
+	select {
```
frankban, line 268: Very cool.

frankban:
This looks good Roger, thanks for updating this branch as suggested by Uros!
:+1: with some minor comments.

internal/legacy/api.go
rogpeppe, line 97: That doesn't work, because we want to append the error if updateEntitySHA256 fails too. I've followed the pattern of the old charmstore code here, from which this code is derived.

internal/legacy/api_test.go
rogpeppe, line 175: See below :-)

internal/legacy/api_test.go
rogpeppe, line 186: I don't care much. The fact that we're passing a gc.C as parameter
is a reasonable indication, I think.

internal/legacy/api_test.go
rogpeppe, line 273: The fact that we're getting the correct SHA256 as a response without calling updateEntitySHA256 should give reasonable assurance of that fact, I think.
If it hadn't stored the sha256 in the document, how could it know that it
shouldn't update it?

internal/legacy/export_test.go
rogpeppe, line 4: done.

internal/legacy/api.go
rogpeppe, line 108: done

internal/legacy/api.go
rogpeppe, line 92: Moved the comment.

internal/legacy/api.go
urosj, line 108: Thanks.

internal/legacy/api.go
```diff
@@ -40,24 +46,98 @@ func (h *Handler) handle(path string, handler http.Handler) {
...
+			entity.BlobHash256, err = h.updateEntitySHA256(curl)
+		}
+		if err == nil {
+			// TODO(rog) increment charm-info stats counter?
+			c.CanonicalURL = curl.String()
+			c.Sha256 = entity.BlobHash256
+			c.Revision = curl.Revision
+			// TODO(rog) include Digest if it exists in extra-info ?
+		} else {
+			// TODO(rog) increment charm-missing stats counter?
```
urosj, line 110: Adding a card for this as well.

urosj:
:+1: Looks much cleaner, thanks!

rogpeppe:
:shipit:

jujugui:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/220/

PR #103, opened 2014-09-16
charmloader: log fails to file

jrwren:
Log all charms and bundles that failed the ingestion process. With err descriptions if possible.

### DIFF
--- a/cmd/charmload/main.go
+++ b/cmd/charmload/main.go
@@ -51,7 +51,7 @@ func load() error {
 			return errgo.Notef(err, "cannot configure loggers")
 		}
 	}
-	writer, err := os.OpenFile("charmload-err", os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
+	writer, err := os.OpenFile("charmload.err", os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
 	if err != nil {
 		return err
 	}
--- a/cmd/charmload/main.go
+++ b/cmd/charmload/main.go
@@ -25,6 +25,7 @@ import (
 )
 
 var logger = loggo.GetLogger("charmload_v4")
+var failsLogger = loggo.GetLogger("charmload_v4.loadfails")
 
 func main() {
 	err := load()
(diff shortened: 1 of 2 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/235/

cmd/charmload/main.go
```diff
@@ -62,6 +63,15 @@ func load() error {
 			return err
 		}
 	}
+	writer, err := os.OpenFile("charmload-err", os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
```
urosj: Could you add descriptions in a README which files and where different events are logged? Tnx.

urosj:
:+1: (with minor comments) It does exactly what it is intended to.

cmd/charmload/main.go
```diff
@@ -62,6 +63,15 @@ func load() error {
...
+	if err != nil {
+		return err
+	}
+	defer writer.Close()
+	registerLoadFailsWriter(writer)
```
rogpeppe, line 59: I'm not sure that the custom writer is necessary.
How about just doing this:

```
err := loggo.RegisterWriter("errors", writer, loggo.ERROR)
```

?
Then any log messages of error and above will get written
to this file, which I think is what we want here.

Note - as written, the condition below would never get
triggered, because err wasn't being assigned.

rogpeppe, line 59: Also, we wouldn't then need the extra logger instance - just use any logger.

jrwren, line 59: The separate logger was intentional to "Log all charms and bundles that failed the ingestion process. With err descriptions if possible." 
1. There is another place in this code which logs an error, and I do feel it is an error, but it is not an error publishing.
2.  We cannot control other loggo loggers, for example juju.charms logger as created and used by juju/charms package. Admittedly, this particular logger never writes at the Error level, but the intent would be lost.

The intent as I understood it was to have a separate log file with only 1 kind of error message. This probably could have been written a little better in the card.

Too many years of log4net, log4j, nlog, python logging for me, I guess.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/253/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/254/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/256/

jrwren:
:shipit:

jujugui:

PR #104, opened 2014-09-17
internal/v4: implement changes/published

rogpeppe:
As recently added to the specification.

### DIFF
--- a/internal/v4/api.go
+++ b/internal/v4/api.go
@@ -9,7 +9,9 @@ import (
 	"net/http"
 	"net/url"
 	"sort"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/juju/errgo"
 	"gopkg.in/juju/charm.v3"
@@ -41,6 +43,7 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
 			"search/interesting": http.HandlerFunc(h.serveSearchInteresting),
 			"stats/":             router.NotFoundHandler(),
 			"stats/counter/":     router.HandleJSON(h.serveStatsCounter),
+			"changes/published":  router.HandleJSON(h.serveChangesPublished),
 		},
 		Id: map[string]router.IdHandler{
 			"archive":   h.serveArchive,
@@ -489,4 +492,60 @@ func (h *Handler) metaArchiveUploadTime(entity *mongodoc.Entity, id *charm.Refer
 	return &params.ArchiveUploadTimeResponse{
 		UploadTime: entity.UploadTime.UTC(),
 	}, nil
+}
+
+type PublishedResponse struct {
+	Id        *charm.Reference
+	Published time.Time
+}
+
+// GET changes/published[?limit=$count][&from=$fromdate][&to=$todate]
+// http://tinyurl.com/qx5zdee
+func (h *Handler) serveChangesPublished(w http.ResponseWriter, r *http.Request) (interface{}, error) {
+	start, stop, err := parseDateRange(r.Form)
+	if err != nil {
+		return nil, errgo.Mask(err, errgo.Is(params.ErrBadRequest))
+	}
+	limit := -1
+	if limitStr := r.Form.Get("limit"); limitStr != "" {
+		limit, err = strconv.Atoi(limitStr)
+		if err != nil {
+			return nil, badRequestf(nil, "invalid 'limit' value")
+		}
+	}
+	var tquery bson.D
+	if !start.IsZero() {
+		tquery = make(bson.D, 0, 2)
+		tquery = append(tquery, bson.DocElem{
+			Name:  "$gte",
+			Value: start,
+		})
+	}
+	if !stop.IsZero() {
+		tquery = append(tquery, bson.DocElem{
+			Name:  "$lte",
+			Value: stop,
+		})
+	}
+	var findQuery bson.D
+	if len(tquery) > 0 {
+		findQuery = bson.D{{"uploadtime", tquery}}
+	}
+	query := h.store.DB.Entities().
+		Find(findQuery).
+		Sort("-uploadtime").
+		Select(bson.D{{"_id", 1}, {"uploadtime", 1}})
+	if limit != -1 {
+		query = query.Limit(limit)
+	}
+
+	results := []params.Published{}
+	var entity mongodoc.Entity
+	for iter := query.Iter(); iter.Next(&entity); {
+		results = append(results, params.Published{
+			Id:          entity.URL,
+			PublishTime: entity.UploadTime,
+		})
+	}
+	return results, nil
 }
--- a/internal/v4/api.go
+++ b/internal/v4/api.go
@@ -38,12 +38,12 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
 	}
 	h.Router = router.New(&router.Handlers{
 		Global: map[string]http.Handler{
+			"changes/published":  router.HandleJSON(h.serveChangesPublished),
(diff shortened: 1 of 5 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/236/

internal/v4/api.go
```diff
@@ -41,6 +43,7 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
 			"search/interesting": http.HandlerFunc(h.serveSearchInteresting),
 			"stats/":             router.NotFoundHandler(),
 			"stats/counter/":     router.HandleJSON(h.serveStatsCounter),
+			"changes/published":  router.HandleJSON(h.serveChangesPublished),
```
urosj: Can we keep the API ordered, please?

internal/v4/api_test.go
```diff
@@ -871,6 +871,93 @@ func (s *APISuite) TestMetaStats(c *gc.C) {
...
+	time: "5432-10-12 00:00",
+}, {
+	id:   "cs:precise-mysql-1",
+	time: "5432-10-12 13:00",
+}, {
+	id:   "cs:precise-wordpress-2",
+	time: "5432-10-12 23:59",
+}, {
+	id:   "cs:precise-mysql-2",
+	time: "5432-10-13 00:00",
```
urosj, line 899: Nice, both daily borderline examples included.

internal/v4/api_test.go
```diff
@@ -871,6 +871,93 @@ func (s *APISuite) TestMetaStats(c *gc.C) {
...
+}, {
+	id:   "cs:precise-mysql-5",
+	time: "5432-10-13 10:00",
+}, {
+	id:   "cs:precise-wordpress-3",
+	time: "5432-10-14 01:00",
+}}
```
urosj, line 906: Maybe we should add tests for different time zones?

internal/v4/api_test.go
```diff
@@ -871,6 +871,93 @@ func (s *APISuite) TestMetaStats(c *gc.C) {
...
+	args:   "?start=5432-10-13&stop=5432-10-13",
+	expect: []int{4, 3},
+}, {
+	args:   "?start=5432-10-12&stop=5432-10-13",
+	expect: []int{4, 3, 2, 1, 0},
+}, {
+	args:   "?start=5432-10-13&stop=5432-10-12",
+	expect: []int{},
+}, {
+	args:   "?limit=3",
```
urosj, line 932: I'd add ?limit=-9999 as well.

internal/v4/api.go
```diff
@@ -490,3 +493,59 @@ func (h *Handler) metaArchiveUploadTime(entity *mongodoc.Entity, id *charm.Refer
...
+// http://tinyurl.com/qx5zdee
+func (h *Handler) serveChangesPublished(w http.ResponseWriter, r *http.Request) (interface{}, error) {
+	start, stop, err := parseDateRange(r.Form)
+	if err != nil {
+		return nil, errgo.Mask(err, errgo.Is(params.ErrBadRequest))
+	}
+	limit := -1
+	if limitStr := r.Form.Get("limit"); limitStr != "" {
+		limit, err = strconv.Atoi(limitStr)
+		if err != nil {
```
urosj: if err != nill or limit < -1?

internal/v4/stats.go
```diff
@@ -38,6 +39,30 @@ func entityStatsKey(url *charm.Reference, kind string) []string {
...
+		}
+	}
+	if v := form.Get("stop"); v != "" {
+		var err error
+		stop, err = time.Parse(dateFormat, v)
+		if err != nil {
+			return time.Time{}, time.Time{}, badRequestf(err, "invalid 'stop' value %q", v)
+		}
+		// Cover all timestamps within the stop day.
+		stop = stop.Add(24*time.Hour - 1*time.Second)
```
urosj, line 61: If i write that i want all charms until 15.9.2014 00:00, this will actually look until 16.9.2014 00:00?

internal/v4/api.go
```diff
@@ -490,3 +493,59 @@ func (h *Handler) metaArchiveUploadTime(entity *mongodoc.Entity, id *charm.Refer
...
+	}
+	var findQuery bson.D
+	if len(tquery) > 0 {
+		findQuery = bson.D{{"uploadtime", tquery}}
+	}
+	query := h.store.DB.Entities().
+		Find(findQuery).
+		Sort("-uploadtime").
+		Select(bson.D{{"_id", 1}, {"uploadtime", 1}})
+	if limit != -1 {
```
urosj, line 538: Needs a check from above.

internal/v4/api.go
```diff
@@ -490,3 +493,59 @@ func (h *Handler) metaArchiveUploadTime(entity *mongodoc.Entity, id *charm.Refer
 		UploadTime: entity.UploadTime.UTC(),
 	}, nil
 }
+
+type PublishedResponse struct {
+	Id        *charm.Reference
+	Published time.Time
+}
+
+// GET changes/published[?limit=$count][&from=$fromdate][&to=$todate]
```
mhilton, line 502: If you have a limit, would you not also want an offset so you can get the first 100 followed by the second 100?

internal/v4/api_test.go
mhilton, line 906: does the API not specify UTC? it should otherwise we should force a time-zone in the request. If we allow an api with ambiguous timezones will cause problems.

internal/v4/api.go
rogpeppe, line 502: If you're going to do that, the client can always specify a limit of 200...
There may be an issue, of course, if there are enough charms to
cause memory problems when all their ids and times are returned
to the client. We need to think some more about possible streaming
APIs for large results. I'm not that keen on the offset approach, as
it can result in missing/inconsistent data when the set of results
changes between the two requests.

For this particular request, the client can also limit things by specifying
the start date - I think it will be a while before the number of charms/bundles
published on a single day poses a memory problem.

internal/v4/api_test.go
rogpeppe, line 906: The API should indeed specify that the dates are interpreted in UTC.
The request does force a time-zone in the request - it's always UTC :-)

internal/v4/api.go
rogpeppe: good point. done.

internal/v4/api.go
rogpeppe: done (< 0)

internal/v4/stats.go
rogpeppe, line 61: You can't currently write that. When we provide support for RFC3339 ranges, this code will have to change (I suspect we'll probably allow only requests where both ends of the range are in RFC3339 format, or neither).

internal/v4/api_test.go
rogpeppe, line 932: Done. Added a bunch more tests for error conditions that I'd forgotten to do.

internal/v4/api.go
rogpeppe, line 538: Done

internal/v4/api.go
mhilton, line 502: OK makes sense

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/237/

mhilton:
:+1: LGTM

urosj:
:+1:

rogpeppe:
:shipit:

jujugui:

PR #105, opened 2014-09-17
charmload: some refactoring to address review comments

rogpeppe:
We add the capability for charmload to take parameters from the same
config file that charmd uses, splitting the config reading code into a separate
package to enable that.

We make the temp file for the archive inside the existing temp directory
for a slightly simplification of the code.

Various other changes as suggested in review.

Some tests will arrive in the next PR - for now, I've done
some verification on the command line.

### DIFF
--- a/cmd/charmd/main.go
+++ b/cmd/charmd/main.go
@@ -5,16 +5,15 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"gopkg.in/mgo.v2"
-	"gopkg.in/yaml.v1"
 
 	"github.com/juju/charmstore"
+	"github.com/juju/charmstore/config"
+	"github.com/juju/charmstore/internal/debug"
 )
 
 func main() {
@@ -35,11 +34,8 @@ func serve() error {
 	if confPath == "" {
 		return fmt.Errorf("usage: %s <config path>", filepath.Base(os.Args[0]))
 	}
-	conf, err := readConfig(confPath)
+	conf, err := config.Read(confPath)
 	if err != nil {
-		return err
-	}
-	if err := conf.validate(); err != nil {
 		return err
 	}
 	session, err := mgo.Dial(conf.MongoURL)
@@ -56,50 +52,5 @@ func serve() error {
 	if err != nil {
 		return err
 	}
-	return http.ListenAndServe(conf.APIAddr, server)
-}
-
-type config struct {
-	MongoURL     string `yaml:"mongo-url"`
-	APIAddr      string `yaml:"api-addr"`
-	AuthUsername string `yaml:"auth-username"`
-	AuthPassword string `yaml:"auth-password"`
-}
-
-func (c *config) validate() error {
-	var missing []string
-	if c.MongoURL == "" {
-		missing = append(missing, "mongo-url")
-	}
-	if c.APIAddr == "" {
-		missing = append(missing, "api-addr")
-	}
-	if c.AuthUsername == "" {
-		missing = append(missing, "auth-username")
-	}
-	if c.AuthPassword == "" {
-		missing = append(missing, "auth-password")
-	}
-	if len(missing) != 0 {
-		return fmt.Errorf("missing fields %s in config file", strings.Join(missing, ", "))
-	}
-	return nil
-}
-
-func readConfig(path string) (*config, error) {
-	f, err := os.Open(path)
-	if err != nil {
-		return nil, fmt.Errorf("opening config file: %v", err)
-	}
-	defer f.Close()
-	data, err := ioutil.ReadAll(f)
-	if err != nil {
-		return nil, fmt.Errorf("reading config file: %v", err)
-	}
-	var conf config
-	err = yaml.Unmarshal(data, &conf)
-	if err != nil {
-		return nil, fmt.Errorf("processing config file: %v", err)
-	}
-	return &conf, nil
+	return http.ListenAndServe(conf.APIAddr, debug.Handler("", server))
 }
--- a/cmd/charmload/main.go
+++ b/cmd/charmload/main.go
@@ -312,6 +312,9 @@ func (cl *charmLoader) doCharmStoreRequest(req *http.Request, result interface{}
 		return &perr
 	}
 	if err := dec.Decode(result); err != nil {
+		// TODO(rog) return a more informative error in this case
+		// (we might actually be talking to a proxy, which may
+		// return any sort of error)
 		return errgo.Notef(err, "cannot unmarshal response")
 	}
 	return nil
--- a/cmd/charmload/main.go
+++ b/cmd/charmload/main.go
@@ -3,8 +3,8 @@ package main
 import (
 	"bytes"
 	"crypto/sha512"
-
 	"encoding/base64"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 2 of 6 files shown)


jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/239/

config/config.go
```diff
@@ -0,0 +1,70 @@
...
+// the charm store.
+package config
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"gopkg.in/errgo.v1"
```
jrwren: The rest of to packages in the tree use github.com/juju/errgo can we do that here too?

rogpeppe: Yeah, this was a mistake (goimports chose the wrong path). Fixed.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/240/

cmd/charmload/main.go
```diff
@@ -84,36 +122,38 @@ func load() error {
...
 			logger.Errorf("skipping branch %v with no revisions", tip.UniqueName)
 			continue
-		} else {
-			logger.Debugf("found %v with revision %v", tip.UniqueName, tip.Revision)
 		}
-		URLs := []*charm.URL{charmURL}
-		schema, name := charmURL.Schema, charmURL.Name
-		URLs = addPromulgatedCharmURLs(tip.OfficialSeries, schema, name, URLs)
-		err = publishBazaarBranch(*storeURL, *storeUser, URLs, branchURL, tip.Revision)
+		logger.Infof("found %v with revision %v", tip.UniqueName, tip.Revision)
```
jrwren, line 132: Preferred the Debug level on this message.

cmd/charmload/main.go
```diff
@@ -134,7 +174,7 @@ func addPromulgatedCharmURLs(officialSeries []string, schema, name string, URLs
 // For testing purposes, if name has a prefix preceding a string in
 // this format, the prefix is stripped out for computing the charm
 // URL, and the unique name is returned unchanged as the branch URL.
-func uniqueNameURLs(name string) (branchURL string, charmURL *charm.URL, err error) {
+func uniqueNameURLs(name string) (branchURL string, charmURL *charm.Reference, err error) {
```
jrwren, line 177: Since they are References now, maybe rename the function to match.  unqiueNameReferences?

cmd/charmload/main.go
```diff
@@ -84,36 +122,38 @@ func load() error {
...
 	}
 	return nil
 }
 
-// addPromulgatedCharmURLs adds urls from officialSeries to
+// appendPromulgatedCharmURLs adds urls from officialSeries to
 // the URLs slice for the given schema and name.
 // Promulgated charms have OfficialSeries in launchpad.
-func addPromulgatedCharmURLs(officialSeries []string, schema, name string, URLs []*charm.URL) []*charm.URL {
+func appendPromulgatedCharmURLs(officialSeries []string, schema, name string, URLs []*charm.Reference) []*charm.Reference {
```
jrwren, line 154: Same as below, maybe rename the function to appendPromulgatedCharmReferences.

cmd/charmload/main.go
```diff
@@ -170,109 +210,111 @@ func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, b
...
 	if err != nil {
-		return err
+		return errgo.Notef(err, "cannot get revision id")
 	}
 	if tipDigest != digest {
 		digest = tipDigest
 		logger.Warningf("tipDigest %v != digest %v", digest, tipDigest)
 	}
-
-	logger.Tracef("read CharmDir from branchDir %v", branchDir)
```
jrwren, line 180: It was trace level and I did find a tiny bit of value in it when trace debugging.

cmd/charmload/main.go
```diff
@@ -170,109 +210,111 @@ func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, b
...
-// and computes the hash needed for posting calls stat on the file to get its
-// size.
-func getPostDataFromCharmDir(charmDir *charm.CharmDir) (tempFile *os.File, hash string, size int64, err error) {
-	sha384 := sha512.New384()
-	tempFile, err = ioutil.TempFile("", "publish-charm")
-	logger.Debugf("writing charm to temporary file %s", tempFile.Name())
+// archiveCharmDir archives the charmDir to a temporary file
+// inside tempDir and returns the file, its hash and size.
+func (cl *charmLoader) archiveCharmDir(charmDir *charm.CharmDir, tempDir string) (archiveFile *os.File, hash string, size int64, err error) {
+	f, err := os.Create(filepath.Join(tempDir, "archive.zip"))
```
jrwren, line 244: Arbitrary change. I tend to dislike arbitrary changes because when looking through a changelog one may get the impression that there was a reason for it, unless I'm missing something and there is a good reason to use the branch dir, in which case I like to see that mentioned in the commit message.

cmd/charmload/main.go
```diff
@@ -170,109 +210,111 @@ func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, b
...
+		return nil, "", 0, errgo.Notef(err, "cannot make temp file")
 	}
-	_, err = tempFile.Seek(0, 0)
+	defer func() {
+		if err != nil {
+			f.Close()
+		}
+	}()
+	sha384 := sha512.New384()
+	err = charmDir.ArchiveTo(io.MultiWriter(f, sha384))
```
jrwren, line 258: Much nicer, Thanks.

cmd/charmload/main.go
```diff
@@ -13,68 +13,106 @@ import (
...
 )
 
-var logger = loggo.GetLogger("charmload_v4")
+var logger = loggo.GetLogger("charmload")
+
+var (
+	staging       = flag.Bool("staging", false, "use the launchpad staging server")
+	storeAddr     = flag.String("storeaddr", "localhost:8080", "the address of the charmstore; overrides configuration file")
+	loggingConfig = flag.String("logging-config", "", "specify log levels for modules e.g. <root>=TRACE")
+	storeUser     = flag.String("u", "", "the colon separated user:password for charmstore; overrides configuration file")
```
jrwren, line 34: admin:example-passwd default was nice default when developing as it matched developing default of charmd.

jrwren:
--show-log not being default was intentional. Once in production and in other non-developing and debugging use cases, the only logs were to forthcoming loadfails logger.

storeurl being a url including the /v4 endpoint was intentional. The use case is when customers run charmstores they may front them with a proxy balancer and remap the endpoint. This use case is probably not needed. I wanted to mention it because its something which no longer functions.

regarding charm.Reference v. charm.URL, the branchtips returned from launchpad do always have a series. I think the change from URL to Reference is effectively changing nothing.

Overall, its a great cleanup. Huge Thanks.

cmd/charmload/main.go
rogpeppe, line 34: Unfortunately that conflicts somewhat with having -u override the user/password from the config file (you can't tell if -u has been specified). We _could_ still apply a default, but since no-one should be using the default user name and password anyway and when testing it's easier just to name the same config file that charmd was started from, I don't think it matters much.

cmd/charmload/main.go
rogpeppe, line 132: It seems reasonable to have it at higher than debug level to me - it's information that's useful even when we're not debugging the actual program logic, I think.

cmd/charmload/main.go
rogpeppe, line 154: We use URL instead of Reference in names elsewhere in the code, even though we use charm.Reference everywhere (I consider charm.URL to be more hassle than it's worth), so this has precedent. I might be tempted to normalise all those names to use "Id" rather than URL, but that would be another PR - for the time being I think it's good to be (at least slightly) more consistent.

cmd/charmload/main.go
rogpeppe, line 180: I tend to err towards less debug messages when the code is debugged. This just gives us the temp directory name, which isn't really _that_ useful, I think. I'm open to arguments to the contrary though.

cmd/charmload/main.go
rogpeppe, line 244: I made the change because it simplified the logic - it means the caller doesn't need to bother about explicitly removing the temporary file because it's in the temporary directory which is removed as a matter of course. It also makes it slightly tidier when the charmload process is interrupted - there's only one kind of thing that's left behind in /tmp.

rogpeppe, line 244: Updated the description.

rogpeppe:
Thanks for the review!

> --show-log not being default was intentional. Once in production and in other
> non-developing and debugging use cases, the only logs were to forthcoming loadfails logger.

I think it's reasonable to log errors and warnings (the default). The logging-config flag
can still be used to disable logging when desired.

> storeurl being a url including the /v4 endpoint was intentional. The use
> case is when customers run charmstores they may front them with a proxy
> balancer and remap the endpoint. This use case is probably not needed. I
> wanted to mention it because its something which no longer functions.

I appreciate that use case, but for the time being I think it's simpler to
just specify the address, because the user doesn't have to know the
form of store URLs, and there's no possibility of them getting it wrong
(the old code would fail if the user didn't specify the final "/" on the
store URL, for example).

If we ever need the capability to address arbitrary URLs, the
change is simple to make.

> regarding charm.Reference v. charm.URL, the branchtips returned from
> launchpad do always have a series. I think the change from URL to
> Reference is effectively changing nothing.

Agreed. I changed it only for consistency with the rest of the store
code - we don't use charm.URL anywhere.

cmd/charmload/main.go
```diff
@@ -170,109 +210,111 @@ func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, b
...
 
-// doCharmStorePost performs the request and returns an error or nil
-// if the post request succeeded.
-func doCharmStorePost(request *http.Request) error {
-	resp, err := http.DefaultClient.Do(request)
-	if resp.StatusCode == http.StatusUnauthorized {
-		return &UnauthorizedError{}
+// doCharmStoreRequest adds appropriate headers to the given HTTP request,
+// sends it to the charm store acting as the given user
+//  and parses the result as JSON into the
```
frankban, line 290: Unnecessary space at the beginning of the comment.

frankban:
:+1: thanks for this!

cmd/charmd/main.go
```diff
@@ -5,16 +5,15 @@ package main
...
 import (
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"gopkg.in/mgo.v2"
-	"gopkg.in/yaml.v1"
```
mhilton, line 15: Through sheer curiosity, do we have a team guideline on using gopkg.in?

cmd/charmd/main.go
```diff
@@ -35,13 +34,10 @@ func serve() error {
...
 	}
-	conf, err := readConfig(confPath)
+	conf, err := config.Read(confPath)
 	if err != nil {
 		return err
 	}
-	if err := conf.validate(); err != nil {
-		return err
-	}
 	session, err := mgo.Dial(conf.MongoURL)
```
mhilton, line 41: This is not relevent to this change, but I'm not sure whether this is the best thing to do here. why bother to check the mongoDB connection on startup? It could get dropped at any time so we have to cope with that anyway. does it really matter if we can't connect at startup?

mhilton, line 41: Or do we, perhaps I should investigate further before opining.

cmd/charmload/main.go
```diff
@@ -170,109 +210,111 @@ func publishBazaarBranch(storeURL string, storeUser string, URLs []*charm.URL, b
...
-				}
-				var buffer bytes.Buffer
-				io.CopyN(&buffer, resp.Body, int64(bodySize))
-				body = buffer.String()
-			}
+	dec := json.NewDecoder(resp.Body)
+	if resp.StatusCode != http.StatusOK {
+		var perr params.Error
+		if err := dec.Decode(&perr); err != nil {
+			return errgo.Notef(err, "cannot unmarshal error response")
```
mhilton, line 307: This is quite likely if the error comes back from apache, squid or haproxy. In those cases we might want to know what HTTP status was or something to help debug it?

config/config.go
```diff
@@ -0,0 +1,70 @@
...
+package config
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/juju/errgo"
+	"gopkg.in/yaml.v1"
```
mhilton, line 15: Is there a scheme for arranging imports? This is at least inconsistent with  cmd/charmd/main.go above. That separates github.com/juju and gopkg.in by a line.

mhilton:
I have a few questions but it looks fine. :+1:

cmd/charmload/main.go
rogpeppe, line 307: That's an interesting point, thanks. We could return a more informative error in that case. I'll add a TODO.

config/config.go
rogpeppe, line 15: The scheme is that imports go into three groups: stdlib, non-stdlib and project-local.
In this case the local project is github.com/juju/charmstore.

   go get github.com/rogpeppe/sortimports

gets you a tool that automatically sorts imports for you in this
way.

cmd/charmload/main.go
rogpeppe, line 307: (Although in practice, I think it's unlikely we'll be talking to a proxy - this thing probably
almost always run on the same machine as a charm store server.)

rogpeppe:
:shipit:

jujugui:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/241/

cmd/charmload/main.go
rogpeppe, line 290: oops, forgot to address this. will do in next PR.

cmd/charmd/main.go
rogpeppe, line 41: We actually use the connection - we pass it into charmstore.NewServer (indirectly - a mgo.DB refers to the session it's derived from), so we do need to dial the mongo server here.

PR #109, opened 2014-09-19
internal/v4: implement debug/status endpoint

rogpeppe:
As newly documented in the charm store.
Not all entries are implemented yet.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -19,10 +19,6 @@ import (
 	"github.com/juju/charmstore/params"
 )
 
-// StartTime holds the time that the server executable
-// started running.
-var StartTime = time.Now()
-
 // Store represents the underlying charm and blob data stores.
 type Store struct {
 	DB        StoreDatabase
@@ -354,6 +350,8 @@ func (s StoreDatabase) Entities() *mgo.Collection {
 	return s.C("entities")
 }
 
+// allCollections holds for each collection used by the charm store a
+// function returns that collection.
 var allCollections = []func(StoreDatabase) *mgo.Collection{
 	StoreDatabase.StatCounters,
 	StoreDatabase.StatTokens,
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -19,6 +19,10 @@ import (
 	"github.com/juju/charmstore/params"
 )
 
+// StartTime holds the time that the server executable
+// started running.
+var StartTime = time.Now()
+
 // Store represents the underlying charm and blob data stores.
 type Store struct {
 	DB        StoreDatabase
@@ -348,4 +352,20 @@ func (s StoreDatabase) Close() {
 // Entities returns the mongo collection where entities are stored.
 func (s StoreDatabase) Entities() *mgo.Collection {
 	return s.C("entities")
+}
+
+var allCollections = []func(StoreDatabase) *mgo.Collection{
+	StoreDatabase.StatCounters,
+	StoreDatabase.StatTokens,
+	StoreDatabase.Entities,
+}
+
+// Collections returns a slice of all the collections used
+// by the charm store.
+func (s StoreDatabase) Collections() []*mgo.Collection {
+	cs := make([]*mgo.Collection, len(allCollections))
+	for i, f := range allCollections {
+		cs[i] = f(s)
+	}
+	return cs
 }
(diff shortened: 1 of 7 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/252/

internal/charmstore/store.go
```diff
@@ -19,6 +19,10 @@ import (
 	"github.com/juju/charmstore/params"
 )
 
+// StartTime holds the time that the server executable
+// started running.
+var StartTime = time.Now()
```
jrwren: Could call the variable ServerExecutableStartTime or ServerStartTime so that it is obvious at use cite.

internal/v4/status.go
```diff
@@ -0,0 +1,106 @@
...
+		return "All required collections exist", true
+	}
+	return fmt.Sprintf("Missing collections: %s", missing), false
+}
+
+func (h *Handler) checkEntities() (string, bool) {
+	iter := h.store.DB.Entities().Find(nil).Iter()
+	charms, bundles, promulgated := 0, 0, 0
+	var entity mongodoc.Entity
+	for iter.Next(&entity) {
```
jrwren, line 90: Can this be written in a mongo side run aggregate?

I don't know mongo well enough, but I know in SQL, I'd much prefer a

```
SELECT COUNT() FROM Entities WHERE URLSeries='bundle'
```

and
    SELECT COUNT() FROM Entities WHERE URLSeries!='bundle'
and
    SELECT COUNT() FROM Entities WHERE URLUser=""

3 queries which can utilize indexes rather than iterating the entire table.

jrwren:
Thanks, Looks great modulo some comments above.

internal/v4/status.go
urosj, line 90: +1 on this and that's also the reason why original specs included three different entries. But we can mark this as an improvement, not a blocker.

internal/charmstore/store.go
urosj: And even add it in a DebugStatus struct?

internal/charmstore/store.go
```diff
@@ -349,3 +353,19 @@ func (s StoreDatabase) Close() {
 func (s StoreDatabase) Entities() *mgo.Collection {
 	return s.C("entities")
 }
+
+var allCollections = []func(StoreDatabase) *mgo.Collection{
```
urosj, line 355: Please add comment describing its purpose.

internal/v4/api.go
```diff
@@ -40,6 +40,7 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
 		Global: map[string]http.Handler{
 			"changes/published":  router.HandleJSON(h.serveChangesPublished),
 			"debug":              http.HandlerFunc(h.serveDebug),
+			"debug/status":       router.HandleJSON(h.serveDebugStatus),
```
urosj, line 43: Thanks for preserving the order :)

internal/charmstore/store.go
rogpeppe: It looks like charmstore.StartTime in the code, which i think is reasonable.
Also, I deliberately don't want it to be necessarily always the server
executable start time - I guess it might be possible that some executable
might be running for ages and then suddenly decide to start the charm store...

Perhaps that's a silly thing to contemplate though and maybe this variable would be best off declared locally inside v4 in fact.

internal/v4/status.go
rogpeppe, line 90: None of the queries in question could use existing indexes. We could add another field (a Promulgated bool) and three more indexes, but maintaining all those indexes just for this call seems like unnecessary overhead on the rest of the charm store operations. Is it faster to do one linear scan over the network or three linear scans remotely? I don't know. The other advantage the current approach brings is the possibility of finding entries that don't fit into any categories, which could be a fourth entry and would be unlikely to be indexable.

urosj:
:+1: Thank you.

internal/charmstore/store.go
rogpeppe, line 355: done.

rogpeppe:
:shipit:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/255/

jujugui:

PR #110, opened 2014-09-19
Ingest bundles

jrwren:


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/259/

lppublish/lpad.go
```diff
@@ -74,6 +74,22 @@ func PublishCharmsDistro(p Params) error {
 	return cl.run()
 }
 
+func (cl *charmLoader) moveBundleTipsToEndOfList(tips []lpad.BranchTip) []lpad.BranchTip {
+	bundleTips := make([]lpad.BranchTip, 0, 0)
```
rogpeppe: var bundleTips []lpad.BranchTip

(and below)

lppublish/lpad.go
```diff
@@ -283,9 +311,13 @@ func (cl *charmLoader) publishBazaarBranch(URLs []*charm.Reference, branchURL st
 	return err
 }
 
-// archiveCharmDir archives the charmDir to a temporary file
+type archiveToer interface {
```
rogpeppe: I'd suggest archiverTo as a more conventional name (c.f. io.ReaderFrom)

lppublish/lpad.go
```diff
@@ -141,8 +158,7 @@ type entityResult struct {
 func (cl *charmLoader) processTips(tips []lpad.BranchTip, results chan<- entityResult) {
```
rogpeppe, line 161: Rather than shuffling the contents of the in moveBundleTipsToEndOfList, I think
it might be simpler and perhaps more obvious to split this function into two
and iterate over the original list twice. This keeps processTips as the canonical
place that processes the tips.

For example:

```
func (cl *charmLoader) processTips(tips []lpad.BranchTip, results chan<- entityResult) {
    counter := 0
    // process all charms before bundles so that the
    // bundles will verify correctly if they refer to the charms.
    for _, tip := range tips {
         if strings.HasSuffix(tip.UniqueName, "/trunk") {
             cl.processTip(tip, results, &counter)
        }
    }
    for _, tip := range tips {
        if strings.HasSuffix(tip.Unique, "/bundle") {
             cl.processTip(tip, results, &counter)
        }
    }
}

func (cl *charmLoader) processTip(tip lpad.BranchTip, results chan<- entityResult, counter *int) {
      if *counter >= cl.Limit {
          return
      }
      *counter++
      body of tip range loop goes here
}
```

Then moveBundleTipsToEndOfList can go and the code is arguable somewhat simpler.

lppublish/lpad.go
```diff
@@ -261,11 +277,23 @@ func (cl *charmLoader) publishBazaarBranch(URLs []*charm.Reference, branchURL st
...
-	charmDir, err := charm.ReadCharmDir(branchDir)
-	if err != nil {
-		return errgo.Notef(err, "cannot read charm dir")
+	var archiveDir archiveToer
+	if URLs[0].Series == "bundles" {
+		// charmstore expects series named bundle instead of bundles
+		for _, url := range URLs {
+			url.Series = "bundle"
+		}
+		archiveDir, err = charm.ReadBundleDir(branchDir)
```
rogpeppe, line 306: // TODO convert from old bundles format when this fails

?

rogpeppe:
Looks great in general, thanks, with a few minor suggestions..

lppublish/lpad.go
rogpeppe: Although, given what the code is doing,

```
bundleTips := make([]lpad.BranchTip, 0, len(tips))
```

would be more appropriate.

lppublish/lpad.go
```diff
@@ -74,6 +74,22 @@ func PublishCharmsDistro(p Params) error {
...
+	bundleTips := make([]lpad.BranchTip, 0, 0)
+	bundleTipsLast := make([]lpad.BranchTip, 0, 0)
+	for _, tip := range tips {
+		if strings.HasSuffix(tip.UniqueName, "/bundle") {
+			bundleTipsLast = append(bundleTipsLast, tip)
+			continue
+		}
+		bundleTips = append(bundleTips, tip)
+	}
+	for _, tip := range bundleTipsLast {
```
rogpeppe: For the record, this could be written:

```
 bundleTips = append(bundleTips, bundleTipsLast...)
```

Also for the record, it's possible to be more efficient (not that it matters here)
when moving a bunch of things to the end of a slice; for example:

```
lastBundle := len(tips)
for i := len(tips)-1; i >= 0; i-- {
    if strings.HasSuffix(tip.UniqueName, "/bundle") {
         tips[i], tips[lastBundle-1] = tips[lastBundle-1], tips[i]
         lastBundle--
    }
 }
```

lppublish/lpad.go
jrwren: done.

lppublish/lpad.go
jrwren: done.

lppublish/lpad.go
jrwren, line 161: processTip would need to return an error and some kind of sentinel to trigger the break if limit is reached or continue on UniqueName failing or Revision being blank, this adds enough complexity that I prefer the solution already here.

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/264/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/265/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/266/

lppublish/lpad.go
```diff
@@ -74,6 +74,20 @@ func PublishCharmsDistro(p Params) error {
...
 
+func (cl *charmLoader) moveBundleTipsToEndOfList(tips []lpad.BranchTip) []lpad.BranchTip {
+	bundleTips := make([]lpad.BranchTip, 0, len(tips))
+	var bundleTipsLast []lpad.BranchTip
+	for _, tip := range tips {
+		if strings.HasSuffix(tip.UniqueName, "/bundle") {
+			bundleTipsLast = append(bundleTipsLast, tip)
+			continue
+		}
+		bundleTips = append(bundleTips, tip)
```
urosj, line 90: I'd add "TODO improve/add control of memory usage"  here for future reference and add a card to the kanban.

urosj:
:+1: with a request in a comment.

lppublish/lpad.go
jrwren, line 90: Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/267/

lppublish/lpad.go
```diff
@@ -227,7 +242,7 @@ func uniqueNameURLs(name string) (branchURL string, charmURL *charm.Reference, e
 	} else {
 		branchURL = "lp:" + name
 	}
-	if len(u) < 5 || u[1] != "charms" || u[4] != "trunk" || len(u[0]) == 0 || u[0][0] != '~' {
+	if len(u) < 5 || u[1] != "charms" || (u[4] != "trunk" && u[4] != "bundle") || len(u[0]) == 0 || u[0][0] != '~' {
```
bac: This check is getting sufficiently complicated perhaps it should be a function.  Would allow things to be named and be more easily tested.

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/271/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/272/

jrwren:
:shipit:

jujugui:

lppublish/lpad.go
rogpeppe, line 161: Sorry, I only just saw this comment. That's actually not the case.
The continue becomes a return, and there's no need to break when the limit
is reached.

PR #114, opened 2014-09-24
Bundle quick and dirty fix

urosj:
Rename, remove first line (start with services)

### DIFF
--- a/lppublish/lpad.go
+++ b/lppublish/lpad.go
@@ -267,37 +267,32 @@ func notSupportedBranchName(u []string) bool {
 const bzrDigestKey = "bzr-digest"
 
 // Copies bundles.yaml to bundle.yaml without the first line.
+// TODO (rog, uros) Replace with proper bundle translation.
 func quickAndDirtyBundleFix(branchDir string) error {
-	//rename bundles.yaml to bundle.yaml
 	oldBundleYamlName := fmt.Sprint(branchDir, "/bundles.yaml")
 	newBundleYamlName := fmt.Sprint(branchDir, "/bundle.yaml")
 
 	file, err := os.Open(oldBundleYamlName)
 	if err != nil {
-		errgo.Notef(err, "could not open bundles.yaml")
+		return errgo.Notef(err, "could not open bundles.yaml")
 	}
 	defer file.Close()
 	newFile, err := os.Create(newBundleYamlName)
 	if err != nil {
-		errgo.Notef(err, "could not open bundle.yaml")
+		return errgo.Notef(err, "could not open bundle.yaml")
 	}
 	defer newFile.Close()
 
 	firstLine := true
 	scanner := bufio.NewScanner(file)
-	for scanner.Scan() {
-		if firstLine == false {
-			newFile.Write([]byte(scanner.Text()))
-			newFile.Write([]byte("\n"))
-		} else {
-			firstLine = false
-		}
+	r := bufio.NewReader(file)
+	_, _, err := r.ReadLine()
+	if err != nil {
+		return errgo.Newf("no first line")
 	}
-
-	if err := scanner.Err(); err != nil {
-		errgo.Notef(err, "reading error")
+	if _, err := io.Copy(newFile, r); err != nil {
+		return err
 	}
-
 	return nil
 }
 
@@ -339,7 +334,11 @@ func (cl *charmLoader) publishBazaarBranch(urls []*charm.Reference, branchURL st
 		for _, url := range urls {
 			url.Series = "bundle"
 		}
-		quickAndDirtyBundleFix(branchDir)
+		// TODO (rog, uros) Replace with proper bundle translation.
+		err := quickAndDirtyBundleFix(branchDir)
+		if err != nil {
+			return errgo.Notef(err, "quick bundle fix failed")
+		}
 
 		archiveDir, err = charm.ReadBundleDir(branchDir)
 		if err != nil {
--- a/lppublish/lpad.go
+++ b/lppublish/lpad.go
@@ -266,15 +266,11 @@ func notSupportedBranchName(u []string) bool {
 
 const bzrDigestKey = "bzr-digest"
 
+// Copies bundles.yaml to bundle.yaml without the first line.
 func quickAndDirtyBundleFix(branchDir string) error {
 	//rename bundles.yaml to bundle.yaml
 	oldBundleYamlName := fmt.Sprint(branchDir, "/bundles.yaml")
 	newBundleYamlName := fmt.Sprint(branchDir, "/bundle.yaml")
-
-	//err := os.Rename(oldBundleYamlName, newBundleYamlName)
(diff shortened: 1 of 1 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/277/

lppublish/lpad.go
```diff
@@ -265,6 +266,41 @@ func notSupportedBranchName(u []string) bool {
 
 const bzrDigestKey = "bzr-digest"
 
+// Copies bundles.yaml to bundle.yaml without the first line.
```
frankban, line 269: // quickAndDirtyBundleFix generates a bundle.yaml file from the old bundle spec found in bundles.yaml?

lppublish/lpad.go
```diff
@@ -265,6 +266,41 @@ func notSupportedBranchName(u []string) bool {
...
+func quickAndDirtyBundleFix(branchDir string) error {
+	//rename bundles.yaml to bundle.yaml
```
frankban: // Rename bundles.yaml to bundle.yaml.
Please add a space after // and use a complete sentence.
However, I am not sure this properly describes the goal of the lines below.

lppublish/lpad.go
```diff
@@ -265,6 +266,41 @@ func notSupportedBranchName(u []string) bool {
...
+	newFile, err := os.Create(newBundleYamlName)
+	if err != nil {
+		errgo.Notef(err, "could not open bundle.yaml")
+	}
+	defer newFile.Close()
+
+	firstLine := true
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		if firstLine == false {
```
frankban: This is indeed quick and dirty ;-)
Perhaps it would be better to parse the YAML and then Marshal it in the new file excluding the root node?
This way we don't have to assume YAML files don't start with empty lines for example.
I also have another question: is it possible that we encounter bundle branches with more than one bundle in their bundles.yaml? Or even using inheritance (IIRC no longer supported)? If that's the case, how do we handle those?

lppublish/lpad.go
```diff
@@ -265,6 +266,41 @@ func notSupportedBranchName(u []string) bool {
...
 
+// Copies bundles.yaml to bundle.yaml without the first line.
+func quickAndDirtyBundleFix(branchDir string) error {
+	//rename bundles.yaml to bundle.yaml
+	oldBundleYamlName := fmt.Sprint(branchDir, "/bundles.yaml")
+	newBundleYamlName := fmt.Sprint(branchDir, "/bundle.yaml")
+
+	file, err := os.Open(oldBundleYamlName)
+	if err != nil {
+		errgo.Notef(err, "could not open bundles.yaml")
```
frankban: return errgo...

lppublish/lpad.go
```diff
@@ -265,6 +266,41 @@ func notSupportedBranchName(u []string) bool {
...
+	}
+	defer file.Close()
+	newFile, err := os.Create(newBundleYamlName)
+	if err != nil {
+		errgo.Notef(err, "could not open bundle.yaml")
```
frankban: Return the error here too.

lppublish/lpad.go
```diff
@@ -265,6 +266,41 @@ func notSupportedBranchName(u []string) bool {
...
+		if firstLine == false {
+			newFile.Write([]byte(scanner.Text()))
+			newFile.Write([]byte("\n"))
+		} else {
+			firstLine = false
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		errgo.Notef(err, "reading error")
```
frankban: Return the error.

lppublish/lpad.go
```diff
@@ -303,6 +339,8 @@ func (cl *charmLoader) publishBazaarBranch(urls []*charm.Reference, branchURL st
 		for _, url := range urls {
 			url.Series = "bundle"
 		}
+		quickAndDirtyBundleFix(branchDir)
```
frankban: Here we should check for possible errors.

frankban:
Hey Uros, thanks for this branch!
I have a couple of questions and comments above.

lppublish/lpad.go
rogpeppe: I think you mean to return the error here.

Checking read errors is one thing, but it's more important to check
write errors.

lppublish/lpad.go
urosj, line 269: Yes. It just removes the first line stating a bundle name and starts with series. This covers most of current bundles.

lppublish/lpad.go
urosj: The proper way will be to use and modify YAML. Roger is doing that. The test here was how many bundles can we ingest with such a simple change. It turns out that most of them (if revisions are correct). 
The complex ones (more than one bundle in bundles.yaml) will fail, they were not part of this code, those require structural change - code that is in the working.

rogpeppe:
LGTM modulo simpler scan code and frankban's suggested fixes.

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/278/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/279/

lppublish/lpad.go
```diff
@@ -265,6 +266,34 @@ func notSupportedBranchName(u []string) bool {
...
+	}
+	defer file.Close()
+	newFile, err := os.Create(newBundleYamlName)
+	if err != nil {
+		return errgo.Notef(err, "could not open bundle.yaml")
+	}
+	defer newFile.Close()
+
+	r := bufio.NewReader(file)
+	_, _, nerr := r.ReadLine()
```
rogpeppe: No need to use a new variable name for the error here.

```
if _, _, err := r.ReadLine(); err != nil {
```

or

```
_, _, err = r.ReadLine()
```

lppublish/lpad.go
```diff
@@ -265,6 +266,34 @@ func notSupportedBranchName(u []string) bool {
...
 
+// Copies bundles.yaml to bundle.yaml without the first line.
+// TODO (rog, uros) Replace with proper bundle translation.
+func quickAndDirtyBundleFix(branchDir string) error {
+	oldBundleYamlName := fmt.Sprint(branchDir, "/bundles.yaml")
+	newBundleYamlName := fmt.Sprint(branchDir, "/bundle.yaml")
+
+	file, err := os.Open(oldBundleYamlName)
+	if err != nil {
+		return errgo.Notef(err, "could not open bundles.yaml")
```
rogpeppe, line 277: FYI the errors from os.Open and os.Create already contain the path name,
so errgo.Mask(err) would probably be fine in both cases.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/280/

PR #117, opened 2014-09-24
lppublish: make charm URLs correct earlier

rogpeppe:
This fixes a bug which would cause bundles to be processed
every time even when already in the store.


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/287/

jrwren:
:+1: 

Thanks for this. I had incorrectly thought that bundles were promulgated like charms, by using the OfficialSeries field in lpad api. They aren't and OfficialSeries is always empty for bundles so this works.

frankban:
:+1:

rogpeppe:
:shipit:

jujugui:

PR #118, opened 2014-09-25
Add es package

jrwren:
Looking for some feedback on this approach to talking to ES.

### DIFF
--- a/internal/elasticsearch/definitions/entity.json
+++ b/internal/elasticsearch/definitions/entity.json
@@ -1,0 +1,241 @@
+{
+    "entity" : {
+        "dynamic" : "false",
+        "properties" : {
+          "id" : {
+            "type" : "string",
+            "index" : "not_analyzed",
+            "omit_norms" : true,
+            "index_options" : "docs"
+          },
+          "baseurl" : {
+            "type" : "string",
+            "omit_norms" : true,
+            "index_options" : "docs"
+          },
+          "blobhash" : {
+            "type" : "string",
+            "index" : "not_analyzed",
+            "omit_norms" : true,
+            "index_options" : "docs"
+          },
+          "uploadtime" : {
+            "type" : "date",
+            "format" : "dateOptionalTime"
+          },
+          "charmmeta" : {
+            "dynamic" : "false",
+            "properties" : {
+              "name" : {
+                "type" : "multi_field",
+                "fields" : {
+                  "name" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "ngrams" : {
+                    "type" : "string",
+                    "analyser" : "n3_20grams",
+                    "include_in_all" : false
+                  }
+                }
+              },
+              "summary" : {
+                "type" : "string"
+              },
+              "description" : {
+                "type" : "string"
+              },
+              "provides" : {
+                "dynamic" : "false",
+                "properties" : {
+                  "name" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "role" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "interface" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "scope" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  }
+                }
+              },
+              "requires" : {
+                "dynamic" : "false",
+                "properties" : {
+                  "name" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "role" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "interface" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "scope" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  }
+                }
+              },
+              "peers" : {
+                "dynamic" : "false",
+                "properties" : {
+                  "name" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "role" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "interface" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "scope" : {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  }
+                }
+              },
+              "categories" : {
+                "type" : "string",
+                "index" : "not_analyzed",
+                "omit_norms" : true,
+                "index_options" : "docs"
+              }
+            }
+          },
+          "charmconfig" : {
+            "dynamic" : "false",
+            "properties" : {
+              "options" : {
+                "dynamic" : "false",
+                "properties" : {
+                  "description" : {
+                    "type" : "string"
+                  },
+                  "default" : {
+                    "type" : "string"
+                  }
+                }
+              }
+            }
+          },
+          "charmactions" : {
+            "dynamic" : "false",
+            "properties" : {
+              "description" : {
+                "type" : "string"
+              },
+              "action_name" : {
+                "type" : "string",
+                "index" : "not_analyzed",
+                "omit_norms" : true,
+                "index_options" : "docs"
+              }
+            }
+          },
+          "charmprovidedinterfaces" : {
+            "type" : "string",
+            "index" : "not_analyzed",
+            "omit_norms" : true,
+            "index_options" : "docs"
+          },
+          "charmrequiredinterfaces" : {
+            "type" : "string",
+            "index" : "not_analyzed",
+            "omit_norms" : true,
+            "index_options" : "docs"
+          },
+
+
+          "bundledata" : {
+            "type": "object",
+            "dynamic": "false",
+            "properties" : {
+              "services" : {
+                "type": "object",
+                "dynamic": "false",
+                "properties": {
+                  "charm": {
+                    "type" : "string",
+                    "index" : "not_analyzed",
+                    "omit_norms" : true,
+                    "index_options" : "docs"
+                  },
+                  "numunits": {
+                    "type" : "integer",
+                    "index": "not_analyzed"
+                  }
+                }
+              },
+              "series" : {
+                "type" : "string"
+              },
+              "relations" : {
+                    "type" : "string",
+                    "index": "not_analyzed"
+                  }
+                }
+              }
+            }
+          },
+          "bundlereadme" : {
+            "type": "string",
+            "index": "not_analyzed",
+            "omit_norms" : true,
+            "index_options" : "docs"
+          },
+          "bundlecharms": {
+            "type": "string",
+            "index": "not_analyzed",
+            "omit_norms" : true,
+            "index_options" : "docs"
+          },
+          "bundlemachinecount": {
+            "type": "integer"
+          },
+          "bundleunitcount": {
+            "type": "integer"
+          }
+        }
+      }
+}
--- a/internal/elasticsearch/elasticsearch.go
+++ b/internal/elasticsearch/elasticsearch.go
@@ -4,11 +4,13 @@
 package elasticsearch
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 
 	//"github.com/juju/charmstore/internal/mongodoc"
 	"github.com/juju/errgo"
@@ -17,11 +19,10 @@ import (
 type Database struct {
 	server string
 	port   int
-	index  string
 }
 
 // AddNewEntity takes a mongo document and indexes it in ElasticSearch.
-func (db *Database) AddNewEntity(doc interface{}) error {
+func (db *Database) AddNewEntity(index string, doc interface{}) error {
 	// DELETE ME
 	// An example auto-index document creation:
 	//$ curl -XPOST 'http://localhost:9200/twitter/tweet' -d '{
@@ -34,7 +35,7 @@ func (db *Database) AddNewEntity(doc interface{}) error {
 		return errgo.Mask(err)
 	}
 	buf := bytes.NewBuffer(json)
-	response, err := http.Post(db.url("entity"), "application/json", buf)
+	response, err := http.Post(db.url(index, "entity"), "application/json", buf)
 	if err != nil {
 		return errgo.Mask(err)
 	}
@@ -48,8 +49,62 @@ func (db *Database) AddNewEntity(doc interface{}) error {
 }
 
 // url constructs the URL for accessing the database.
-func (db *Database) url(entityName string) string {
+// entityName is the ElasticSearch "type"
+func (db *Database) url(index, entityName string) string {
 
-	return fmt.Sprintf("http://%s:%d/%s/%s", db.server, db.port, db.index, entityName)
+	return fmt.Sprintf("http://%s:%d/%s/%s/", db.server, db.port, index, entityName)
 
+}
+
+// Get makes a GET request to the Database using the path parameter
+// and returns the response
+func (db *Database) Get(path string) (*http.Response, error) {
+	return http.Get(fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path))
+}
+
+// Delete makes a DELETE request to the database using the path parameter
+func (db *Database) Delete(path string) (*http.Response, error) {
+	req, err := http.NewRequest("DELETE", fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path), nil)
+	if err != nil {
+		return nil, err
+	}
+	return http.DefaultClient.Do(req)
+}
+
+// CatIndices does a GET request to the Database and parses the string
+// response returning a slice containing the name of each index
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_list_all_indexes.html
+func (db *Database) CatIndices() ([]string, error) {
+	resp, err := db.Get("_cat/indices")
+	if err != nil {
+		return nil, err
+	}
+	scanner := bufio.NewScanner(resp.Body)
+	var indices []string
+	for scanner.Scan() {
+		line := scanner.Text()
+		fields := strings.Fields(line)
+		if len(fields) > 1 {
+			indices = append(indices, fields[1])
+		}
+	}
+	return indices, nil
+}
+
+// DeleteIndex does a DELETE request to the database for the given index
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/indices-delete-index.html
+// If the index does not exist or if the database cannot be
+// reached, then an error is returned.
+func (db *Database) DeleteIndex(index string) error {
+	resp, err := db.Delete(index)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode == http.StatusOK {
+		return nil
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		return fmt.Errorf("index %s not found", index)
+	}
+	return fmt.Errorf("unexpected http response trying to delete index %s: %d", index, resp.StatusCode)
 }
--- a/internal/elasticsearch/elasticsearch.go
+++ b/internal/elasticsearch/elasticsearch.go
@@ -1,6 +1,13 @@
 // Copyright 2014 Canonical Ltd.
 // Licensed under the LGPLv3, see LICENCE file for details.
 
+// elasticsearch package api attempts to name methods to match the
+// corresponding elasticsearch endpoint. Methods names like CatIndices are
+// named as such because they correspond to /_cat/indices elasticsearch
+// endpoint.
+// There is no reason to use different vocabulary from that of elasticsearch.
+// Use the elasticsearch terminology and avoid mapping names of things.
+
 package elasticsearch
 
 import (
@@ -12,7 +19,6 @@ import (
 	"net/http"
 	"strings"
 
-	//"github.com/juju/charmstore/internal/mongodoc"
 	"github.com/juju/errgo"
 )
 
@@ -21,8 +27,7 @@ type Database struct {
 	port   int
 }
 
-// AddNewEntity takes a mongo document and indexes it in ElasticSearch.
-func (db *Database) AddNewEntity(index string, doc interface{}) error {
+func (db *Database) PostDocument(index, type_ string, doc interface{}) error {
 	// DELETE ME
 	// An example auto-index document creation:
 	//$ curl -XPOST 'http://localhost:9200/twitter/tweet' -d '{
@@ -30,12 +35,12 @@ func (db *Database) AddNewEntity(index string, doc interface{}) error {
 	//"post_date" : "2009-11-15T14:12:12",
 	//"message" : "trying out Elasticsearch"
 	//}'
-	json, err := json.Marshal(doc)
+	data, err := json.Marshal(doc)
 	if err != nil {
 		return errgo.Mask(err)
 	}
-	buf := bytes.NewBuffer(json)
-	response, err := http.Post(db.url(index, "entity"), "application/json", buf)
+	buf := bytes.NewReader(data)
+	response, err := http.Post(db.url(index, type_), "application/json", buf)
 	if err != nil {
 		return errgo.Mask(err)
 	}
@@ -43,28 +48,21 @@ func (db *Database) AddNewEntity(index string, doc interface{}) error {
 	if response.StatusCode != http.StatusCreated {
 		body, _ := ioutil.ReadAll(response.Body)
 		// Error checking within this error handler is not helpful.
-		return fmt.Errorf("ElasticSearch POST response status: %d, body: %s", response.StatusCode, body)
+		return errgo.Newf("ElasticSearch POST response status: %d %s, body: %s", response.StatusCode, response.Status, body)
 	}
 	return nil
 }
 
 // url constructs the URL for accessing the database.
-// entityName is the ElasticSearch "type"
-func (db *Database) url(index, entityName string) string {
-
-	return fmt.Sprintf("http://%s:%d/%s/%s/", db.server, db.port, index, entityName)
+func (db *Database) url(pathParts ...string) string {
+	path := strings.Join(pathParts, "/")
+	return fmt.Sprintf("http://%s:%d/%s/", db.server, db.port, path)
 
 }
 
-// Get makes a GET request to the Database using the path parameter
-// and returns the response
-func (db *Database) Get(path string) (*http.Response, error) {
-	return http.Get(fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path))
-}
-
-// Delete makes a DELETE request to the database using the path parameter
-func (db *Database) Delete(path string) (*http.Response, error) {
-	req, err := http.NewRequest("DELETE", fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path), nil)
+// Delete makes a DELETE request to the database using the path parameter.
+func (db *Database) delete(path string) (*http.Response, error) {
+	req, err := http.NewRequest("DELETE", db.url(path), nil)
 	if err != nil {
 		return nil, err
 	}
@@ -72,13 +70,14 @@ func (db *Database) Delete(path string) (*http.Response, error) {
 }
 
 // CatIndices does a GET request to the Database and parses the string
-// response returning a slice containing the name of each index
+// response returning a slice containing the name of each index.
 // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_list_all_indexes.html
 func (db *Database) CatIndices() ([]string, error) {
-	resp, err := db.Get("_cat/indices")
+	resp, err := http.Get(db.url("_cat/indices"))
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
 	scanner := bufio.NewScanner(resp.Body)
 	var indices []string
 	for scanner.Scan() {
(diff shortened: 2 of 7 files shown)


jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/289/

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package elasticsearch
+
+import (
+	"testing"
+
+	jujutesting "github.com/juju/testing"
+	gc "launchpad.net/gocheck"
```
rogpeppe: This should be gopkg.in/check.v1

internal/elasticsearch/definitions/entity.json
```diff
@@ -0,0 +1,241 @@
+{
```
rogpeppe, line 1: Do you think it would be possible to generate this automatically from the mongodoc.Entity type (with some additional info as to which fields to index) ?

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	//"github.com/juju/charmstore/internal/mongodoc"
```
frankban: Can we remove this commented import?

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+	//$ curl -XPOST 'http://localhost:9200/twitter/tweet' -d '{
+	//"user" : "kimchy",
+	//"post_date" : "2009-11-15T14:12:12",
+	//"message" : "trying out Elasticsearch"
+	//}'
+	json, err := json.Marshal(doc)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	buf := bytes.NewBuffer(json)
```
rogpeppe: For a read-only buffer, which this is, it's more conventional to use bytes.NewReader.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+	if response.StatusCode != http.StatusCreated {
+		body, _ := ioutil.ReadAll(response.Body)
+		// Error checking within this error handler is not helpful.
+		return fmt.Errorf("ElasticSearch POST response status: %d, body: %s", response.StatusCode, body)
+	}
+	return nil
+}
+
+// url constructs the URL for accessing the database.
+// entityName is the ElasticSearch "type"
```
frankban: Please complete the sentence.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+func (db *Database) url(index, entityName string) string {
+
```
frankban: Unnecessary new line.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+// AddNewEntity takes a mongo document and indexes it in ElasticSearch.
+func (db *Database) AddNewEntity(index string, doc interface{}) error {
+	// DELETE ME
+	// An example auto-index document creation:
+	//$ curl -XPOST 'http://localhost:9200/twitter/tweet' -d '{
+	//"user" : "kimchy",
+	//"post_date" : "2009-11-15T14:12:12",
+	//"message" : "trying out Elasticsearch"
+	//}'
+	json, err := json.Marshal(doc)
```
rogpeppe: Perhaps "json" isn't a great name here, as it matches the json package id.
"data" ?

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+	buf := bytes.NewBuffer(json)
+	response, err := http.Post(db.url(index, "entity"), "application/json", buf)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusCreated {
+		body, _ := ioutil.ReadAll(response.Body)
+		// Error checking within this error handler is not helpful.
+		return fmt.Errorf("ElasticSearch POST response status: %d, body: %s", response.StatusCode, body)
```
frankban: errgo.Newf?

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+// url constructs the URL for accessing the database.
+// entityName is the ElasticSearch "type"
+func (db *Database) url(index, entityName string) string {
+
+	return fmt.Sprintf("http://%s:%d/%s/%s/", db.server, db.port, index, entityName)
+
+}
+
+// Get makes a GET request to the Database using the path parameter
+// and returns the response
```
frankban: Please complete the sentence.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+		return fmt.Errorf("ElasticSearch POST response status: %d, body: %s", response.StatusCode, body)
+	}
+	return nil
+}
+
+// url constructs the URL for accessing the database.
+// entityName is the ElasticSearch "type"
+func (db *Database) url(index, entityName string) string {
+
+	return fmt.Sprintf("http://%s:%d/%s/%s/", db.server, db.port, index, entityName)
```
frankban: What about changing this to something like:
func (db *Database) url(path string) string
so that it can be reused also in other parts of the code (e.g. db.Get)?

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+
+}
+
+// Get makes a GET request to the Database using the path parameter
+// and returns the response
+func (db *Database) Get(path string) (*http.Response, error) {
+	return http.Get(fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path))
+}
+
+// Delete makes a DELETE request to the database using the path parameter
```
frankban: Please complete the sentence.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+
+// url constructs the URL for accessing the database.
+// entityName is the ElasticSearch "type"
+func (db *Database) url(index, entityName string) string {
+
+	return fmt.Sprintf("http://%s:%d/%s/%s/", db.server, db.port, index, entityName)
+
+}
+
+// Get makes a GET request to the Database using the path parameter
```
rogpeppe: I'd prefer it if these exported functions were couched in higher level terms. I'm not really sure that we want callers to need to know the details of the elastic search name space - this package should be hiding that, I think.

There are possible reasons that might not be a good idea though, but my inclination would be to unexport this and the Delete method and create entry points that represent the operations that we will actually want to use.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+func (db *Database) Delete(path string) (*http.Response, error) {
+	req, err := http.NewRequest("DELETE", fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path), nil)
+	if err != nil {
+		return nil, err
+	}
+	return http.DefaultClient.Do(req)
+}
+
+// CatIndices does a GET request to the Database and parses the string
+// response returning a slice containing the name of each index
```
frankban: Ditto.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+func (db *Database) DeleteIndex(index string) error {
+	resp, err := db.Delete(index)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode == http.StatusOK {
+		return nil
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		return fmt.Errorf("index %s not found", index)
```
frankban: Specifically return a ErrNotFound?

internal/elasticsearch/package_test.go
```diff
@@ -0,0 +1,254 @@
...
+	"syscall"
+	"testing"
+
+	"github.com/juju/loggo"
+	jujutesting "github.com/juju/testing"
+
+	gc "launchpad.net/gocheck"
+)
+
+type ElasticSearchInstance struct {
```
frankban, line 30: Do we require this to be exported?

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+// Delete makes a DELETE request to the database using the path parameter
+func (db *Database) Delete(path string) (*http.Response, error) {
+	req, err := http.NewRequest("DELETE", fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path), nil)
+	if err != nil {
+		return nil, err
+	}
+	return http.DefaultClient.Do(req)
+}
+
+// CatIndices does a GET request to the Database and parses the string
```
rogpeppe: CatIndices is a slightly odd name here. Is "Cat" short for "Catalog" or is it like the unix command "cat" ?

How about just "AllIndexes" as a name? (indexes is the plural used
by elastic search itself and a quick google seems to indicate that
it's fine)

I'd suggest a doc comment like this:

```
// AllIndexes returns the names of all the indexes in the database.
```

rather than mentioning the implementation details.

internal/elasticsearch/package_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+var (
+	ElasticSearchServer = &ElasticSearchInstance{}
+	logger              = loggo.GetLogger("juju.testing.elasticsearchsuite")
+)
+
+const (
+	maxStartAttempts = 5
+
+	//elasticsearch exits with code 143 on SIGTERM
```
frankban: // ElasticSearch exits with code 143 on SIGTERM.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+// response returning a slice containing the name of each index
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_list_all_indexes.html
```
rogpeppe, line 74: Thanks for the link - it's great to have this.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+func (db *Database) url(index, entityName string) string {
+
+	return fmt.Sprintf("http://%s:%d/%s/%s/", db.server, db.port, index, entityName)
+
+}
+
+// Get makes a GET request to the Database using the path parameter
+// and returns the response
+func (db *Database) Get(path string) (*http.Response, error) {
+	return http.Get(fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path))
```
rogpeppe: We perhaps want to check that the status code is OK and return an error if not.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+
+// CatIndices does a GET request to the Database and parses the string
+// response returning a slice containing the name of each index
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_list_all_indexes.html
+func (db *Database) CatIndices() ([]string, error) {
+	resp, err := db.Get("_cat/indices")
+	if err != nil {
+		return nil, err
+	}
+	scanner := bufio.NewScanner(resp.Body)
```
rogpeppe, line 81: How about this instead?

```
data, err := ioutil.ReadAll(resp.Body)
if err != nil {
    return nil, errgo.Notef(err, "error reading body")
}
return strings.Fields(string(data))
```

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+		line := scanner.Text()
+		fields := strings.Fields(line)
+		if len(fields) > 1 {
+			indices = append(indices, fields[1])
+		}
+	}
+	return indices, nil
+}
+
+// DeleteIndex does a DELETE request to the database for the given index
```
rogpeppe: // DeleteIndex deletes the index with the given name from the database.

?

internal/elasticsearch/elasticsearch.go
rogpeppe, line 81: Also, you'll need:

```
defer resp.Body.Close()
```

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,73 @@
...
+	jujutesting.IsolationSuite
+	ElasticSearchSuite
+}
+
+func (s *IsolatedElasticSearchSuite) SetUpSuite(c *gc.C) {
+	s.IsolationSuite.SetUpSuite(c)
+	s.ElasticSearchSuite.SetUpSuite(c)
+}
+func (s *IsolatedElasticSearchSuite) TearDownSuite(c *gc.C) {
+	s.IsolationSuite.TearDownSuite(c)
```
frankban: I guess it's a bit safer to revert the order of the suites in TearDown.

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,73 @@
...
+	s.ElasticSearchSuite.TearDownSuite(c)
+}
+func (s *IsolatedElasticSearchSuite) SetUpTest(c *gc.C) {
+	s.IsolationSuite.SetUpTest(c)
+	s.ElasticSearchSuite.SetUpTest(c)
+}
+func (s *IsolatedElasticSearchSuite) TearDownTest(c *gc.C) {
+	s.IsolationSuite.TearDownTest(c)
+	s.ElasticSearchSuite.TearDownTest(c)
```
frankban, line 35: Ditto.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+	return http.Get(fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path))
+}
+
+// Delete makes a DELETE request to the database using the path parameter
+func (db *Database) Delete(path string) (*http.Response, error) {
+	req, err := http.NewRequest("DELETE", fmt.Sprintf("http://%s:%d/%s", db.server, db.port, path), nil)
+	if err != nil {
+		return nil, err
+	}
+	return http.DefaultClient.Do(req)
```
rogpeppe, line 69: I'm not sure it's worth returning the entire response here.
All we're interested in is the response status code, which
we could check here. I suggest returning an error only,
so we don't burden all the callers with checking the http
status and closing the body.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+
+// DeleteIndex does a DELETE request to the database for the given index
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/indices-delete-index.html
+// If the index does not exist or if the database cannot be
+// reached, then an error is returned.
+func (db *Database) DeleteIndex(index string) error {
+	resp, err := db.Delete(index)
+	if err != nil {
+		return err
+	}
```
rogpeppe, line 101: resp.Body needs to be closed here.

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,73 @@
...
+		"a": "b",
+	}
+	s.db.AddNewEntity("testindex", doc)
+	err := s.db.DeleteIndex("testindex")
+	c.Assert(err, gc.IsNil)
+}
+
+func (s *IsolatedElasticSearchSuite) TestDeleteErrorOnNonExistingIndex(c *gc.C) {
+	err := s.db.DeleteIndex("nope")
+	c.Assert(err, gc.NotNil)
```
frankban: It would be nice to check the specific error message.

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (s *IsolatedElasticSearchSuite) TestIndexesEmpty(c *gc.C) {
+	indexes, _ := s.db.CatIndices()
```
frankban: Also check the error?

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,110 @@
...
+	if resp.StatusCode == http.StatusOK {
+		return nil
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		return fmt.Errorf("index %s not found", index)
+	}
+	return fmt.Errorf("unexpected http response trying to delete index %s: %d", index, resp.StatusCode)
```
rogpeppe: Consider including the body of the error here rather than just the status code,
depending on what it looks like. At the least we could include resp.Status, which
is a bit more friendly than a decimal number.

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,73 @@
...
+	c.Assert(indexes, gc.HasLen, 0)
+}
+
+func (s *IsolatedElasticSearchSuite) TestIndexesCreatedAutomatically(c *gc.C) {
+	doc := map[string]string{"a": "b"}
+	s.db.AddNewEntity("testindex", doc)
+	indexes, _ := s.db.CatIndices()
```
frankban: Here too.

internal/elasticsearch/elasticsearch_test.go
rogpeppe: Please include the error message here (and in other places
that check for NotNil errors)

e.g.
     c.Assert(err, gc.ErrorMatches, "index does not exist")

so that we can verify that the error messages look reasonable.

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,73 @@
...
+	c.Assert(indexes[0], gc.Equals, "testindex")
```
rogpeppe, line 75:
```
c.Assert(indexes, gc.DeepEquals, []string{"testindex"})
```

internal/elasticsearch/elasticsearch_test.go
rogpeppe: +1

internal/elasticsearch/package_test.go
```diff
@@ -0,0 +1,254 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+// This file is a simplyfied copy of github.com/juju/testing/mgo.go
```
rogpeppe: s/simplyfied/simplified/

frankban:
This is a very good start for the ES stuff we need.
I made some suggestions above, thank you!

internal/elasticsearch/package_test.go
```diff
@@ -0,0 +1,254 @@
...
+func (es *ElasticSearchInstance) writeConfig() (string, error) {
+	if es.Dir == "" {
+		return "", fmt.Errorf("directory not set")
+	}
+	file, err := os.Create(es.Dir + "/elasticsearch.yml")
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+	fmt.Fprintf(file, "cluster.name: elasticsearch_testing\n")
```
rogpeppe: Perhaps use a template here? (see the text/template package)

e.g.

at global scope:

```
var config = template.Must(template.New("").Parse(`
path.data: {{.Dir}}/data
path.logs: {{.Dir}}/log/
network.host: 127.0.0.1
http.port: {{.HttpPort}}
`))
```

then in the function:

```
 config.Execute(file, es)
```

internal/elasticsearch/package_test.go
```diff
@@ -0,0 +1,254 @@
...
+	fmt.Fprintf(file, "path.data: %s/data\n", es.Dir)
+	fmt.Fprintf(file, "path.logs: %s/log/\n", es.Dir)
+	fmt.Fprintf(file, "network.host: 127.0.0.1\n")
+	fmt.Fprintf(file, "http.port: %d\n", es.HttpPort)
+	return file.Name(), nil
+}
+
+func (es *ElasticSearchInstance) Destroy() {
+	if es.Server != nil {
```
rogpeppe, line 168: Invert the condition here to save indentation?

e.g.
      if es.Server == nil {
          logger.Errorf(....)
          return
     }
     etc

internal/elasticsearch/package_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+	"github.com/juju/loggo"
+	jujutesting "github.com/juju/testing"
+
+	gc "launchpad.net/gocheck"
+)
+
+type ElasticSearchInstance struct {
+	Dir      string
+	HttpPort int
```
rogpeppe: s/HttpPort/HTTPPort/

(acronyms are either all capitalised or all lower case)

internal/elasticsearch/definitions/entity.json
urosj, line 1: Was already thinking about that. It sure can be done and would help, but we need some search functional before that. It's an "improvement" task, an important one as we'll have 1-on-1 mapping with what actual Entity in code is.

internal/elasticsearch/package_test.go
```diff
@@ -0,0 +1,254 @@
...
+)
+
+const (
+	maxStartAttempts = 5
+
+	//elasticsearch exits with code 143 on SIGTERM
+	elasticSearchSigTermErrCode = 143
+)
+
+func ElasticSearchTestPackage(t *testing.T) {
```
rogpeppe, line 55: I understand that this code is using precedent set by mgo.MgoTestPackage,
but I'm not sure it will work well for us. Consider that we will almost certainly
want to have some tests in internal/v4 that test how the API works along
with elastic search, which means we will need both mongo and elastic search
to run at the same time. With this approach, that's not possible, because
we can't have tests running both within MgoTestPackage and ElasticSearchTestPackage
at the same time.

Perhaps I could suggest another approach, one that we were considering for juju-core too: require that the user start an elastic search instance before running the tests. Then all packages can use the same elastic search engine, and we won't be hit by 6s startup times (and potentially multiple concurrent running JVMs).

To do this, we'd need two things (at least):
- an environment variable that points to the elastic search engine to use (perhaps ELASTICSEARCH=host:port; or ELASTICSEARCH=none to disable the elastic search tests)
  - the ability to tell the code to use a given prefix on all indexes and schemas

Then we'd have an ElasticSearchSuite that generated a random prefix, probably including the date too, and used that for the index/schema prefix. It would delete all its indexes and schemas at the end, and possibly garbage-collect old ones so we don't accumulate endless garbage indexes from interrupted tests.

This means that most of the code in this file would be unnecessary. What do you think?

rogpeppe:
This is a nice start, thanks.
There are some issues that need to be resolved, I think.

internal/elasticsearch/package_test.go
jrwren, line 55: My original goal was to write a suite like MgoSuite so that we could write tests in a similar way that we do with the tests which use mongodb.

It is possible to have tests running both within MgoTestPackage and ElasticSearchTestPackage by creating a new package which uses both and a new Suite which combines the two Suite types just like IsolatedMgoSuite combines IsolationSuite and MgoSuite. Just because you can do something doesn't mean you should do something. I think we agree. ElasticSearchIsolationMgoSuite does not sound like a good idea.

internal/elasticsearch/elasticsearch.go
jrwren: I was intentionally NOT going this far. I agree that this would be the ideal. At this point, I'm not trying to make a full elasticsearch client package. That might be all the more reason to not export identifiers which we aren't going to be used by charmstore. Thanks.

internal/elasticsearch/elasticsearch.go
jrwren: mgo has an ErrNotFound. It seems wrong to use it here.
params has an ErrNotFound. It is a string constant used for json response.

I'll create an ErrNotFound type here. DONE.

internal/elasticsearch/package_test.go
jrwren, line 30: Unexporting.
I was following juju/testing/mgo.go.

internal/elasticsearch/elasticsearch.go
jrwren: I am attempting to name methods to match their elasticsearch endpoint.
This is why there is a link to http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_list_all_indexes.html

There is no reason to use different vocabulary from that of elasticsearch. I much prefer using their terminology and avoid mapping names of things.

internal/elasticsearch/package_test.go
jrwren: Done.

internal/elasticsearch/elasticsearch_test.go
jrwren: Done.

internal/elasticsearch/elasticsearch.go
jrwren: Done.

internal/elasticsearch/elasticsearch.go
jrwren: Done. Great to know. Thank You.

internal/elasticsearch/elasticsearch.go
jrwren: Done.

internal/elasticsearch/elasticsearch.go
jrwren: Done.

internal/elasticsearch/elasticsearch.go
jrwren: Done.

internal/elasticsearch/elasticsearch.go
jrwren: Done. Thank You.

internal/elasticsearch/elasticsearch.go
jrwren: Done.

internal/elasticsearch/elasticsearch.go
jrwren: Done. This is nicer. Thanks.

internal/elasticsearch/elasticsearch.go
jrwren: Since url is nicer now, I'm removing this get.

internal/elasticsearch/elasticsearch.go
jrwren, line 81: Avoiding ReadAll. Really do need to cat like awk '{print $2}' here. The response is whitespace separated fields in newline separated plain text.

internal/elasticsearch/elasticsearch.go
jrwren: Done.

internal/elasticsearch/elasticsearch_test.go
jrwren: Done.

internal/elasticsearch/elasticsearch_test.go
jrwren, line 35: Done.

internal/elasticsearch/elasticsearch.go
jrwren, line 101: Done. Thank You.

internal/elasticsearch/elasticsearch.go
jrwren: Done. Thank You.

internal/elasticsearch/package_test.go
jrwren: Done.

internal/elasticsearch/package_test.go
jrwren: Done.

internal/elasticsearch/package_test.go
jrwren, line 55: I implemented something basic.

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/292/

internal/elasticsearch/elasticsearch.go
rogpeppe: The elastic search uses "list all indexes" as the title of the page. This seems like a nice name to me. Let's use methods names that make sense even if you're not familiar with the elastic search API please. It's idiomatic to make the package API look nice and easy to use even if we're encapsulating something that isn't so well named.

internal/elasticsearch/elasticsearch.go
rogpeppe: We don't want to make a full elastic search client package, just one that provides the functionality that we directly require. Providing a generalised Get request moves us more in the direction of providing a full elastic search client package, in fact, and means that more of the code will be directly dependent on the details of elastic search's API.

internal/elasticsearch/elasticsearch.go
rogpeppe, line 81: There is no point in avoiding ReadAll when you're going to return the entire contents as a slice of strings anyway. Avoiding ReadAll is worth is when you can process data without using O(n) memory. Even more so as it looks like ES returns everything on a single line anyway.

(I still suggest using ReadAll in this case, but for future reference,
bufio.Scanner already has functionality for splitting on words:

```
scanner := bufio.NewScanner(resp.Body)
var indices []string
scanner.Split(bufio.ScanWords)
for scanner.Scan() {
    indices = append(indices, scanner.Text())
}
```

)

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/293/

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/294/

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/juju/errgo"
+)
+
+type Database struct {
+	server string
+	port   int
+}
+
+func (db *Database) PostDocument(index, type_ string, doc interface{}) error {
+	// DELETE ME
```
mhilton: To be deleted later?

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,121 @@
...
+		return errgo.Mask(err)
+	}
+	buf := bytes.NewReader(data)
+	response, err := http.Post(db.url(index, type_), "application/json", buf)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusCreated {
+		body, _ := ioutil.ReadAll(response.Body)
```
mhilton, line 44: You should probably be reading the body under all circumstances, even if you don't care what it is. Go won't reuse the connection if the body hasn't been read, which tends to exhaust ports, and is inefficient.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,121 @@
...
+		// Error checking within this error handler is not helpful.
+		return errgo.Newf("ElasticSearch POST response status: %d %s, body: %s", response.StatusCode, response.Status, body)
```
mhilton, line 46: I believe http.Response.Status includes the StatusCode, It seems a little redundant to have both.

internal/elasticsearch/elasticsearch.go
jrwren, line 44: Even if it is closed but not read?

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return nil
+}
+
+// url constructs the URL for accessing the database.
+func (db *Database) url(pathParts ...string) string {
+	path := strings.Join(pathParts, "/")
```
mhilton: path.Join maybe (golang.org/pkg/path)

internal/elasticsearch/elasticsearch.go
mhilton, line 44: Yeah, the logic behind it is a bit subtle. Basically it will only reuse a connection if the calling application has read all of the body. If it's been closed before being finished it will be for two reasons either the application isn't interested, or it;s decided the data is too big and doesn't want to download it. If it's  the latter then the standard library has no business downloading it all just so it can reuse the connection. I gained intimate knowledge of the workings of this in my last job when solr was happily sending me a completely unneeded \n at then end of some json that the unmarshaller didn't get to and so ended up spinning through connections because of a single unread byte.

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,121 @@
...
+	return fmt.Sprintf("http://%s:%d/%s/", db.server, db.port, path)
```
mhilton: perhaps you should be using a url.URL for this

internal/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, err
+	}
+	return http.DefaultClient.Do(req)
+}
+
+// ListAllIndexes does a GET request to the Database and parses the string
+// response returning a slice containing the name of each index.
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_list_all_indexes.html
+func (db *Database) ListAllIndexes() ([]string, error) {
+	resp, err := http.Get(db.url("_cat/indices"))
```
mhilton, line 76: We should probably be trying to use one of the JSON APIs here rather than parsing text.

jrwren, line 76: Please suggest a URL which returns this data. I cannot find one.

mhilton, line 76: This stack overflow gives a few choices. http://stackoverflow.com/questions/17426521/list-all-indexes-on-elasticsearch-server

I hadn't researched it before hand, it just seemed to me that it must be possible.

jrwren, line 76: thanks.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/296/

jrwren:
:shipit:

jujugui:

PR #121, opened 2014-10-01
lppublish: migrate bundles from legacy format

rogpeppe:
It doesn't provide the bundle's charms for the migration process yet,
as we need charm archive PUT for the charm revisions to work
correctly with respect to the charm URLs mentioned in the bundles.

Manually verified that at least some bundles get correctly imported.

### DIFF
--- a/lppublish/lpad.go
+++ b/lppublish/lpad.go
@@ -4,7 +4,6 @@
 package lppublish
 
 import (
-	"bufio"
 	"bytes"
 	"crypto/sha512"
 	"encoding/base64"
@@ -22,7 +21,10 @@ import (
 
 	"github.com/juju/errgo"
 	"github.com/juju/loggo"
+	"github.com/juju/utils/fs"
 	"gopkg.in/juju/charm.v4"
+	"gopkg.in/juju/charm.v4/migratebundle"
+	"gopkg.in/yaml.v1"
 	"launchpad.net/lpad"
 
 	"github.com/juju/charmstore/params"
@@ -34,6 +36,8 @@ const BzrDigestKey = "bzr-digest"
 
 var logger = loggo.GetLogger("charmload")
 var failsLogger = loggo.GetLogger("charmload_v4.loadfails")
+
+const officialBundlePromulgator = "charmers"
 
 type PublishBranchError struct {
 	URL string
@@ -197,12 +201,7 @@ func (cl *charmLoader) publisher(results <-chan entityResult, errs chan<- error)
 	for result := range results {
 		logger.Debugf("start publishing URLs for %s", result.charmURL)
 		urls := []*charm.Reference{result.charmURL}
-		urls = appendPromulgatedCharmURLs(
-			result.tip.OfficialSeries,
-			result.charmURL.Schema,
-			result.charmURL.Name,
-			urls,
-		)
+		urls = append(urls, promulgatedURLs(result.charmURL, result.tip.OfficialSeries)...)
 		err := cl.publishBazaarBranch(urls, result.branchURL, result.tip.Revision)
 		if err != nil {
 			failsLogger.Errorf("cannot publish branch %v to charm store: %v", result.branchURL, err)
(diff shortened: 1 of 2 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/299/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/300/

lppublish/lpad.go
```diff
@@ -329,25 +311,104 @@ func (cl *charmLoader) publishBazaarBranch(urls []*charm.Reference, branchURL st
...
+		if err := cl.publish(urls, ch, tempDir, tipDigest); err != nil {
+			return errgo.Notef(err, "cannot publish")
+		}
+		return nil
+	}
+	bundles, err := readBundles(branchDir, tempDir)
+	if err != nil {
+		return errgo.Notef(err, "cannot read bundle %q", urls[0])
+	}
+	for name, bundle := range bundles {
```
mhilton, line 335: I think you could really do with a comment here explaining what this is doing and why.

mhilton:
:+1: The code looks good, but I struggled to understand bits of it.

lppublish/lpad.go
```diff
@@ -329,25 +311,104 @@ func (cl *charmLoader) publishBazaarBranch(urls []*charm.Reference, branchURL st
...
+		var finalURLs []*charm.Reference
+		if len(bundles) == 1 {
+			finalURLs = urls
+		} else {
+			finalURLs = make([]*charm.Reference, len(urls))
+			for i, url := range urls {
+				finalURLs[i] = new(charm.Reference)
+				*finalURLs[i] = *url
```
urosj, line 343: Interesting. Gotta remember this and not doing copy() calls.

lppublish/lpad.go
```diff
@@ -329,25 +311,104 @@ func (cl *charmLoader) publishBazaarBranch(urls []*charm.Reference, branchURL st
...
+				finalURLs[i].Name += "-" + name
+			}
 		}
+		if err := cl.publish(finalURLs, bundle, tempDir, tipDigest); err != nil {
+			return errgo.Notef(err, "cannot publish %q", name)
```
urosj, line 348: What if we try to publish as much bundles as possible and not just return on the first error? On the one hand, we get all the errors on one ingestion, on the other hand, it does nothing to actually solve the error of the bundle. Not a blocker or something we need right away, just a thought.

lppublish/lpad.go
```diff
@@ -329,25 +311,104 @@ func (cl *charmLoader) publishBazaarBranch(urls []*charm.Reference, branchURL st
...
+	}
+	f, err := os.Open(filepath.Join(dir, "bundles.yaml"))
+	if err != nil {
+		return nil, errgo.Mask(err)
+	}
+	data, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, errgo.Notef(err, "cannot read bundles.yaml")
+	}
+	bds, err := migratebundle.Migrate(data, func(id *charm.Reference) (*charm.Meta, error) {
```
urosj, line 376: Could you make a quick explanation why Migrate call is designed as it is, please? (esp. regarding the getCharm part of the Migrate parameters)

lppublish/lpad.go
```diff
@@ -329,25 +311,104 @@ func (cl *charmLoader) publishBazaarBranch(urls []*charm.Reference, branchURL st
...
+			return nil, errgo.Notef(err, "cannot marshal bundle %q", name)
+		}
+		if err := ioutil.WriteFile(filepath.Join(destDir, "bundle.yaml"), data, 0666); err != nil {
+			return nil, errgo.Mask(err)
 		}
+		bundle, err := charm.ReadBundleDir(destDir)
+		if err != nil {
+			return nil, errgo.Notef(err, "cannot read bundle %q", name)
+		}
+		bundles[name] = bundle
```
urosj, line 408: Would it make sense to prepend the README with a note about auto-generated bundle and all the bundles that were generated with this one?

urosj:
:+1: Awesome. Have some questions out of curiosity and some suggestions for the future.

lppublish/lpad.go
rogpeppe, line 376: The getCharm part is needed because the new bundle spec requires relation endpoints to be specified completely, including the relation name. Previously they were optional. The only way of finding out the intended relation names is by fetching the charms themselves and resolving the implied relation names. The fetcher will be implemented after the PUT request has been implemented, so that there's some hope of charm URLs resolving correctly.

lppublish/lpad.go
rogpeppe, line 408: Quite possibly. It's never going to look great though. Will reserve for a subsequent PR.

lppublish/lpad.go
rogpeppe, line 343: copy works on slices, but this isn't doing something on a slice.

I swithered a bit between this and:

```
finalURLs[i] = &charm.Reference{
   Name: url.Name + "-" + name,
   User: url.User,
   Revision: url.Revision,
   Series: url.Series,
```

   }

I'm still not entirely sure I chose the right way.

urosj, line 343: I'd probably write the latter, but OTH I like what you did as it is "independent" from actual members of the struct.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/302/

rogpeppe:
:shipit:

jujugui:

PR #134, opened 2014-10-08
When adding charms or bundles to mongo, index in ES if configured.

bac:
Also add EnsureID to ElasticSearch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ server: install
 
 # Update the project Go dependencies to the required revision.
 deps: $(GOPATH)/bin/godeps
-	godeps -u dependencies.tsv
+	$(GOPATH)/bin/godeps -u dependencies.tsv
 
 # Generate the dependencies file.
 create-deps: $(GOPATH)/bin/godeps
@@ -98,9 +98,10 @@ endif
 	sudo apt-get update
 	@sudo apt-get --yes install $(strip $(DEPENDENCIES)) \
 	$(shell apt-cache madison juju-mongodb mongodb-server | head -1 | cut -d '|' -f1)
-endif
+else
 	@echo sysdeps runs only on systems with apt-get
-	@echo on MacOSX with homebrew try: brew install bazaar mongodb elasticsearch
+	@echo on OS X with homebrew try: brew install bazaar mongodb elasticsearch
+endif
 
 help:
 	@echo -e 'Charmstore - list of make targets:\n'
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -128,7 +128,7 @@ func (s *Store) uploadCharmOrBundle(c interface{}) (blobName, blobHash string, s
 // associated with the given URL.
 func (s *Store) AddCharm(url *charm.Reference, c charm.Charm, blobName, blobHash string, blobSize int64) error {
 	// Add charm metadata to the entities collection.
-	err := s.DB.Entities().Insert(&mongodoc.Entity{
+	entity := &mongodoc.Entity{
 		URL:                     url,
 		BaseURL:                 baseURL(url),
 		BlobHash:                blobHash,
@@ -140,11 +140,20 @@ func (s *Store) AddCharm(url *charm.Reference, c charm.Charm, blobName, blobHash
 		CharmActions:            c.Actions(),
 		CharmProvidedInterfaces: interfacesForRelations(c.Meta().Provides),
 		CharmRequiredInterfaces: interfacesForRelations(c.Meta().Requires),
-	})
+	}
+	err := s.DB.Entities().Insert(entity)
 	if mgo.IsDup(err) {
 		return params.ErrDuplicateUpload
 	}
-	return errgo.Mask(err)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+
+	// Add to ElasticSearch.
+	if err := s.ES.Put(entity); err != nil {
+		return errgo.Notef(err, "cannot index %s to ElasticSearch", entity.URL)
+	}
+	return nil
 }
 
 // FindEntities finds all entities in the store matching the given URL.
@@ -239,7 +248,7 @@ func (s *Store) AddBundle(url *charm.Reference, b charm.Bundle, blobName, blobHa
 	if err != nil {
 		return errgo.Mask(err)
 	}
-	err = s.DB.Entities().Insert(&mongodoc.Entity{
+	entity := &mongodoc.Entity{
 		URL:                url,
 		BaseURL:            baseURL(url),
 		BlobHash:           blobHash,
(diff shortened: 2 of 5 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/352/

internal/charmstore/store.go
```diff
@@ -140,10 +140,21 @@ func (s *Store) AddCharm(url *charm.Reference, c charm.Charm, blobName, blobHash
...
 	if mgo.IsDup(err) {
 		return params.ErrDuplicateUpload
 	}
+	if err != nil {
+		return errgo.Mask(err)
+	}
+
+	// Add to ElasticSearch.
+	if err := s.ES.Put(&entity); err != nil {
+		// What to do if the charm is successfully added to mongo but fails
```
frankban: Perhaps we should log the fact? I am also not sure about whether or not to return an error to the users.

internal/charmstore/store.go
```diff
@@ -140,10 +140,21 @@ func (s *Store) AddCharm(url *charm.Reference, c charm.Charm, blobName, blobHash
...
+		// indexing into ElasticSearch?
+		return errgo.Notef(err, "cannot index %s to ElasticSearch", entity.URL)
+	}
 	return errgo.Mask(err)
```
frankban: It seems we can just return nil here, since possible errors have been already checked.

internal/charmstore/store.go
```diff
@@ -251,10 +262,20 @@ func (s *Store) AddBundle(url *charm.Reference, b charm.Bundle, blobName, blobHa
...
+	err = s.DB.Entities().Insert(&entity)
 	if mgo.IsDup(err) {
 		return params.ErrDuplicateUpload
 	}
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	// Add to ElasticSearch.
+	if err := s.ES.Put(entity); err != nil {
+		// What to do if the charm is successfully added to mongo but fails
```
frankban: // What to do if the bundle...

internal/charmstore/store.go
```diff
@@ -251,10 +262,20 @@ func (s *Store) AddBundle(url *charm.Reference, b charm.Bundle, blobName, blobHa
...
+		// indexing into ElasticSearch?
+		return errgo.Notef(err, "cannot index %s to ElasticSearch", entity.URL)
+	}
 	return errgo.Mask(err)
```
frankban: Same as above.

internal/charmstore/store.go
```diff
@@ -251,10 +262,20 @@ func (s *Store) AddBundle(url *charm.Reference, b charm.Bundle, blobName, blobHa
 		BundleMachineCount: newInt(bundleMachineCount(bundleData)),
 		BundleReadMe:       b.ReadMe(),
 		BundleCharms:       urls,
-	})
+	}
+	err = s.DB.Entities().Insert(&entity)
```
frankban: Isn't entity already a pointer?
So we are passing the address of the pointer type?

internal/charmstore/store.go
```diff
@@ -128,7 +128,7 @@ func (s *Store) uploadCharmOrBundle(c interface{}) (blobName, blobHash string, s
 // associated with the given URL.
 func (s *Store) AddCharm(url *charm.Reference, c charm.Charm, blobName, blobHash string, blobSize int64) error {
 	// Add charm metadata to the entities collection.
-	err := s.DB.Entities().Insert(&mongodoc.Entity{
+	var entity = mongodoc.Entity{
```
frankban: Since we define a pointer in AddBundle, we could add the same here too (entity = &mongodoc.Entity). [shrug]

internal/charmstore/store.go
```diff
@@ -415,10 +436,15 @@ type StoreElasticSearch struct {
...
 func (ses *StoreElasticSearch) Put(entity *mongodoc.Entity) error {
-	if ses == nil {
+	if ses == nil || ses.Database == nil {
 		return nil
 	}
-	return ses.PutDocument(ses.Index, "entity", url.QueryEscape(entity.URL.String()), entity)
+	return ses.PutDocument(ses.Index, "entity", ses.GetID(entity), entity)
+}
+
+// GetID returns the id to be used for indexing the entity into ElasticSearch.
```
frankban: Nice reorganization, thank you.
Is it required for GetID to be exported?
Can it be just a function (like func escapeEntityId(entity *mongodoc.Entity) string)?

internal/charmstore/store.go
bac: Done

internal/elasticsearch/elasticsearch.go
```diff
@@ -28,6 +28,22 @@ type Database struct {
...
+// in ElasticSearch.
+func (db *Database) EnsureID(index, type_, id string) (bool, error) {
+	// TODO (bac) We should limit the fields to the id to avoid retrieving
+	// data we don't use.
+	response, err := http.Get(db.url(index, type_, id, "_source"))
+	if err != nil {
+		return false, errgo.Mask(err)
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
```
frankban: just `return response.StatusCode == http.StatusOK`, nil?

frankban:
:+1: thank you!

internal/charmstore/store.go
bac: We could log the error but what do we return? Can't really return success if ES failed.
Let's continue returning the error but know this may be a place for improvement.

internal/charmstore/store.go
bac: done

internal/charmstore/store.go
bac: Good catch.  Odd that it worked.

internal/charmstore/store.go
bac: done

internal/charmstore/store.go
bac: Removed the comment

internal/charmstore/store.go
bac: It didn't need to be exported.  Changed to getID.

internal/elasticsearch/elasticsearch.go
bac: Yep, done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/353/

Makefile
```diff
@@ -99,8 +99,6 @@ endif
 	@sudo apt-get --yes install $(strip $(DEPENDENCIES)) \
 	$(shell apt-cache madison juju-mongodb mongodb-server | head -1 | cut -d '|' -f1)
 endif
-	@echo sysdeps runs only on systems with apt-get
-	@echo on MacOSX with homebrew try: brew install bazaar mongodb elasticsearch
```
mhilton, line 103: Is this message no longer true, or no longer useful

mhilton:
:+1: LGTM

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/354/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/355/

bac:
:shipit:

jujugui:

PR #142, opened 2014-10-14
Added search filter functions.

mhilton:
Added support in the elasticsearch DSL for the
filter functions defined in http://tinyurl.com/qzobc69

### DIFF
--- a/internal/charmstore/search.go
+++ b/internal/charmstore/search.go
@@ -1,0 +1,285 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package charmstore
+
+import (
+	"encoding/json"
+	"net/url"
+
+	"github.com/juju/errgo"
+
+	"github.com/juju/charmstore/internal/elasticsearch"
+	"github.com/juju/charmstore/internal/mongodoc"
+)
+
+// StoreElasticSearch provides strongly typed methods for accessing the
+// elasticsearch database. These methods will not return errors if
+// elasticsearch is not configured, allowing them to be safely called even if
+// it is not enabled in this service.
+type StoreElasticSearch struct {
+	*elasticsearch.Database
+	Index string
+}
+
+const typeName = "entity"
+
+// Put inserts the mongodoc.Entity into elasticsearch if elasticsearch
+// is configured.
+func (ses *StoreElasticSearch) put(entity *mongodoc.Entity) error {
+	if ses == nil || ses.Database == nil {
+		return nil
+	}
+	return ses.PutDocument(ses.Index, typeName, ses.getID(entity), entity)
+}
+
+// getID returns the id to be used for indexing the entity into ElasticSearch.
+func (ses *StoreElasticSearch) getID(entity *mongodoc.Entity) string {
+	return url.QueryEscape(entity.URL.String())
+}
+
+// Search searches for entities. The query is a json string which conforms
+// to the elasticsearch querydsl.
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl.html
+// It returns a slice containing each Id of the matching documents returned by
+// elasticsearch.
+func (ses *StoreElasticSearch) search(query string) ([]string, error) {
+	if ses == nil || ses.Database == nil {
+		return nil, nil
+	}
+	results, err := ses.Database.Search(ses.Index, typeName, query)
+	if err != nil {
+		return nil, err
+	}
+	ids := make([]string, 0, len(results.Hits.Hits))
+	for _, hit := range results.Hits.Hits {
+		ids = append(ids, hit.ID)
+	}
+	return ids, nil
+	// TODO: return more than just ids e.g. total, hits, score, information, serach time
+}
+
+// ExportToElasticSearch reads all of the mongodoc Entities and writes
+// them to elasticsearch
+func (store *Store) ExportToElasticSearch() error {
+	var result mongodoc.Entity
+	iter := store.DB.Entities().Find(nil).Iter()
+	defer iter.Close()
+	for iter.Next(&result) {
+		if err := store.ES.put(&result); err != nil {
+			return errgo.Notef(err, "cannot index %s", result.URL)
+		}
+	}
+	if err := iter.Close(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// SearchParams represents the search parameters used by the store.
+type SearchParams struct {
+	// The text to use in the full text search query.
+	Text string
+	// If autocomplete is specified, the search will return only charms and
+	// bundles with a name that has text as a prefix.
+	AutoComplete bool
+	// Limit the search to items with attributes that match the specified filter value.
+	Filters map[string][]string
+	// Limit the number of returned items to the specified count.
+	Limit int
+}
+
+// Search the store for the given SearchParams.
+// Returns a slice containing each Id of the matching documents returned by
+// elasticsearch.
+func (store *Store) Search(sp SearchParams) ([]string, error) {
+	query := createSearchDSL(sp)
+	return store.ES.search(query)
+}
+
+// queryFields provides a map of fields to weighting to use with the
+// elasticsearch query.
+func queryFields(sp SearchParams) map[string]float64 {
+	fields := map[string]float64{
+		"CharmMeta.Description": 3,
+	}
+	if sp.AutoComplete {
+		fields["CharmMeta.Name.ngrams"] = 10
+	} else {
+		fields["CharmMeta.Name"] = 10
+	}
+	return fields
+}
+
+// encodeFields takes a map of field name to weight and builds a slice of strings
+// representing those weighted fields for a MultiMatchQuery.
+func encodeFields(fields map[string]float64) []string {
+	fs := make([]string, 0, len(fields))
+	for k, v := range fields {
+		fs = append(fs, elasticsearch.BoostField(k, v))
+	}
+	return fs
+}
+
+// createSearchDSL builds an elasticsearch query from the query parameters.
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl.html
+func createSearchDSL(sp SearchParams) string {
+	qdsl := elasticsearch.QueryDSL{
+		Size: sp.Limit,
+	}
+
+	// Full text search
+	var q elasticsearch.Query
+	if sp.Text == "" {
+		q = elasticsearch.MatchAllQuery{}
+	} else {
+		q = elasticsearch.MultiMatchQuery{
+			Query:  sp.Text,
+			Fields: encodeFields(queryFields(sp)),
+		}
+	}
+
+	// Attenuation function
+	q = elasticsearch.FunctionScoreQuery{
+		Query: q,
+		Functions: []elasticsearch.Function{
+			{
+				Function: "linear",
+				Field:    "UploadTime",
+				Scale:    "365d",
+			},
+		},
+	}
+
+	// Filters
+	qdsl.Query = elasticsearch.FilteredQuery{
+		Query:  q,
+		Filter: createFilters(sp.Filters),
+	}
+
+	bytes, err := json.Marshal(qdsl)
+	if err != nil {
+		panic(err)
+	}
+	return string(bytes)
+}
+
+// createFilters converts the filters requested with the serch API into
+// filters in the elasticsearch query DSL. Please see http://tinyurl.com/qzobc69
+// for details of how filters are specified in the API. For each key in f a filter is
+// created that matches any one of the set of values specified for that key.
+// The created filter will only match when at least one of the requested values
+// matches for all of the requested keys. createFilters will silently skip over any
+// filter names that are not mapped.
+func createFilters(f map[string][]string) elasticsearch.Filter {
+	af := make(elasticsearch.AndFilter, 0, len(f))
+	for k, vals := range f {
+		filter, ok := filters[k]
+		if !ok {
+			continue
+		}
+		of := make(elasticsearch.OrFilter, 0, len(vals))
+		for _, v := range vals {
+			of = append(of, filter(v))
+		}
+		af = append(af, of)
+	}
+	return af
+}
+
+// filters contains a mapping from a filter parameter in the API to a
+// function that will generate an elasticsearch query DSL filter for the
+// given value.
+var filters = map[string]func(string) elasticsearch.Filter{
+	"description": descriptionFilter,
+	"name":        termFilter("CharmMeta.Name"),
+	"owner":       ownerFilter,
+	"provides":    termFilter("CharmProvidedInterfaces"),
+	"requires":    termFilter("CharmRequiredInterfaces"),
+	"series":      seriesFilter,
+	"summary":     summaryFilter,
+	"tags":        tagsFilter,
+	"type":        typeFilter,
+}
+
+// descriptionFilter generates a filter that will match against the
+// description field of the charm data.
+func descriptionFilter(value string) elasticsearch.Filter {
+	return elasticsearch.QueryFilter{
+		Query: elasticsearch.MatchQuery{
+			Field: "CharmMeta.Description",
+			Query: value,
+			Type:  "phrase",
+		},
+	}
+}
+
+// ownerFilter generates a filter that will match against the
+// owner taken from the URL.
+func ownerFilter(value string) elasticsearch.Filter {
+	return elasticsearch.RegexpFilter{
+		Field:  "URL",
+		Regexp: "cs:~" + value + "/.*",
+	}
+}
+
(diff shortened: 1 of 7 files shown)


mhilton:
Please note a big chunk of this change is moving the search functionality into it's own file as store.go was getting a little long.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/375/

internal/charmstore/search.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return string(bytes)
+}
+
+// createFilters generates elasticsearch filters, based on the filter
+// parameters
+func createFilters(f map[string][]string) elasticsearch.Filter {
+	af := elasticsearch.AndFilter{}
```
rogpeppe:
```
af := make(elasticsearch.AndFilter, 0, len(filters))
```

to save a bit of garbage?

internal/charmstore/search.go
```diff
@@ -0,0 +1,284 @@
...
+	for k, v := range f {
+		af = append(af, filters[k](v))
+	}
+	return af
+}
+
+// filters contains a mapping from a filter parameter in the API to an
+// elasticsearch query dsl filter.
+var filters = map[string]func([]string) elasticsearch.Filter{
```
rogpeppe: very nice.

internal/charmstore/search.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	bytes, err := json.Marshal(qdsl)
+	if err != nil {
+		panic(err)
+	}
+	return string(bytes)
+}
+
+// createFilters generates elasticsearch filters, based on the filter
```
rogpeppe: A better description of the parameter map might be helpful here.
What's the key? Why is the value a []string, not a string?

Also (channeling frankban) please finish the comments with a full stop.

internal/charmstore/search.go
```diff
@@ -0,0 +1,284 @@
...
+	"provides":    termFilter("CharmProvidedInterfaces"),
+	"requires":    termFilter("CharmRequiredInterfaces"),
+	"series":      seriesFilter,
+	"summary":     summaryFilter,
+	"tags":        tagsFilter,
+	"type":        typeFilter,
+}
+
+// descriptionFilter generates a filter that will match against the
+// description field of the charm data.
```
rogpeppe, line 206: What does the values parameter imply here?

Is it that it's matching the description against any of the provided
values?

internal/charmstore/search.go
```diff
@@ -0,0 +1,284 @@
...
+		})
+	}
+	return filter
+}
+
+// tagsFilter generates a filter that will match against the "tags" field
+// in the data. For charms this is the Categories field and for bundles this
+// is the Tags field.
+func tagsFilter(values []string) elasticsearch.Filter {
+	filter := elasticsearch.OrFilter{}
```
rogpeppe: There seems to be a pattern emerging here with the OrFilter slice.

How about a higher level function that abstracts that out?

e.g.

```
func orFilter(singleFilter func(value string) elasticsearch.Filter) func([]string) elasticsearch.Filter {
    return func(values []string) elasticsearch.Filter {
        filter := make(elasticsearch.OrFilter, 0, len(values))
        for _, v := range values {
             filter = append(filter, singleFilter(v))
        }
        return filter
    }
}

func tagsFilter(value string) elasticsearch.Filter {
    return elasticsearch.TermFilter{
        Field: "CharmMeta.Categories",
        Value: value,
   }
}
```

Then you could have:

```
"tags": orFilter(tagsFilter),
```

above, and others too.

internal/charmstore/search.go

internal/charmstore/search.go
mhilton: hopefully it is clearer now.

internal/charmstore/search.go
mhilton, line 206: I have reworked this, I hope it is clearer.

internal/charmstore/search.go
mhilton: I moved the orFilter into createFilters as we always want to apply it.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/377/

internal/charmstore/search.go
```diff
@@ -0,0 +1,263 @@
...
+	var q elasticsearch.Query
+	if sp.Text == "" {
+		q = elasticsearch.MatchAllQuery{}
+	} else if sp.AutoComplete {
+		q = elasticsearch.MultiMatchQuery{
+			Query: sp.Text,
+			Fields: []string{
+				elasticsearch.BoostField("CharmMeta.Name.ngrams", 10),
+				elasticsearch.BoostField("CharmMeta.Description", 3),
+			},
```
urosj: I'd suggest adding a function that in the first step returns a map[string]integer that is specific for some case and in second step returns []string from whatever map[string]integer it gets from the first step. Instead of being hardcoded somewhere inside createDSL function ... it will also make createSearchDSL shorter and more clear. 
This way we can have different boosting mappings and just switch between them when experimenting, not touching/loosing previous values.

internal/charmstore/search.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+	} else {
+		q = elasticsearch.MultiMatchQuery{
+			Query: sp.Text,
+			Fields: []string{
+				elasticsearch.BoostField("CharmMeta.Name", 10),
+				elasticsearch.BoostField("CharmMeta.Description", 3),
+			},
```
urosj: Same as above.

internal/charmstore/search.go
```diff
@@ -0,0 +1,263 @@
...
+// StoreElasticSearch provides strongly typed methods for accessing the
+// elasticsearch database. These methods will not return errors if
+// elasticsearch is not configured, allowing them to be safely called even if
+// it is not enabled in this service.
+type StoreElasticSearch struct {
+	*elasticsearch.Database
+	Index string
+}
+
+const typeName = "entity"
```
urosj, line 25: I still think that in some future PR this typeName should be deduced from the mongodoc.Entity and that it is redundant in all of the ES APIs.

internal/charmstore/search.go
```diff
@@ -0,0 +1,263 @@
...
+// getID returns the id to be used for indexing the entity into ElasticSearch.
+func (ses *StoreElasticSearch) getID(entity *mongodoc.Entity) string {
+	return url.QueryEscape(entity.URL.String())
+}
+
+// Search searches for entities. The query is a json string which conforms
+// to the elasticsearch querydsl.
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl.html
+// Returns a slice containing each Id of the matching documents returned by
+// elasticsearch.
```
urosj, line 45: Awesome, thanks for the smart comments. I'd add that we used 1.3.4, that reference in the URL might change over time.

internal/charmstore/search.go
```diff
@@ -0,0 +1,263 @@
...
+	results, err := ses.Database.Search(ses.Index, typeName, query)
+	if err != nil {
+		return nil, err
+	}
+	ids := make([]string, 0, len(results.Hits.Hits))
+	for _, hit := range results.Hits.Hits {
+		ids = append(ids, hit.ID)
+	}
+	return ids, nil
+	// TODO: return more than just ids e.g. total, hits, score, information, serach time
```
urosj, line 59: Thank you for thinking in advance.

internal/charmstore/search.go
```diff
@@ -0,0 +1,263 @@
...
+	af := make(elasticsearch.AndFilter, 0, len(f))
+	for k, vals := range f {
+		of := make(elasticsearch.OrFilter, 0, len(vals))
+		for _, v := range vals {
+			of = append(of, filters[k](v))
+		}
+		af = append(af, of)
+	}
+	return af
+}
```
urosj, line 188: Nice. I'd do the same for any part of the query SDL (in another PR).

internal/charmstore/search_test.go
```diff
@@ -158,10 +172,173 @@ func (s *StoreSearchSuite) TestFilteredSearchWithUnAliasedField(c *gc.C) {
...
+		},
+	}
+	ids, err := s.store.Search(sp)
+	c.Assert(err, gc.IsNil)
+	c.Assert(ids, jc.SameContents, []string{
+		url.QueryEscape(exportTestCharms["wordpress"]),
+		url.QueryEscape(exportTestCharms["mysql"]),
+		url.QueryEscape(exportTestCharms["varnish"]),
+	})
+}
```
urosj, line 344: Add tests with multiple filters. Add tests with "wrong" filter. Add tests with mixture of proper and wrong filters, where wrong ones should only be ignored, but the proper ones remain used.

internal/elasticsearch/query.go
```diff
@@ -121,6 +127,37 @@ func (o OrFilter) MarshalJSON() ([]byte, error) {
...
+// regular expression.
+type RegexpFilter struct {
+	Field  string
+	Regexp string
+}
+
+func (r RegexpFilter) MarshalJSON() ([]byte, error) {
+	return marshalJSON("regexp", map[string]string{r.Field: r.Regexp})
+}
+
```
urosj, line 160: Shouldn't there be query_test.go along as well?

urosj:
:+1: Great work, @mhilton ! Some comments inside, mostly for subsequent PRs.

internal/charmstore/search.go
```diff
@@ -0,0 +1,263 @@
...
+
+// getID returns the id to be used for indexing the entity into ElasticSearch.
+func (ses *StoreElasticSearch) getID(entity *mongodoc.Entity) string {
+	return url.QueryEscape(entity.URL.String())
+}
+
+// Search searches for entities. The query is a json string which conforms
+// to the elasticsearch querydsl.
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl.html
+// Returns a slice containing each Id of the matching documents returned by
```
rogpeppe: s/Returns/It returns/

internal/charmstore/search.go
```diff
@@ -0,0 +1,263 @@
...
+// createFilters converts the filters requested with the serch API into
+// filters in the elasticsearch query DSL. Please see http://tinyurl.com/qzobc69 
+// for details of how filters are specified in the API. For each key in f a filter is
+// created that matches any one of the set of values specified for that key.
+// The created filter will only match when at least one of the requested values
+// matches for all of the requested keys.
+func createFilters(f map[string][]string) elasticsearch.Filter {
+	af := make(elasticsearch.AndFilter, 0, len(f))
+	for k, vals := range f {
+		of := make(elasticsearch.OrFilter, 0, len(vals))
```
rogpeppe, line 181: Much nicer having this factored out, thanks.

internal/charmstore/search_test.go
mhilton, line 344: Currently wrong filters will panic, the API rejects requests with filters we don't know about. But it is simple to deal with an errors in other parts of the system shouldn't cause cascading failures unnecessarily.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/380/

internal/charmstore/search.go
```diff
@@ -0,0 +1,285 @@
...
+	// If autocomplete is specified, the search will return only charms and
+	// bundles with a name that has text as a prefix.
+	AutoComplete bool
+	// Limit the search to items with attributes that match the specified filter value.
+	Filters map[string][]string
+	// Limit the number of returned items to the specified count.
+	Limit int
+}
+
+// Search the store for the given SearchParams.
```
rogpeppe, line 92: Please use full sentences:

   // Search searches ...
   // It returns ...

internal/charmstore/search.go
```diff
@@ -0,0 +1,285 @@
...
+func (ses *StoreElasticSearch) put(entity *mongodoc.Entity) error {
+	if ses == nil || ses.Database == nil {
+		return nil
+	}
+	return ses.PutDocument(ses.Index, typeName, ses.getID(entity), entity)
+}
+
+// getID returns the id to be used for indexing the entity into ElasticSearch.
+func (ses *StoreElasticSearch) getID(entity *mongodoc.Entity) string {
+	return url.QueryEscape(entity.URL.String())
```
frankban, line 38: Should this be entity.URL.Path()? I am not sure about what it's need to be passed to ES here.

internal/charmstore/search.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// Search searches for entities. The query is a json string which conforms
```
frankban, line 41: ... the query is a JSON string...

internal/charmstore/search.go
rogpeppe, line 38: shouldn't PutDocument be doing this escaping itself?

internal/charmstore/search.go
```diff
@@ -0,0 +1,285 @@
...
+// to the elasticsearch querydsl.
```
frankban, line 42: /// to the elasticsearch query dsl?

internal/charmstore/search.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	return string(bytes)
+}
+
+// createFilters converts the filters requested with the serch API into
+// filters in the elasticsearch query DSL. Please see http://tinyurl.com/qzobc69
+// for details of how filters are specified in the API. For each key in f a filter is
+// created that matches any one of the set of values specified for that key.
+// The created filter will only match when at least one of the requested values
+// matches for all of the requested keys. createFilters will silently skip over any
```
rogpeppe, line 172: // Any filter names that are not defined in the filters map will be silently skipped.
(to avoid the lower-case at start of sentence awkwardness) ?

internal/charmstore/search.go
```diff
@@ -0,0 +1,285 @@
...
+	return ids, nil
+	// TODO: return more than just ids e.g. total, hits, score, information, serach time
+}
+
+// ExportToElasticSearch reads all of the mongodoc Entities and writes
+// them to elasticsearch
+func (store *Store) ExportToElasticSearch() error {
+	var result mongodoc.Entity
+	iter := store.DB.Entities().Find(nil).Iter()
+	defer iter.Close()
```
frankban, line 67: iter.Close is deferred here but then explicitly called below. Is this the desired behavior? If so, perhaps we should add a comment.

internal/charmstore/search.go
```diff
@@ -0,0 +1,285 @@
...
+			Value: value,
+		}
+	}
+}
+
+// bundleFilter is a filter that matches against bundles, based on
+// the URL.
+var bundleFilter = seriesFilter("bundle")
+
+// typeFilter generates a filter that is used to match either only charms,
```
rogpeppe, line 278: s/typeFilter/entityTypeFilter/ ?

internal/charmstore/search_test.go
```diff
@@ -158,10 +172,233 @@ func (s *StoreSearchSuite) TestFilteredSearchWithUnAliasedField(c *gc.C) {
...
+		Filters: map[string][]string{
+			"provides": {"mysql"},
 		},
 	}
 	ids, err := s.store.Search(sp)
 	c.Assert(err, gc.IsNil)
 	c.Assert(ids, jc.DeepEquals, []string{url.QueryEscape(exportTestCharms["mysql"])})
 }
+
+func (s *StoreSearchSuite) TestSearchWithRequiresFilter(c *gc.C) {
```
rogpeppe, line 243: ISTM that a bunch of these fit exactly the same pattern and could be described
in a table, making it easier to see if there are any subtle differences.

internal/elasticsearch/query.go
```diff
@@ -39,10 +39,16 @@ func (m MatchAllQuery) MarshalJSON() ([]byte, error) {
...
 }
 
 func (m MatchQuery) MarshalJSON() ([]byte, error) {
-	return marshalJSON("match", map[string]interface{}{m.Field: m.Query})
+	params := map[string]interface{}{"query": m.Query}
+	if m.Type != "" {
+		params["type"] = m.Type
+	}
+
+	return marshalJSON("match", map[string]interface{}{m.Field: params})
```
rogpeppe, line 51: I think "marshalJSON" isn't a great name here - it gives no indication of how it's
different from json.Marshal.

marshalAttrVal? marshalJSONAttr? marshalJSONKeyVal? marshalPair ?

rogpeppe:
Great stuff! LGTM with a few minor suggestions above and assuming the test coverage is good.

frankban:
:+1: thank you!

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/386/

mhilton:
:shipit:

jujugui:

PR #145, opened 2014-10-14
csclient: new package

rogpeppe:
This is the start of the client-side charm store code.

We also factor out the JSON equality checking so that the csclient tests
can use it.

### DIFF
--- a/csclient/csclient.go
+++ b/csclient/csclient.go
@@ -1,0 +1,119 @@
+// The csclient package provides access to the charm store API.
+package csclient
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/juju/errgo"
+
+	"github.com/juju/charmstore/params"
+)
+
+const apiVersion = "v4"
+
+// Client represents the client side of a charm store.
+type Client struct {
+	params Params
+}
+
+// Params holds parameters for creating a new charm store client.
+type Params struct {
+	// URL holds the root endpoint URL of the charmstore,
+	// with no trailing slash, not including the version.
+	// For example http://charms.ubuntu.com
+	// TODO default this to global charm store address.
+	URL string
+
+	// User and Password hold the authentication credentials
+	// for the client. If User is empty, no credentials will be
+	// sent.
+	User     string
+	Password string
+
+	// HTTPClient holds the HTTP client to use when making
+	// requests to the store. If nil, http.DefaultClient will
+	// be used.
+	HTTPClient *http.Client
+}
+
+// New returns a new charm store client.
+func New(p Params) *Client {
+	if p.HTTPClient == nil {
+		p.HTTPClient = http.DefaultClient
+	}
+	return &Client{
+		params: p,
+	}
+}
+
+// Do makes an arbitrary request to the charm store.
+// It adds appropriate headers to the given HTTP request,
+// sends it to the charm store and parses the result
+// as JSON into the given result value, which should be a pointer to the
+// expected data, but may be nil if no result is expected.
+//
+// The URL field in the request is ignored and overwritten.
+//
+// This is a low level method - more specific Client methods
+// should be used when possible.
+func (c *Client) Do(req *http.Request, path string, result interface{}) error {
+	if c.params.User != "" {
+		userPass := c.params.User + ":" + c.params.Password
+		authBasic := base64.StdEncoding.EncodeToString([]byte(userPass))
+		req.Header.Set("Authorization", "Basic "+authBasic)
+	}
+
+	// Prepare the request.
+	if !strings.HasPrefix(path, "/") {
+		return errgo.Newf("path %q is not absolute", path)
+	}
+	u, err := url.Parse(c.params.URL + "/" + apiVersion + path)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	req.URL = u
+
+	// Send the request.
+	resp, err := c.params.HTTPClient.Do(req)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	defer resp.Body.Close()
+
+	// Parse the response.
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


csclient/csclient.go
```diff
@@ -0,0 +1,119 @@
+// The csclient package provides access to the charm store API.
+package csclient
```
bac, line 2: Isn't 'csclient' a bit redundant since it is part of juju/charmstore? You do name the struct Client not CSClient. <shrug>

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/381/

csclient/csclient.go
```diff
@@ -0,0 +1,119 @@
...
+// Do makes an arbitrary request to the charm store.
+// It adds appropriate headers to the given HTTP request,
+// sends it to the charm store and parses the result
+// as JSON into the given result value, which should be a pointer to the
+// expected data, but may be nil if no result is expected.
+//
+// The URL field in the request is ignored and overwritten.
+//
+// This is a low level method - more specific Client methods
+// should be used when possible.
```
bac, line 64: Nice docstring.  Thanks.

csclient/csclient.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/juju/errgo"
+
+	"github.com/juju/charmstore/params"
+)
+
+const apiVersion = "v4"
+
+// Client represents the client side of a charm store.
+type Client struct {
```
mhilton, line 21: I'm sure there's a reason for this, but the Client and Params structs seem a bit convoluted to me. What was the motivation for having both?

csclient/csclient.go
```diff
@@ -0,0 +1,119 @@
...
+		return errgo.Notef(err, "cannot unmarshal response %q", sizeLimit(data))
+	}
+	return nil
+}
+
+func sizeLimit(data []byte) []byte {
+	if len(data) < 1024 {
+		return data
+	}
+	return append(data[0:1024], fmt.Sprintf(" ... [%d bytes omitted]", len(data)-1024)...)
```
bac: Isn't the rule if you use a value more than twice you need to define a constant?

csclient/csclient.go
```diff
@@ -0,0 +1,119 @@
...
+// It adds appropriate headers to the given HTTP request,
+// sends it to the charm store and parses the result
+// as JSON into the given result value, which should be a pointer to the
+// expected data, but may be nil if no result is expected.
+//
+// The URL field in the request is ignored and overwritten.
+//
+// This is a low level method - more specific Client methods
+// should be used when possible.
+func (c *Client) Do(req *http.Request, path string, result interface{}) error {
```
mhilton, line 65: We found with the search stuff that putting in logging at debug level can be very helpful in following what's going on across connections.

csclient/csclient_test.go
```diff
@@ -0,0 +1,298 @@
...
+}{{
+	about: "success",
+	path:  "/wordpress/expand-id",
+	expectResult: []params.ExpandedId{{
+		Id: "cs:utopic/wordpress-42",
+	}},
+}, {
+	about:        "success with nil result",
+	path:         "/wordpress/expand-id",
+	expectResult: nil,
```
bac: This test is surprising and should be described.  Why are the results different from the previous one?

csclient/csclient.go
rogpeppe, line 2: The difference is that "csclient" is a top-level identifier, whereas the Client type is always qualified with the package name when qualified. Thus we have csclient.Client, not client.Client. Examples from the standard library include io/ioutil (not io/util) and http/httptest (not http/test).

csclient/csclient.go
rogpeppe, line 21: This allows us to have a Client struct which potentially has mutable or derived state (although at the moment we're not using that capability). Another example in the charm store is v4.NewAPIHandler.

If all the parameters were just in Client, we wouldn't have the possibility of a New function.

csclient/csclient.go
rogpeppe, line 65: In general I think that it's easier to put logging in at the server level (and we already have debug.Handler for that as well). Perhaps we might want some debugging here in the future when talking to the real charm store, but I think we'll leave it for now.

mhilton:
:+1: LGTM

csclient/csclient.go
rogpeppe: Not a rule that I recognise. :-) Readability trumps any hard and fast rule in my view.

But changed anyway.

csclient/csclient_test.go
rogpeppe: We swithered about this. The test passes a nil parameter to Do if the expected result is nil, but perhaps it should be another field in the test instead.

bac:
:+1: with comments

csclient/csclient.go
mhilton: This is really pedantic, but your slice could cut a rune in half causing up to one rune's worth of corrupt output.

csclient/csclient.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	if result == nil {
+		// The caller doesn't care about the response body.
+		return nil
+	}
+	if err := json.Unmarshal(data, result); err != nil {
+		return errgo.Notef(err, "cannot unmarshal response %q", sizeLimit(data))
+	}
+	return nil
+}
```
urosj, line 112: Nice construct!

urosj:
:+1: Excellent start!

rogpeppe:
:shipit:

jujugui:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/382/

csclient/csclient.go
rogpeppe: Github seems to have deleted this comment, but I just came
across it and wanted to reply.

Yes, I did think about this, and it would be nice to do it right.
A "slice to nearest rune" primitive would be nice to have.
Just using utf8.LastRune isn't good enough.
In the end, I didn't think it was worth the effort here,
but I am a big fan of multi-wide awareness in general,
so I'd like to do it right at some point.

PR #147, opened 2014-10-15
csclient: add Client.Meta method

rogpeppe:
Also add a Get method.

### DIFF
--- a/csclient/csclient.go
+++ b/csclient/csclient.go
@@ -86,16 +86,16 @@ func New(p Params) *Client {
 //	id, err := client.Meta(id, &result)
 func (c *Client) Meta(id *charm.Reference, result interface{}) (*charm.Reference, error) {
 	if result == nil {
-		panic(fmt.Errorf("expected valid result pointer, not nil"))
+		return nil, fmt.Errorf("expected valid result pointer, not nil")
 	}
 	resultv := reflect.ValueOf(result)
 	resultt := resultv.Type()
 	if resultt.Kind() != reflect.Ptr {
-		panic(fmt.Errorf("expected pointer, not %T", result))
+		return nil, fmt.Errorf("expected pointer, not %T", result)
 	}
 	resultt = resultt.Elem()
 	if resultt.Kind() != reflect.Struct {
-		panic(fmt.Errorf("expected pointer to struct, not %T", result))
+		return nil, fmt.Errorf("expected pointer to struct, not %T", result)
 	}
 	resultv = resultv.Elem()
 
@@ -118,7 +118,7 @@ func (c *Client) Meta(id *charm.Reference, result interface{}) (*charm.Reference
 			// At some point in the future, it might be nice to
 			// support anonymous fields, but for now the
 			// additional complexity doesn't seem worth it.
-			panic(fmt.Errorf("anonymous fields not supported"))
+			return nil, fmt.Errorf("anonymous fields not supported")
 		}
 		apiName := field.Tag.Get("csclient")
 		if apiName == "" {
--- a/csclient/csclient.go
+++ b/csclient/csclient.go
@@ -1,16 +1,23 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
 // The csclient package provides access to the charm store API.
 package csclient
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"reflect"
 	"strings"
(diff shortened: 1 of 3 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/384/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/385/

csclient/csclient.go
```diff
@@ -52,6 +59,143 @@ func New(p Params) *Client {
...
+		field := resultt.Field(i)
+		if field.PkgPath != "" {
+			// Field is private; ignore it.
+			continue
+		}
+		if field.Anonymous {
+			// At some point in the future, it might be nice to
+			// support anonymous fields, but for now the
+			// additional complexity doesn't seem worth it.
+			panic(fmt.Errorf("anonymous fields not supported"))
```
mhilton: Is there a reason to panic, why not just skip it?

csclient/csclient.go
```diff
@@ -52,6 +59,143 @@ func New(p Params) *Client {
...
+	}
+	return buf.String()
+}
+
+// Get makes a GET request to the charm store, parsing the
+// result as JSON into the given result value, which should be
+// a pointer to the expected data, but may be nil if no result is
+// desired.
+func (c *Client) Get(path string, result interface{}) error {
+	// The only reason NewRequest can fail is if the URL
```
mhilton: The API documentation doesn't actually state this, so there could in future be other reasons to error, in this case I suspect we won't see them, but as this function returns an error already I see little reason not to check for it. Especially as the error handling code is probably less typing than explanation of why it isn't needed.

csclient/export_test.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package csclient
+
+var Hyphenate = hyphenate
```
mhilton, line 6: This one is out of curiosity. hyphenate is pretty self contained so could be a test in the csclient package, why choose to export it only for tests?

mhilton:
:+1: LGTM, with comments (mostly queries)

csclient/csclient.go
rogpeppe: If we want to support anonymous fields in the future,
we don't want people relying on them being skipped.

csclient/csclient.go
rogpeppe: done

csclient/export_test.go
rogpeppe, line 6: just so we don't need another test suite for internal tests.
In general, I prefer to use export_test rather than using
an internal test package when we're already using external
tests. An exception to that is when it's not easy to do
(for example when the internal functions use internal types).
There are no hard and fast rules here.

csclient/csclient.go
```diff
@@ -52,6 +59,143 @@ func New(p Params) *Client {
...
+//	}
+//	id, err := client.Meta(id, &result)
+func (c *Client) Meta(id *charm.Reference, result interface{}) (*charm.Reference, error) {
+	if result == nil {
+		panic(fmt.Errorf("expected valid result pointer, not nil"))
+	}
+	resultv := reflect.ValueOf(result)
+	resultt := resultv.Type()
+	if resultt.Kind() != reflect.Ptr {
+		panic(fmt.Errorf("expected pointer, not %T", result))
```
urosj: Khm. Not sure about panics. Panic in a CLI is fine, but what if a client is something else, a "chaos monkey" for example. Why not error if we know that it's there?

csclient/csclient.go
```diff
@@ -52,6 +59,143 @@ func New(p Params) *Client {
...
+	}
+	resultt = resultt.Elem()
+	if resultt.Kind() != reflect.Struct {
+		panic(fmt.Errorf("expected pointer to struct, not %T", result))
```
urosj: Same here.

csclient/csclient.go
urosj: Ah, I see I'm not the only one who finds panic too "harsh".

urosj:
:+1: with some "don't panic" comments.

jrwren:
:+1: Are there going to be strongly typed methods for each api call, with defined structure type as the response so that a user of this lib can prefer those and use this panicking one as a last resort?

rogpeppe:
:shipit:

jujugui:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/389/

PR #148, opened 2014-10-15
essync configuration parameters.

mhilton:
Added a -settings and a -mapping flag to essync to
allow the user to configure the elasticsearch index correctly
before syncing the data. This saves manually setting up the server.

### DIFF
--- a/cmd/essync/main.go
+++ b/cmd/essync/main.go
@@ -4,11 +4,13 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
 
+	"gopkg.in/errgo.v1"
 	"gopkg.in/mgo.v2"
 
 	"github.com/juju/charmstore/config"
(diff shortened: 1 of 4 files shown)


mhilton:
This is a small update to the essync tool to make it easier to sync into new indexes. To QA run `essync -settings=index.json -mapping=entity.json` and check that the indexes are created correctly.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/388/

jrwren:
:+1:

cmd/essync/main.go
```diff
@@ -53,5 +57,55 @@ func populate(confPath string) error {
 	if err != nil {
 		return err
 	}
+	if *settings != "" {
+		err = writeSettings(es, *index, *settings)
+		if err != nil {
```
rogpeppe, line 62: FWIW when there are no required side effects, and all else being equal,
I'd prefer to keep err variables local. It means there's less to look
for when scanning the code. In this case (and others further down)
I'd probably use the "if err := ...; err != nil {" idiom.

mhilton, line 62: I'm still getting used to that one. I've always done it this way in the past, but I've no strong opinion on it.

PR #154, opened 2014-10-16
Loggo for ESSync main

fabricematrat:


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/398/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/399/

cmd/essync/main.go
```diff
@@ -14,9 +14,17 @@ import (
 	"github.com/juju/charmstore/config"
 	"github.com/juju/charmstore/internal/charmstore"
 	"github.com/juju/charmstore/internal/elasticsearch"
+	"github.com/juju/errgo"
```
mhilton: I believe we should be using gopkg.in/errgo now

cmd/essync/main.go
```diff
@@ -14,9 +14,17 @@ import (
...
+	"github.com/juju/loggo"
 )
 
-var index = flag.String("index", "charmstore", "Name of index to populate.")
+var logger = loggo.GetLogger("essync")
+var failsLogger = loggo.GetLogger("charmload_v4.loadfails")
```
rogpeppe: Is this used?

cmd/essync/main.go
```diff
@@ -29,29 +37,41 @@ func main() {
...
 func populate(confPath string) error {
+	if *loggingConfig != "" {
+		if err := loggo.ConfigureLoggers(*loggingConfig); err != nil {
+			return errgo.Notef(err, "cannot configure loggers")
+		}
+	}
 	conf, err := config.Read(confPath)
 	if err != nil {
-		return err
+		return errgo.Notef(err, "cannot read config: %s", confPath)
```
rogpeppe: I usually tend to reserve the colon for subsequent errors, so I'd phrase this as

```
errgo.Notef(err, "cannot read config file %q", confPath)
```

(note the file name in quotes too)

cmd/essync/main.go
```diff
@@ -29,29 +37,41 @@ func main() {
...
 	}
 	if conf.ESAddr == "" {
-		return fmt.Errorf("no elasticsearch-addr specified in %s", confPath)
+		return errgo.Newf("no elasticsearch-addr specified in %s", confPath)
```
rogpeppe: again I'd use %q not %s
(makes it easier to diagnose problems where there happens to be an extra piece
of white space for some reason)

cmd/essync/main.go
```diff
@@ -29,29 +37,41 @@ func main() {
...
 	}
 	es := &elasticsearch.Database{conf.ESAddr}
+
+	if conf.MongoURL == "" {
+		return errgo.Newf("no mongodb-addr specified in %s", confPath)
+	}
+	logger.Infof("config: %#v", conf)
```
rogpeppe: I'm not sure we want to be logging this. It's got the password in.

cmd/essync/main.go
```diff
@@ -29,29 +37,41 @@ func main() {
...
+
 	session, err := mgo.Dial(conf.MongoURL)
 	if err != nil {
-		return err
+		return errgo.Notef(err, "cannot dial mongo at: %s", conf.MongoURL)
```
rogpeppe: "cannot dial mongo at %q"

cmd/essync/main.go
```diff
@@ -29,29 +37,41 @@ func main() {
...
 	}
 	defer session.Close()
 	db := session.DB("juju")
 	store, err := charmstore.NewStore(db, &charmstore.StoreElasticSearch{es, *index})
 	if err != nil {
-		return err
+		return errgo.Notef(err, "unable to create store for ESSync")
 	}
+	logger.Infof("start export to Elastic Search")
```
rogpeppe: This seems more like a debug level log to me:

   logger.Debugf("starting export to Elastic Search")

?

rogpeppe:
LGTM modulo a few trivial suggestions above. Thanks!

cmd/essync/main.go
mhilton: charmload_v4 looks wrong here.

cmd/essync/main.go
```diff
@@ -29,29 +37,41 @@ func main() {
...
 	}
 	if err := populate(flag.Arg(0)); err != nil {
-		fmt.Fprintf(os.Stderr, "cannot populate elasticsearch: %v", err)
+		logger.Errorf("cannot populate elasticsearch: %v", err)
 		os.Exit(1)
 	}
 }
 
 func populate(confPath string) error {
+	if *loggingConfig != "" {
```
mhilton: This appears to be in the wrong place. Configuring logging seems to have little to do with population, perhaps move this into main, or it's own function.

cmd/essync/main.go
```diff
@@ -29,29 +37,41 @@ func main() {
...
-		return err
+		return errgo.Notef(err, "cannot read config: %s", confPath)
 	}
 	if conf.ESAddr == "" {
-		return fmt.Errorf("no elasticsearch-addr specified in %s", confPath)
+		return errgo.Newf("no elasticsearch-addr specified in %s", confPath)
 	}
 	es := &elasticsearch.Database{conf.ESAddr}
+
+	if conf.MongoURL == "" {
```
mhilton: This is unnecessary, config.Read will fail if MongoURL isn't set. Elasticsearch is different because it is optional for the rest of the system.

mhilton:
Looks pretty good, it is going to clash with https://github.com/juju/charmstore/pull/148 though. It might be worth rolling those changes into this (they're fairly small).

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/400/

cmd/essync/main.go
```diff
@@ -14,9 +14,16 @@ import (
 	"github.com/juju/charmstore/config"
 	"github.com/juju/charmstore/internal/charmstore"
 	"github.com/juju/charmstore/internal/elasticsearch"
+	"github.com/juju/loggo"
```
mhilton: These includes are in the wrong place, they are really library packages so should go in the second set. For the purposes of this project only things that start github.com/juju/charmstore should go in the third set.

cmd/essync/main.go
```diff
@@ -28,8 +35,13 @@ func main() {
 	if flag.NArg() != 1 {
 		flag.Usage()
 	}
+	if *loggingConfig != "" {
+		if err := loggo.ConfigureLoggers(*loggingConfig); err != nil {
+			return errgo.Notef(err, "cannot configure loggers")
```
mhilton: I would wager this doesn't compile now, you can't return a value from main.

fabricematrat: Why does the build work ?

cmd/essync/main.go
```diff
@@ -37,21 +49,25 @@ func main() {
...
-		return err
+		return errgo.Notef(err, "cannot read config file %q", confPath)
 	}
 	if conf.ESAddr == "" {
-		return fmt.Errorf("no elasticsearch-addr specified in %s", confPath)
+		return errgo.Newf("no elasticsearch-addr specified in config file %q", confPath)
 	}
 	es := &elasticsearch.Database{conf.ESAddr}
+
+	logger.Infof("config: %#v", conf)
```
mhilton: I'm not sure this should be output at info level, maybe debug, but I don't think it's needed at all. Maybe it would be better to have a message saying what file is being read and then if there is a problem the user can go and have a look at it.

cmd/essync/main.go
mhilton: It looks like because the CI server doesn't actually build this part of the code. essync doesn't really do much, so doesn't have any tests. The CI server only builds packages to test them. We should probably make it build everything too.

fabricematrat: yep, I think a make build would be good :)

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/401/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/402/

fabricematrat:
:shipit: Thanks @mhilton @rogpeppe

jujugui:

PR #159, opened 2014-10-17
Adding Metadata to search results.

mhilton:
Search results can now have metadata attached to them using the same mechanism
as the /meta handler uses. The change also includes:
- Some cleanup of the elasticsearch interface
- Search Time and Total in search response
- More tests

### DIFF
--- a/cmd/essync/main.go
+++ b/cmd/essync/main.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
@@ -23,11 +24,13 @@ var logger = loggo.GetLogger("essync")
 var (
 	index         = flag.String("index", "charmstore", "Name of index to populate.")
 	loggingConfig = flag.String("logging-config", "", "specify log levels for modules e.g. <root>=TRACE")
+	mapping       = flag.String("mapping", "", "File to use to configure the entity mapping.")
+	settings      = flag.String("settings", "", "File to use to configure the index.")
 )
 
 func main() {
 	flag.Usage = func() {
-		fmt.Fprintf(os.Stderr, "usage: %s <config path>\n", filepath.Base(os.Args[0]))
+		fmt.Fprintf(os.Stderr, "usage: %s [options] <config path>\n", filepath.Base(os.Args[0]))
 		flag.PrintDefaults()
 		os.Exit(2)
 	}
@@ -63,10 +66,58 @@ func populate(confPath string) error {
 	}
 	defer session.Close()
 	db := session.DB("juju")
-	store, err := charmstore.NewStore(db, &charmstore.StoreElasticSearch{es, *index})
+	store, err := charmstore.NewStore(db, &charmstore.StoreElasticSearch{es.Index(*index)})
 	if err != nil {
 		return errgo.Notef(err, "unable to create store for ESSync")
 	}
+	if *settings != "" {
+		if err := writeSettings(es, *index, *settings); err != nil {
+			return err
+		}
+	}
+	if *mapping != "" {
+		err = writeMapping(es, *index, "entity", *mapping)
+		if err != nil {
+			return err
+		}
+	}
 	logger.Debugf("starting export to Elastic Search")
 	return store.ExportToElasticSearch()
+}
+
+// writeSetttings creates a new index with the settings loaded from path. An error
+// will be returned from elasticsearch if the index already exists.
+func writeSettings(es *elasticsearch.Database, index, path string) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return errgo.Notef(err, "cannot read index settings")
+	}
+	defer f.Close()
+	dec := json.NewDecoder(f)
+	var data map[string]interface{}
+	if err := dec.Decode(&data); err != nil {
+		return errgo.Notef(err, "cannot read index settings")
+	}
+	if err := es.PutIndex(index, data); err != nil {
+		return errgo.Notef(err, "cannot set index settings")
+	}
+	return nil
+}
+
+// writeMapping writes the mapping loaded from path as documentType on index.
+func writeMapping(es *elasticsearch.Database, index, documentType, path string) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return errgo.Notef(err, "cannot read %s mapping", documentType)
+	}
+	defer f.Close()
+	dec := json.NewDecoder(f)
+	var data map[string]interface{}
+	if err := dec.Decode(&data); err != nil {
+		return errgo.Notef(err, "cannot read %s mapping", documentType)
+	}
+	if err := es.PutMapping(index, documentType, data); err != nil {
+		return errgo.Notef(err, "cannot set %s mapping", documentType)
+	}
+	return nil
 }
--- a/internal/charmstore/search.go
+++ b/internal/charmstore/search.go
@@ -4,10 +4,13 @@
 package charmstore
 
 import (
-	"encoding/json"
-	"net/url"
+	"crypto/sha1"
+	"encoding/base64"
+	"strings"
+	"time"
 
-	"github.com/juju/errgo"
+	"gopkg.in/errgo.v1"
+	"gopkg.in/juju/charm.v4"
 
 	"github.com/juju/charmstore/internal/elasticsearch"
 	"github.com/juju/charmstore/internal/mongodoc"
@@ -18,8 +21,7 @@ import (
 // elasticsearch is not configured, allowing them to be safely called even if
 // it is not enabled in this service.
 type StoreElasticSearch struct {
-	*elasticsearch.Database
-	Index string
+	*elasticsearch.Index
 }
 
 const typeName = "entity"
@@ -27,36 +29,48 @@ const typeName = "entity"
 // Put inserts the mongodoc.Entity into elasticsearch if elasticsearch
 // is configured.
 func (ses *StoreElasticSearch) put(entity *mongodoc.Entity) error {
-	if ses == nil || ses.Database == nil {
+	if ses == nil || ses.Index == nil {
 		return nil
 	}
-	return ses.PutDocument(ses.Index, typeName, ses.getID(entity), entity)
+	return ses.PutDocument(typeName, ses.getID(entity), entity)
 }
 
-// getID returns the id to be used for indexing the entity into ElasticSearch.
+// getID returns an ID for the elasticsearch document based on the contents of the
+// mongoDB document. This is to allow elasticsearch documents to be replaced with
+// updated versions when charm data is changed.
 func (ses *StoreElasticSearch) getID(entity *mongodoc.Entity) string {
-	return url.QueryEscape(entity.URL.String())
+	b := sha1.Sum([]byte(entity.URL.String()))
+	s := base64.URLEncoding.EncodeToString(b[:])
+	// Cut off any trailing = as there is no need for them and they will get URL escaped.
+	return strings.TrimRight(s, "=")
 }
 
-// Search searches for entities. The query is a json string which conforms
-// to the elasticsearch querydsl.
-// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl.html
-// It returns a slice containing each Id of the matching documents returned by
-// elasticsearch.
-func (ses *StoreElasticSearch) search(query string) ([]string, error) {
-	if ses == nil || ses.Database == nil {
-		return nil, nil
+// Search searches for matching entities in the configured elasticsearch index.
+// If there is no elasticsearch index configured then it will return an empty
+// SearchResult, as if no results were found.
+func (ses *StoreElasticSearch) search(sp SearchParams) (SearchResult, error) {
+	if ses == nil || ses.Index == nil {
+		return SearchResult{}, nil
 	}
-	results, err := ses.Database.Search(ses.Index, typeName, query)
+	q := createSearchDSL(sp)
+	q.Fields = append(q.Fields, "URL")
+	esr, err := ses.Search(typeName, q)
 	if err != nil {
-		return nil, err
+		return SearchResult{}, errgo.Mask(err)
 	}
-	ids := make([]string, 0, len(results.Hits.Hits))
-	for _, hit := range results.Hits.Hits {
-		ids = append(ids, hit.ID)
+	r := SearchResult{
+		SearchTime: time.Duration(esr.Took) * time.Millisecond,
+		Total:      esr.Hits.Total,
+		Results:    make([]*charm.Reference, 0, len(esr.Hits.Hits)),
 	}
-	return ids, nil
-	// TODO: return more than just ids e.g. total, hits, score, information, serach time
+	for _, h := range esr.Hits.Hits {
+		ref, err := charm.ParseReference(h.Fields.GetString("URL"))
+		if err != nil {
+			return SearchResult{}, errgo.Notef(err, "invalid result %q", h.Fields.GetString("URL"))
+		}
+		r.Results = append(r.Results, ref)
+	}
+	return r, nil
 }
 
 // ExportToElasticSearch reads all of the mongodoc Entities and writes
@@ -64,7 +78,7 @@ func (ses *StoreElasticSearch) search(query string) ([]string, error) {
 func (store *Store) ExportToElasticSearch() error {
 	var result mongodoc.Entity
 	iter := store.DB.Entities().Find(nil).Iter()
-	defer iter.Close()
+	defer iter.Close() // Make sure we always close on error.
 	for iter.Next(&result) {
 		if err := store.ES.put(&result); err != nil {
 			return errgo.Notef(err, "cannot index %s", result.URL)
@@ -76,7 +90,7 @@ func (store *Store) ExportToElasticSearch() error {
 	return nil
 }
 
-// SearchParams represents the search parameters used by the store.
+// SearchParams represents the search parameters used to search the store.
 type SearchParams struct {
 	// The text to use in the full text search query.
 	Text string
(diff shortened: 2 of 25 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/414/

mhilton:
This one's a bit big, sorry. To QA (Assuming you already have a full mongoDB):
1. Delete any existing index `http DELETE :9200/charmstore`
2. Populate the search database `essync -settings=internal/elasticsearch/definiitions/index.json` -mapping=internal/elasticsearch/definiitions/entity.json charmd.yaml`
3. Run charmstore`charmd charmd.yaml`
4. Search to your hearts content`http :8080/v4/search?name=mediawiki&include=archive-size`

cmd/essync/main.go
```diff
@@ -63,10 +66,60 @@ func populate(confPath string) error {
 	}
 	defer session.Close()
 	db := session.DB("juju")
-	store, err := charmstore.NewStore(db, &charmstore.StoreElasticSearch{es, *index})
+	store, err := charmstore.NewStore(db, &charmstore.StoreElasticSearch{es.Index(*index)})
 	if err != nil {
 		return errgo.Notef(err, "unable to create store for ESSync")
 	}
+	if *settings != "" {
+		err = writeSettings(es, *index, *settings)
```
rogpeppe: minor point:

   if err := writeSettings(...); err != nil {
   }

is preferable when reasonable. saves looking for potential side effects.

cmd/essync/main.go
```diff
@@ -63,10 +66,60 @@ func populate(confPath string) error {
...
+		}
+	}
 	logger.Debugf("starting export to Elastic Search")
 	return store.ExportToElasticSearch()
 }
+
+func writeSettings(es *elasticsearch.Database, index, fn string) error {
+	f, err := os.Open(fn)
+	if err != nil {
+		return errgo.NoteMask(err, "cannot read index settings")
```
rogpeppe: Or Notef.
I tend only to use NoteMask when I want to specifically allow an error type.

cmd/essync/main.go
```diff
@@ -63,10 +66,60 @@ func populate(confPath string) error {
...
+	}
+	err = es.PutIndex(index, data)
+	if err != nil {
+		return errgo.NoteMask(err, "cannot set index settings")
+	}
+	return nil
+}
+
+func writeMapping(es *elasticsearch.Database, index, type_, fn string) error {
```
rogpeppe: s/type_/indexType/
s/fn/path/ (or fileName)
?
perhaps a doc comment might be useful here.

internal/charmstore/search.go
```diff
@@ -18,8 +21,7 @@ import (
 // elasticsearch is not configured, allowing them to be safely called even if
 // it is not enabled in this service.
 type StoreElasticSearch struct {
-	*elasticsearch.Database
-	Index string
+	*elasticsearch.Index
```
rogpeppe, line 24: Nice change, thanks.

internal/charmstore/search.go
```diff
@@ -27,15 +29,20 @@ const typeName = "entity"
...
+	return ses.PutDocument(typeName, ses.getID(entity), entity)
 }
 
-// getID returns the id to be used for indexing the entity into ElasticSearch.
+// getID returns an ID for the elasticsearch document based on the contents of the
+// mongoDB document. This is to allow elasticsearch documents to be replaced with
+// updated versions when charm data is changed.
 func (ses *StoreElasticSearch) getID(entity *mongodoc.Entity) string {
-	return url.QueryEscape(entity.URL.String())
+	b := sha1.Sum([]byte(entity.URL.String()))
```
rogpeppe, line 42: Any particular reason to use a hash here rather than, say charm.Quote(entity.URL.String()) ?

internal/charmstore/search.go
```diff
@@ -43,20 +50,29 @@ func (ses *StoreElasticSearch) getID(entity *mongodoc.Entity) string {
 // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl.html
 // It returns a slice containing each Id of the matching documents returned by
```
rogpeppe: This doc comment appears to be out of date now.

cmd/essync/main.go
mhilton: Sorry you mentioned that last time and I forgot to change it.

internal/charmstore/search.go
mhilton, line 42: I wanted to remove the temptation to use it as a meaningful value when processing the results. That ID is for elasticsearch to tell the difference between things not for us to get information from, so by making it opaque now it can potentially be changed in future without breaking code that uses it incorrectly.

internal/elasticsearch/elasticsearch.go
```diff
@@ -25,125 +27,210 @@ import (
...
+
 // EnsureID tests to see a document of the given index, type_, and id exists
 // in ElasticSearch.
 func (db *Database) EnsureID(index, type_, id string) (bool, error) {
-	// TODO (bac) We should limit the fields to the id to avoid retrieving
-	// data we don't use.
-	response, err := http.Get(db.url(index, type_, id, "_source"))
-	if err != nil {
-		return false, errgo.Mask(err)
+	if err := db.get(db.url(index, type_, id)+"?_source=false", nil, nil); err != nil {
```
urosj, line 57: This should also be used in the /debug/status call when ES is not nil. Will add a todo.

urosj:
:+1: 

Huge branch! I did QA - but due to lack of time I just skimmed through code. Looks fine on the first glance.

internal/charmstore/search.go
```diff
@@ -169,8 +192,8 @@ func createSearchDSL(sp SearchParams) string {
 // for details of how filters are specified in the API. For each key in f a filter is
 // created that matches any one of the set of values specified for that key.
 // The created filter will only match when at least one of the requested values
-// matches for all of the requested keys. createFilters will silently skip over any
-// filter names that are not mapped.
+// matches for all of the requested keys. Any filter names that are not defined
+// in the filters map will be silenty skipped.
```
rogpeppe: silently

internal/charmstore/search.go
```diff
@@ -214,12 +237,29 @@ func descriptionFilter(value string) elasticsearch.Filter {
 	}
 }
 
+// nameFilter generates a filter that will match against the
+// name of the charm or bundle.
+func nameFilter(value string) elasticsearch.Filter {
+	// expand any * wildcards in the name filter
```
rogpeppe: Is this behaviour (the \* wildcards) documented somewhere?

rogpeppe: Also, shouldn't you quota meta characters in the value other than \* ?

internal/charmstore/search.go
```diff
@@ -248,26 +288,36 @@ func summaryFilter(value string) elasticsearch.Filter {
...
-		elasticsearch.TermFilter{
-			Field: "CharmMeta.Categories",
-			Value: value,
-		},
-		elasticsearch.TermFilter{
-			Field: "BundleData.Tags",
-			Value: value,
-		},
+	tags := strings.Split(value, " ")
+	af := make(elasticsearch.AndFilter, 0, len(tags))
```
rogpeppe, line 289: will this do the right thing when value is "" ?

internal/elasticsearch/elasticsearch.go
```diff
@@ -25,125 +27,210 @@ import (
...
+func (db *Database) GetDocument(index, type_, id string, v interface{}) error {
+	if err := db.get(db.url(index, type_, id, "_source"), nil, v); err != nil {
 		return errgo.Mask(err)
 	}
-	if status != http.StatusOK {
-		return errgo.Newf("ElasticSearch GET response status: %d, body: %s", status, body)
+	return nil
+}
+
+// ListAllIndexes does a GET request to the Database and parses the string
```
rogpeppe: Most of this is implementation detail which doesn't deserve to be in a doc comment.

```
// ListAllIndexes returns the names of all the indexes in the database.
// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_list_all_indexes.html
```

should be sufficient, I think.

internal/elasticsearch/elasticsearch.go
```diff
@@ -25,125 +27,210 @@ import (
...
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_list_all_indexes.html
+func (db *Database) ListAllIndexes() ([]string, error) {
+	var result map[string]interface{}
+	if err := db.get(db.url("_aliases"), nil, &result); err != nil {
+		return nil, errgo.Mask(err)
 	}
-	err = json.Unmarshal(body, doc)
-	if err != nil {
-		return errgo.Notef(err, "cannot unmarshal body: %s", body)
+	var indices []string
```
rogpeppe: s/indices/indexes/

might as well be consistent in the spelling.

internal/elasticsearch/elasticsearch.go
```diff
@@ -25,125 +27,210 @@ import (
...
 	}
-	if status != http.StatusCreated {
-		return "", errgo.Newf("ElasticSearch POST response status: %d, body: %s", status, body)
-	}
-	var respdata map[string]interface{}
-	if err := json.Unmarshal(body, &respdata); err != nil {
-		return "", errgo.Notef(err, "cannot unmarshal body")
-	}
-	return respdata["_id"].(string), nil
+	return resp["_id"].(string), nil
```
rogpeppe: This will panic if the response isn't the expected string.
Better to return an error in this case, I think.

internal/elasticsearch/elasticsearch.go
```diff
@@ -25,125 +27,210 @@ import (
...
+// into tha value it points to.
+func (db *Database) delete(url string, body, v interface{}) error {
+	return db.do("DELETE", url, body, v)
+}
+
+// get makes a GET request to the database url. A non-nil body will be
+// sent with the request and if v is not nill then the response will be unmarshaled
+// into tha value it points to.
+func (db *Database) get(url string, body, v interface{}) error {
+	return db.do("get", url, body, v)
```
rogpeppe: Shouldn't this be "GET" ?

params/params.go
```diff
@@ -104,5 +104,12 @@ type SearchResult struct {
 	Meta map[string]interface{} `json:",omitempty"`
 }
 
+// SearchResponse holds the response from a search operation.
+type SearchResponse struct {
+	SearchTime time.Duration
```
rogpeppe, line 115: This might be hard to parse in Javascript. Do we care though?

rogpeppe:
LGTM modulo some remarks above.
Thanks a lot, this is really nice.

cmd/essync/main.go
mhilton: done

cmd/essync/main.go
mhilton: done

cmd/essync/main.go
mhilton: done, although I named the variable documentType rather than indexType as that describes it better.

internal/charmstore/search.go

internal/charmstore/search.go
mhilton: well spotted thanks.

internal/charmstore/search.go
mhilton: The documentation is here https://docs.google.com/a/canonical.com/document/d/1eo44uCVJfha6r0v2Jz-QxbbkHtsXrabTpj_Q_ulcis4/edit#heading=h.8jimvw5a50mr

internal/charmstore/search.go
mhilton, line 289: I don;t think it did, I have made it skip empty strings.

internal/elasticsearch/elasticsearch.go
mhilton: done.

internal/elasticsearch/elasticsearch.go
mhilton: I've replaced the map with a structure to unmarshal into.

internal/elasticsearch/elasticsearch.go
mhilton: It worked, but I'm not sure it should. Changed.

params/params.go
mhilton, line 115: It gets returned as a count in nanoseconds, which is fairly easy to parse, so I'm not going to worry.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/417/

internal/elasticsearch/elasticsearch.go
```diff
@@ -154,144 +243,104 @@ func (db *Database) url(pathParts ...string) string {
...
+	}
+	return s
 }
 
-type ErrNotFound struct {
-	errgo.Err
+// EscapeRegexp returns the supplied string with any special characters escaped.
+// A regular expression match on the returned string will match exactly the characters
+// in the supplied string.
+func EscapeRegexp(s string) string {
```
rogpeppe, line 318: just use regexp.QuoteMeta(s)
?

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -7,9 +7,10 @@ import (
 	"testing"
 
 	jujutesting "github.com/juju/testing"
+	//	jc "github.com/juju/testing/checkers"
```
rogpeppe: d

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/419/

internal/elasticsearch/elasticsearch.go
mhilton, line 318: elasticsearch regular expressions use a different set of metacharacters to the go ones, so that wouldn't work.

internal/charmstore/search_test.go
```diff
@@ -91,315 +75,266 @@ func (s *StoreSearchSuite) addCharmsToStore(store *Store) {
...
-		},
-	}
-	ids, err := s.store.Search(sp)
-	c.Assert(err, gc.IsNil)
-	c.Assert(ids, jc.SameContents, []string{
-		url.QueryEscape(exportTestCharms["wordpress"]),
-		url.QueryEscape(exportTestCharms["mysql"]),
-		url.QueryEscape(exportTestCharms["varnish"]),
-	})
+var ResultsContain gc.Checker = &resultsContainChecker{
```
rogpeppe: Could we have a doc comment here please, describing the
types of the arguments to the checker?

internal/elasticsearch/elasticsearch_test.go
mhilton: done

internal/charmstore/search_test.go
rogpeppe: tbh though, it's awkward enough writing a Checker that I'd probably
avoid doing it except when it's going to be useful in lots of places.

  func assertResultsContain(*gc.C, SearchResult, []string)

is probably nicer overall.

One other thing: the "conjtain" in ResultsContain seems to imply that the results
can be a superset of the expected values. By my superficial
reading of the code, that isn't actually the case... or is it?

internal/elasticsearch/elasticsearch.go
```diff
@@ -154,144 +243,104 @@ func (db *Database) url(pathParts ...string) string {
...
 }
 
-// Delete makes a DELETE request to the database using the path parameter.
-func (db *Database) delete(path string) (*http.Response, error) {
-	req, err := http.NewRequest("DELETE", db.url(path), nil)
-	if err != nil {
-		return nil, err
+// Index creates a reference to an index in the elasticsearch database.
+func (db *Database) Index(name string) *Index {
+	if db == nil {
```
rogpeppe: Why are we doing this test here but not in any of the other Database methods?

internal/v4/search.go
```diff
@@ -22,17 +22,26 @@ func (h *Handler) serveSearch(w http.ResponseWriter, req *http.Request) (interfa
...
-		sr := params.SearchResult{Id: id}
-		// TODO (mhilton) get metadata
-		results = append(results, sr)
+	response := params.SearchResponse{
+		SearchTime: results.SearchTime,
+		Total:      results.Total,
+		Results:    make([]params.SearchResult, 0, len(results.Results)),
 	}
-	return results, nil
+	for _, ref := range results.Results {
```
rogpeppe, line 35:
```
// TODO do this concurrently.
```

?

params/params.go
rogpeppe, line 115: Ah, cool. I thought perhaps it marshaled as its string value.

rogpeppe:
LGTM modulo a few more comments above. Thanks!

internal/elasticsearch/elasticsearch.go
mhilton: That's because the Index method is to create a reference to access the database with a canned index, which is the only way we ever do it. This check means that if the database isn't configured the index won't be either. For all other methods it is a failure.

internal/v4/search.go
mhilton, line 35: done

internal/charmstore/search_test.go
mhilton: done

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/422/

internal/elasticsearch/elasticsearch.go
rogpeppe: It still seems a bit odd to me that we're calling methods on a nil Database, but
at the least this behaviour should be documented.

rogpeppe:
Still LGTM with some a doc comment on the Index method. FWIW can't we just put an "if" statement around the call to Index?

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/423/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/424/

mhilton:
:shipit:

jujugui:

PR #173, opened 2014-10-22
Boost Promulgated Charms

mhilton:
Charms in the flat namespace are boosted above similar ones
in a user namespace.

### DIFF
--- a/internal/charmstore/search.go
+++ b/internal/charmstore/search.go
@@ -164,14 +164,17 @@ func createSearchDSL(sp SearchParams) elasticsearch.QueryDSL {
 		}
 	}
 
-	// Attenuation function
 	q = elasticsearch.FunctionScoreQuery{
 		Query: q,
 		Functions: []elasticsearch.Function{
-			{
+			elasticsearch.DecayFunction{
 				Function: "linear",
 				Field:    "UploadTime",
 				Scale:    "365d",
+			},
+			elasticsearch.BoostFactorFunction{
+				Filter:      ownerFilter(""),
+				BoostFactor: 1.25,
 			},
 		},
 	}
--- a/internal/charmstore/search_test.go
+++ b/internal/charmstore/search_test.go
@@ -293,6 +293,24 @@ func (s *StoreSearchSuite) TestLimitTestSearch(c *gc.C) {
 	c.Assert(res.Results, gc.HasLen, 1)
 }
 
+func (s *StoreSearchSuite) TestPromulgatedRank(c *gc.C) {
+	charmArchive := testing.Charms.CharmDir("varnish")
+	url := charm.MustParseReference("cs:trusty/varnish-1")
+	s.store.AddCharmWithArchive(url, charmArchive)
+	err := s.store.ExportToElasticSearch()
+	s.store.ES.Database.RefreshIndex(s.TestIndex)
+	sp := SearchParams{
+		Filters: map[string][]string{
+			"name": {"varnish"},
+		},
+	}
+	res, err := s.store.Search(sp)
+	c.Assert(err, gc.IsNil)
+	c.Assert(res.Results, gc.HasLen, 2)
+	c.Assert(res.Results[0].String(), gc.Equals, "cs:trusty/varnish-1")
+	c.Assert(res.Results[1].String(), gc.Equals, exportTestCharms["varnish"])
+}
+
 // assertSearchResults checks that the results obtained from a search are the same
 // as those in the expected set, but in any order.
 func assertSearchResults(c *gc.C, obtained SearchResult, expected []string) {
--- a/internal/elasticsearch/query.go
+++ b/internal/elasticsearch/query.go
@@ -20,6 +20,9 @@ type Filter interface {
 	json.Marshaler
 }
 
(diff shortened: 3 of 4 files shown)


mhilton:
To QA perform a query with lots of results e.g. http://localhost:8080/v4/search?text=mysql&limit=100 and observe that the promulgated charms come first.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/441/

internal/charmstore/search.go
```diff
@@ -164,15 +164,18 @@ func createSearchDSL(sp SearchParams) elasticsearch.QueryDSL {
...
 		Query: q,
 		Functions: []elasticsearch.Function{
-			{
+			elasticsearch.DecayFunction{
 				Function: "linear",
 				Field:    "UploadTime",
 				Scale:    "365d",
 			},
+			elasticsearch.BoostFactorFunction{
+				Filter:      ownerFilter(""),
```
mitechie, line 176: I know this is outside of this branch, but this really needs to be an is_promulgated field in the index to boost on as a blank owner isn't always true/legit in the case of the openstack charms, juju-gui, etc.

mitechie:
Just for reference, the boostings we used in charmworld can be found at http://bazaar.launchpad.net/~juju-gui-bot/charmworld/trunk/view/head:/charmworld/search.py#L33

mitechie:
ah and not in that list is a 10 boot for the promulgated later in the code.

urosj:
:+1:

internal/charmstore/search.go
```diff
@@ -164,15 +164,18 @@ func createSearchDSL(sp SearchParams) elasticsearch.QueryDSL {
 		}
 	}
 
-	// Attenuation function
 	q = elasticsearch.FunctionScoreQuery{
 		Query: q,
 		Functions: []elasticsearch.Function{
-			{
+			elasticsearch.DecayFunction{
```
urosj, line 170: Somehow, decay function is not working properly.
http :8080/v4/"search?text=mongo&autocomplete=1"

 "Results":
 "Id": "cs:trusty/mongodb-0", "Id": "cs:trusty/mongodb-1", "Id": "cs:trusty/mongodb-2", "Id": "cs:trusty/mongodb-3", ...

But:
http get :8080/v4/mongodb-3/meta/archive-upload-time
"UploadTime": "2014-10-13T08:23:13.232Z"

http get :8080/v4/mongodb-0/meta/archive-upload-time
"UploadTime": "2014-10-13T08:23:11.006Z"

Also, this is a reason why we should never run promulgation with --p > 1, then the revision numbers will be out of date. For LP charms we should also look at extra info about the revision change dates, but that is getting more complex as it needs to be right now.

mhilton, line 170: The decay function decays linearly over the course of a year, and your quibbling about 2.226 seconds. In terms of the decay function that's just noise.

internal/charmstore/search.go
rogpeppe, line 176: A charm can be promulgated without losing its ~owner part?
I thought that was the point of promulgation - that it makes a charm
available in the global non-user name space.

internal/elasticsearch/query.go
```diff
@@ -118,6 +126,20 @@ func (f Function) MarshalJSON() ([]byte, error) {
 	})
 }
 
+type BoostFactorFunction struct {
+	Filter      Filter
+	BoostFactor float64
+}
+
+func (f BoostFactorFunction) MarshalJSON() ([]byte, error) {
```
rogpeppe: Do we need this - can't we just annotate the fields of the struct?

```
type BoostFactorFunction struct {
    Filter Filter `json:"filter,omitempty"`
    BoostFactor float64 `json:"boost_factor"`
}
```

internal/charmstore/search_test.go
```diff
@@ -293,6 +293,24 @@ func (s *StoreSearchSuite) TestLimitTestSearch(c *gc.C) {
 	c.Assert(res.Results, gc.HasLen, 1)
 }
 
+func (s *StoreSearchSuite) TestPromulgatedRank(c *gc.C) {
```
rogpeppe, line 296: Have you checked that this test fails when the new ranking code is turned off?

rogpeppe:
LGTM with a few minor comments and a nagging feeling that the ranking test might just be passing by chance.

internal/elasticsearch/query.go
mhilton: It needs to be there to satisfy the interface, but the function could be simpler. I'll change it.

internal/charmstore/search_test.go
mhilton, line 296: err no, and I wouldn't trust that that would always happen the other way round either. I've QA'd to check  that the boost works on much bigger data sets. This will just give a warning if we make a change that stops this from being true.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/442/

mhilton:
:shipit:

jujugui:

PR #177, opened 2014-10-22
router: implement RelativeURLPath

rogpeppe:
We can use this for generating correct relative URLs in pages that
we serve (we'll use it initially in the generated SVGs for bundles).

Also add some copyright notices that were needed.

### DIFF
--- a/internal/router/fieldinclude.go
+++ b/internal/router/fieldinclude.go
@@ -1,3 +1,6 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
 package router
 
 import (
--- a/internal/router/util.go
+++ b/internal/router/util.go
@@ -6,6 +6,7 @@ package router
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/juju/utils/jsonhttp"
 	"gopkg.in/errgo.v1"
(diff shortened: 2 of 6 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/444/

internal/router/util.go
```diff
@@ -110,3 +111,44 @@ func (mux *ServeMux) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 }
+
+// RelativeURLPath returns a relative URL path that is lexically equivalent to
+// targpath when interpreted by url.URL.ResolveReference.
+// On succes, the returned path will always be relative to basePath, even if basePath
+// and targPath share no elements. An error is returned if targPath can't
+// be made relative to basePath (for example when either basePath
+// or targetPath are non-absolute).
+func RelativeURLPath(basePath, targPath string) (string, error) {
+	if !strings.HasPrefix(basePath, "/") {
```
mhilton, line 122: is there any reason to prefer this to `basePath[0] != '/'`?

internal/router/util.go
```diff
@@ -110,3 +111,44 @@ func (mux *ServeMux) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	// Find the common prefix between the two paths:
+	var i int
+	for ; i < len(baseParts); i++ {
+		if i >= len(targParts) || baseParts[i] != targParts[i] {
+			break
+		}
+	}
+	dotdotCount := len(baseParts) - i
+	targOnly := targParts[i:]
+	result := make([]string, 0, dotdotCount+len(targOnly)+1)
```
mhilton, line 147: You could potentially use `strings.Repeat("../", dotdotCount)` instead of this loop, I'm not sure it would make it better though.

mhilton:
:+1: LGTM

internal/router/util.go
rogpeppe, line 122: It won't panic if the string is empty.

internal/router/util.go
rogpeppe, line 147: I don't think that would work as we're building up a slice of strings (path elements) here, not a single string, and if we added a single element with all those .. elements in, we would add an extra / when appending targOnly.

internal/router/fieldinclude.go
```diff
@@ -1,3 +1,6 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
```
urosj, line 3: Thank you!

internal/router/util.go
```diff
@@ -110,3 +111,44 @@ func (mux *ServeMux) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	targParts = targParts[0 : len(targParts)-1]
+
+	// Find the common prefix between the two paths:
+	var i int
+	for ; i < len(baseParts); i++ {
+		if i >= len(targParts) || baseParts[i] != targParts[i] {
+			break
+		}
+	}
+	dotdotCount := len(baseParts) - i
```
urosj, line 145: Hehe, slick name.

urosj:
:+1:

rogpeppe:
:shipit:

jujugui:

PR #178, opened 2014-10-22
Only index latest charm version.

mhilton:
Use elasticsearch versioning to make sure that only
the latest version of the charm or bundle is indexed.
Therefore only returning the latest version in search
results.

### DIFF
--- a/internal/elasticsearch/elasticsearch.go
+++ b/internal/elasticsearch/elasticsearch.go
@@ -92,7 +92,9 @@ func (db *Database) ListAllIndexes() ([]string, error) {
 }
 
 // PostDocument creates a new auto id document with the given index and _type
-// and returns the generated id of the document.
+// and returns the generated id of the document. The type_ parameter controls how
+// the document will be mapped in the index. See http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-index_.html
+// for more details.
 func (db *Database) PostDocument(index, type_ string, doc interface{}) (string, error) {
 	var resp struct {
 		ID string `json:"_id"`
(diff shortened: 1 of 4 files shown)


mhilton:
Before QAing this you will need to clear the current elasticsearch index as it will not overwrite existing stored charm and bundle data.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/445/

urosj:
I propose that this is the default index. But that we keep another index of all documents as well, which is used if and only if "&search-all" flag is added to search command. (As a next PR, not part of this one)

internal/elasticsearch/elasticsearch.go
```diff
@@ -112,6 +112,21 @@ func (db *Database) PutDocument(index, type_, id string, doc interface{}) error
...
 
+// PutDocumentVersion creates or updates the document with the given index
+// if the version is equal or newer than the currently stored version.
+// PutDocumentVersion returns a bool indicating if the data was stored and an error.
+// If the data was not stored due to a version conflict the returned value will be false, nil.
+func (db *Database) PutDocumentVersion(index, type_, id string, version int64, doc interface{}) (bool, error) {
+	url := fmt.Sprintf("%s?version=%d&version_type=external_gte", db.url(index, type_, id), version)
+	if err := db.put(url, doc, nil); err != nil {
+		if eserr, ok := err.(ElasticSearchError); ok && eserr.Status == http.StatusConflict {
+			return false, nil
```
urosj, line 134: Why is conflict not an error?

internal/elasticsearch/elasticsearch.go
```diff
@@ -258,6 +273,10 @@ func (i *Index) PutDocument(type_, id string, doc interface{}) error {
 	return i.Database.PutDocument(i.Index, type_, id, doc)
 }
 
+func (i *Index) PutDocumentVersion(type_, id string, version int64, doc interface{}) (bool, error) {
```
urosj, line 292: A thought: should we always put documents with Index only, and not Database? Then we are sure, that index exists (and thus not exposing (db *Database) putDocumentVersion)?

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -87,6 +86,94 @@ func (s *Suite) TestSuccessfulPutUpdatedDocument(c *gc.C) {
...
+	c.Assert(err, gc.IsNil)
+	c.Assert(stored, gc.Equals, false)
+	var result map[string]string
+	err = s.ES.GetDocument(s.TestIndex, "testtype", "a", &result)
+	c.Assert(result["a"], gc.Equals, "b")
+	exists, err = s.ES.EnsureID(s.TestIndex, "testtype", "a")
+	c.Assert(err, gc.IsNil)
+	c.Assert(exists, gc.Equals, true)
+}
+
```
urosj, line 176: Nice tests.

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -181,6 +268,98 @@ func (s *Suite) TestIndexPutDocument(c *gc.C) {
...
```
urosj, line 362: Looks like duplication from previous tests? Cleanup not required in this PR though.

urosj:
:+1:

internal/elasticsearch/elasticsearch.go
rogpeppe, line 134: That's a reasonable question.
I'd be tempted to make this just return error, and define an error value, say:

   var ErrConflict = errgo.New("conflict inserting document")

and document that PutDocumentVersion returns this error on conflict.

internal/elasticsearch/elasticsearch.go
rogpeppe, line 292: doc comment please.

rogpeppe:
LGTM modulo a suggestion above. A nice solution to the problem.

internal/elasticsearch/elasticsearch.go
mhilton, line 134: because that's what we want it to do. If we try and push an old one it will return conflict, which isn't an error for out application.

internal/elasticsearch/elasticsearch.go
mhilton, line 292: It's useful to have both, especially for tests, I'm more inclined to keep it consistent with everything else.

internal/elasticsearch/elasticsearch.go
mhilton, line 134: rogpeppe: maybe , although I'd rather stay consistent with EnsureID.

internal/elasticsearch/elasticsearch_test.go
mhilton, line 362: These are the same tests, but on the index rather than the Database.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/453/

mhilton:
:shipit:

jujugui:

PR #185, opened 2014-10-23
internal/charmstore: add ZipFileReader and NewZipFile

rogpeppe:
These are the methods we'll use to create the Content on entities.

### DIFF
--- a/internal/charmstore/zip.go
+++ b/internal/charmstore/zip.go
@@ -1,0 +1,49 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package charmstore
+
+import (
+	"archive/zip"
+	"compress/flate"
+	"io"
+
+	"gopkg.in/errgo.v1"
+
+	"github.com/juju/charmstore/internal/mongodoc"
+)
+
+// ZipFileReader returns a reader that will read
+// content referred to by f within zipr, which should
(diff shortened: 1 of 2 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/461/

internal/charmstore/zip.go
```diff
@@ -0,0 +1,49 @@
...
+	offset, err := f.DataOffset()
+	if err != nil {
+		return mongodoc.ZipFile{}, errgo.Notef(err, "cannot determine data offset for %q", f.Name)
+	}
+	zf := mongodoc.ZipFile{
+		Offset: offset,
+		Size:   int64(f.CompressedSize64),
+	}
+	switch f.Method {
+	case zip.Store:
```
mhilton, line 42: I know you don't strictly need to, but I would put `zf.Compressed = false` here to make it clear it doesn't run into the next case.

mhilton:
:+1: LGTM with the one suggestion.

urosj:
:+1:

internal/charmstore/zip.go
rogpeppe, line 42: I'd prefer to assume that people understand Go switch semantics.
Otherwise we'll be putting a break after each clause too :)

rogpeppe:
:shipit:

jujugui:

PR #188, opened 2014-10-24
v4: implement meta/tags endpoint

rogpeppe:
As discussed online, we choose tags over categories if specified.

### DIFF
--- a/internal/v4/api.go
+++ b/internal/v4/api.go
@@ -59,17 +59,14 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
 		Meta: map[string]router.BulkIncludeHandler{
 			"archive-size":         h.entityHandler(h.metaArchiveSize, "size"),
 			"archive-upload-time":  h.entityHandler(h.metaArchiveUploadTime, "uploadtime"),
-			"bundle-metadata":      h.entityHandler(h.metaBundleMetadata, "bundledata"),
-			"bundle-unit-count":    h.entityHandler(h.metaBundleUnitCount, "bundleunitcount"),
 			"bundle-machine-count": h.entityHandler(h.metaBundleMachineCount, "bundlemachinecount"),
+			"bundle-metadata":      h.entityHandler(h.metaBundleMetadata, "bundledata"),
 			"bundles-containing":   h.entityHandler(h.metaBundlesContaining),
+			"bundle-unit-count":    h.entityHandler(h.metaBundleUnitCount, "bundleunitcount"),
 			"charm-actions":        h.entityHandler(h.metaCharmActions, "charmactions"),
 			"charm-config":         h.entityHandler(h.metaCharmConfig, "charmconfig"),
 			"charm-metadata":       h.entityHandler(h.metaCharmMetadata, "charmmeta"),
 			"charm-related":        h.entityHandler(h.metaCharmRelated, "charmprovidedinterfaces", "charmrequiredinterfaces"),
-			"manifest":             h.entityHandler(h.metaManifest, "blobname"),
-			"revision-info":        router.SingleIncludeHandler(h.metaRevisionInfo),
-			"stats":                h.entityHandler(h.metaStats),
 			"extra-info": h.puttableEntityHandler(
 				h.metaExtraInfo,
 				h.putMetaExtraInfo,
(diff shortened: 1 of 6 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/468/

mitechie:
I missed the conversation, why a different endpoint vs making it part of the charm/bundle metadata calls?

rogpeppe:
It _is_ part of the charm/bundle metadata calls. 

That is, it can now be used as an "include" in /meta/any along with all the other charm and bundle metadata endpoints. We continue to preserve the invariant that both meta/charm-metadata and meta/bundle-metadata are exactly derived from their files inside the archive blob.

mitechie:
I understand that it's a new meta endpoint, but Tags are merely us renaming the Category field in metadata.yaml. The goal is to, as quickly as we can, never have any more Categories and only have Tags. The eco team is currently working updating the bundles and charms to this end. So it seems odd, for the migration period, to create a new includable endpoint for a bit of data that is part of the metadata from the archive blob. It's temporary and when do we stop making every chunk of charm/bundle data not require a different include block? I guess I'd be more on board if we never expected to resolve the differences in the two and you had two long standing sources of the data we want, in a way like the README issue.  I don't know that it's the same though.

internal/v4/api.go
```diff
@@ -80,6 +77,10 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
 				h.putMetaExtraInfoWithKey,
 				"extrainfo",
 			),
+			"manifest":      h.entityHandler(h.metaManifest, "blobname"),
+			"revision-info": router.SingleIncludeHandler(h.metaRevisionInfo),
+			"stats":         h.entityHandler(h.metaStats),
+			"tags":          h.entityHandler(h.metaTags, "charmmeta", "bundledata"),
```
urosj, line 83: Thanks.

internal/v4/api.go
```diff
@@ -364,6 +365,24 @@ func (h *Handler) metaArchiveSize(entity *mongodoc.Entity, id *charm.Reference,
...
 
+// GET id/meta/tags
+// http://tinyurl.com/njyqwj2
+func (h *Handler) metaTags(entity *mongodoc.Entity, id *charm.Reference, path string, flags url.Values) (interface{}, error) {
+	var tags []string
+	switch {
+	case id.Series == "bundle":
+		tags = entity.BundleData.Tags
+	case len(entity.CharmMeta.Tags) > 0:
+		// TODO only return whitelisted tags.
```
urosj, line 376: I think the whitelist check will be done in the "publish" process and is not needed here.

internal/v4/api.go
```diff
@@ -364,6 +365,24 @@ func (h *Handler) metaArchiveSize(entity *mongodoc.Entity, id *charm.Reference,
...
+		tags = entity.CharmMeta.Tags
+	default:
+		tags = entity.CharmMeta.Categories
```
urosj, line 379: However, we need to filter legacy Categories instead ... after a second thought, no need for this check.

internal/v4/archive.go
```diff
@@ -369,6 +369,10 @@ func (e *entityCharm) Meta() *charm.Meta {
 	return e.CharmMeta
 }
 
+func (e *entityCharm) Metrics() *charm.Metrics {
```
urosj, line 372: Do we need this?

urosj:
:+1:

urosj:
@mitechie The /meta/charm-metadata will include Tags, if they exists, and whatever eco is doing will be used from day1. When Tags exist, Categories are "thrown away". I do not want to do "bundle migration issue" again in such a short period. When time comes and we introduce "Charm.v5" which obsoletes Categories and do all post-processing of ingested data, then Categories will be gone for good. The transition period does not affect use of Tags. And clients can choose/will ignore categories that are not whitelisted tags anyway.

internal/v4/api.go
frankban, line 376: Yeah, that's also an option.

internal/v4/archive.go
frankban, line 372: Yes we do because the Metrics method has been recently added to the Charm interface, so we need to define it in order to implement the interface.

frankban:
:shipit:

jujugui:

urosj:
Note, that this is not to be used in the classical GUI scenario, where information is retrieved with /meta/charm-metadata and the logic to selecting tags vs categories is implemented in the client instead. This might come handy in some other client.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/477/

PR #194, opened 2014-10-28
internal/v4: fix search error handling

rogpeppe:
We use the new HandleJSON API which doesn't give a ResponseWriter
to use.

### DIFF
--- a/internal/v4/search.go
+++ b/internal/v4/search.go
@@ -16,7 +16,7 @@ import (
 
 // GET search[?text=text][&autocomplete=1][&filter=value…][&limit=limit][&include=meta]
 // http://tinyurl.com/qzobc69
-func (h *Handler) serveSearch(w http.ResponseWriter, req *http.Request) (interface{}, error) {
+func (h *Handler) serveSearch(_ http.Header, req *http.Request) (interface{}, error) {
 	sp, err := parseSearchParams(req)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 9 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/482/

internal/v4/search_test.go
```diff
@@ -68,7 +70,7 @@ func getBundle(name string) *charm.BundleDir {
 	return ba
 }
 
-func (s *SearchSuite) TestParseSerchParams(c *gc.C) {
+func (s *SearchSuite) TestParseSearchParams(c *gc.C) {
```
mhilton, line 73: thanks

internal/v4/search.go
```diff
@@ -24,18 +24,22 @@ func (h *Handler) serveSearch(w http.ResponseWriter, req *http.Request) (interfa
...
+	// TODO(mhilton) collect the metadata concurrently.
 	for _, ref := range results.Results {
 		meta, err := h.Router.GetMetadata(ref, sp.Include)
 		if err != nil {
-			router.WriteError(w, errgo.Notef(err, "error retrieving metadata"))
+			// Unfortunately it is possible to get errors here due to
+			// internal inconsistency, so rather than throwing away
+			// all the search results, we just log the error and move on.
+			logger.Errorf("cannot retrieve metadata for %v: %v", ref, err)
+			continue
```
mhilton, line 42: I would argue that its better to return the result without the bit of metadata than to skip the result entirely. There are plenty of occasions where we don't include the metadata if it doesn't exist.

rogpeppe, line 42: It's explicitly documented that we omit metadata when it's not relevant to the entity (see https://docs.google.com/a/canonical.com/document/d/1TgRA7jW_mmXoKH3JiwBbtPvQu7WiM6XMrz1wSrhTMXw/edit#bookmark=id.uu3ojqeyq0b8). That is already catered for by GetMetadata. This case we're considering is different, I feel, because our database is internally inconsistent - this particular charm or bundle is actually unusable because the archive doesn't exist. If we were to include the entity, it would have no metadata at all, which wouldn't really be very useful.

mhilton, line 42: OK if that is the only failure mode then I don't see a problem with omitting it. If you can't do anything useful with the search result then don't return it seems like a good idea.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/485/

mhilton:
:+1: LGTM

frankban:
:shipit:

jujugui:

PR #210, opened 2014-11-10
internal/v4, internal/router: use concurrency

rogpeppe:
This might speed up search results considerably.

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -13,8 +13,10 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/juju/utils/jsonhttp"
+	"github.com/juju/utils/parallel"
 	"gopkg.in/errgo.v1"
 	charm "gopkg.in/juju/charm.v4"
 
@@ -506,6 +508,12 @@ func (r *Router) serveBulkMetaPutOne(req *http.Request, id string, val *json.Raw
 	return nil
 }
 
+// maxMetadataConcurrency specifies the maximum number
+// of goroutines started to service a given GetMetadata request.
+// 5 is enough to more that cover the number of metadata
+// group handlers in the current API.
+const maxMetadataConcurrency = 5
+
 // GetMetadata retrieves metadata for the given charm or bundle id,
 // including information as specified by the includes slice.
 func (r *Router) GetMetadata(id *charm.Reference, includes []string) (map[string]interface{}, error) {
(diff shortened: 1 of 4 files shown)


jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/522/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/523/

mhilton:
:+1: LGTM

internal/router/router.go
```diff
@@ -526,37 +530,53 @@ func (r *Router) GetMetadata(id *charm.Reference, includes []string) (map[string
 		includesByGroup[key] = append(includesByGroup[key], include)
 	}
 	results := make(map[string]interface{})
+	// TODO when the number of groups is 1 (a common case,
```
frankban, line 537: This is really a TODO for utils/parallel, correct?

internal/router/router.go
```diff
@@ -506,6 +508,8 @@ func (r *Router) serveBulkMetaPutOne(req *http.Request, id string, val *json.Raw
 	return nil
 }
 
+const maxMetadataConcurrency = 5
```
frankban, line 515: Is this number related on the usual number of meta info requested?
If so, could you please add a comment?
If not, is it possible to increase this so that it includes all the possible metadata endpoints?

internal/router/router.go
```diff
@@ -526,37 +530,53 @@ func (r *Router) GetMetadata(id *charm.Reference, includes []string) (map[string
...
-			return nil, errgo.Mask(err, errgo.Any)
-		}
-		for i, result := range groupResults {
-			// Omit nil results from map. Note: omit statically typed
-			// nil results too to make it easy for handlers to return
-			// possibly nil data with a static type.
-			// http://tinyurl.com/o5ptfkk
-			if !isNull(result) {
-				results[groupIncludes[i]] = result
+		g := g
```
frankban, line 544: Heh, so that the callable passed to Do correctly references the g at the time of the loop?

internal/router/router.go
```diff
@@ -526,37 +530,53 @@ func (r *Router) GetMetadata(id *charm.Reference, includes []string) (map[string
...
+				if !isNull(result) {
+					results[groupIncludes[i]] = result
+				}
+			}
+			mu.Unlock()
+			return nil
+		})
+	}
+	if err := run.Wait(); err != nil {
+		// We could have got multiple errors, but we'll only return one of them
```
frankban: Final dot to complete the sentence.

internal/router/router.go
rogpeppe, line 515: Not directly. It's related to the number of bulk hundler types that we can have for any given request. Currently in v4 there are only 3 (I mistakenly thought that the manifest was one of them, but it's also styled as an entity handler, which is not unreasonable, as it does need the entity too). So it does already cover all the possible metadata endpoints. I'll add a comment.

internal/router/router.go
rogpeppe, line 537: Yes. But it's an unusual enough case that it sits better here, I think. It might actually be better solved here, in fact.

internal/v4/search.go
```diff
@@ -29,23 +33,43 @@ func (h *Handler) serveSearch(_ http.Header, req *http.Request) (interface{}, er
...
+				return nil
+			}
+			response.Results[i] = params.SearchResult{
+				Id:   ref,
+				Meta: meta,
+			}
+			return nil
 		})
 	}
+	run.Wait()
```
frankban, line 59: So wait never returns an error if the functions passed to Do don't. 
I'd add a comment here, or just check the error as usual: if in the future those functions can fail, we don't want to skip the error checking here.

internal/v4/search.go
```diff
@@ -29,23 +33,43 @@ func (h *Handler) serveSearch(_ http.Header, req *http.Request) (interface{}, er
...
+	if missing == 0 {
+		return response, nil
+	}
+	// We're missing some results - shuffle all the results down to
+	// fill the gaps.
+	j := 0
+	for _, result := range response.Results {
```
frankban, line 66: Nice.

frankban:
:+1: very nice thank you!

internal/router/router.go
rogpeppe, line 544: Yup. Standard idiom.

internal/router/router.go
rogpeppe: done

internal/v4/search.go
rogpeppe, line 59: added a comment.

rogpeppe:
:shipit:

jujugui:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/524/

PR #214, opened 2014-11-11
CS to use denormalised ID values in ES

urosj:
QA: running ingestion to see how it behaves. Not landing before it ends.

### DIFF
--- a/internal/charmstore/search.go
+++ b/internal/charmstore/search.go
@@ -145,6 +145,8 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
 func queryFields(sp SearchParams) map[string]float64 {
 	fields := map[string]float64{
 		"URL.ngrams":              8,
+		"Name":                    8,
+		"Series":                  8,
 		"CharmMeta.Categories":    5,
 		"BundleData.Tags":         5,
 		"CharmProvidedInterfaces": 3,
@@ -266,34 +268,36 @@ func descriptionFilter(value string) elasticsearch.Filter {
 // nameFilter generates a filter that will match against the
 // name of the charm or bundle.
 func nameFilter(value string) elasticsearch.Filter {
-	// TODO(mhilton) implement wildcards as in http://tinyurl.com/k46xexe
-	return elasticsearch.RegexpFilter{
-		Field:  "URL",
-		Regexp: `cs:(\~[^/]*/)?[^/]*/` + elasticsearch.EscapeRegexp(value) + "-[0-9]+",
+	return elasticsearch.QueryFilter{
+		Query: elasticsearch.MatchQuery{
+			Field: "Name",
+			Query: value,
+			Type:  "phrase",
+		},
 	}
 }
 
 // ownerFilter generates a filter that will match against the
 // owner taken from the URL.
 func ownerFilter(value string) elasticsearch.Filter {
-	var re string
-	if value == "" {
-		re = `cs:[^\~].*`
-	} else {
-		re = `cs:\~` + elasticsearch.EscapeRegexp(value) + "/.*"
-	}
-	return elasticsearch.RegexpFilter{
-		Field:  "URL",
-		Regexp: re,
+	return elasticsearch.QueryFilter{
+		Query: elasticsearch.MatchQuery{
+			Field: "User",
+			Query: value,
+			Type:  "phrase",
+		},
 	}
 }
 
 // seriesFilter generates a filter that will match against the
 // series taken from the URL.
 func seriesFilter(value string) elasticsearch.Filter {
-	return elasticsearch.RegexpFilter{
-		Field:  "URL",
-		Regexp: `cs:(\~[^/]*/)?` + elasticsearch.EscapeRegexp(value) + "/.*-[0-9]+",
+	return elasticsearch.QueryFilter{
+		Query: elasticsearch.MatchQuery{
+			Field: "Series",
+			Query: value,
+			Type:  "phrase",
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/531/

internal/charmstore/search.go
```diff
@@ -145,6 +145,8 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
 func queryFields(sp SearchParams) map[string]float64 {
 	fields := map[string]float64{
 		"URL.ngrams":              8,
+		"Name":                    8,
+		"Series":                  8,
```
mitechie: is the user not available here? We had owner=?

internal/elasticsearch/definitions/entity.json
```diff
@@ -1,7 +1,25 @@
...
+        "omit_norms" : true,
+        "index_options" : "docs"
+      },
+      "Series" : {
+        "type" : "string",
+        "index" : "not_analyzed",
+        "omit_norms" : true,
+        "index_options" : "docs"
+      },
+      "User" : {
```
mitechie, line 17: so this is user, will it still map to owner in the filter? Does this mean a generic search for text=rharding might find my charms?

mitechie:
:+1: with a couple of questions. Did not QA.

internal/charmstore/search.go
mhilton: I'm not sure this is right, you're boosting names and series rather a lot here. If you match trusty say then you'll get 8 boost from the URL and another 8 from the series despite it being essentially the same bit of information.

internal/charmstore/search.go
```diff
@@ -266,34 +268,36 @@ func descriptionFilter(value string) elasticsearch.Filter {
 // nameFilter generates a filter that will match against the
 // name of the charm or bundle.
 func nameFilter(value string) elasticsearch.Filter {
-	// TODO(mhilton) implement wildcards as in http://tinyurl.com/k46xexe
-	return elasticsearch.RegexpFilter{
-		Field:  "URL",
-		Regexp: `cs:(\~[^/]*/)?[^/]*/` + elasticsearch.EscapeRegexp(value) + "-[0-9]+",
+	return elasticsearch.QueryFilter{
+		Query: elasticsearch.MatchQuery{
```
mhilton, line 270: You probably want a TermFilter here as we only really care about matching exact names.

internal/charmstore/search.go
```diff
@@ -266,34 +268,36 @@ func descriptionFilter(value string) elasticsearch.Filter {
...
-	var re string
-	if value == "" {
-		re = `cs:[^\~].*`
-	} else {
-		re = `cs:\~` + elasticsearch.EscapeRegexp(value) + "/.*"
-	}
-	return elasticsearch.RegexpFilter{
-		Field:  "URL",
-		Regexp: re,
+	return elasticsearch.QueryFilter{
```
mhilton, line 281: TermFilter here too.

internal/charmstore/search.go
```diff
@@ -266,34 +268,36 @@ func descriptionFilter(value string) elasticsearch.Filter {
...
 	}
 }
 
 // seriesFilter generates a filter that will match against the
 // series taken from the URL.
 func seriesFilter(value string) elasticsearch.Filter {
-	return elasticsearch.RegexpFilter{
-		Field:  "URL",
-		Regexp: `cs:(\~[^/]*/)?` + elasticsearch.EscapeRegexp(value) + "/.*-[0-9]+",
+	return elasticsearch.QueryFilter{
```
mhilton, line 293: And a TermFilter here

internal/charmstore/search.go
```diff
@@ -266,34 +268,36 @@ func descriptionFilter(value string) elasticsearch.Filter {
...
-		re = `cs:[^\~].*`
-	} else {
-		re = `cs:\~` + elasticsearch.EscapeRegexp(value) + "/.*"
-	}
-	return elasticsearch.RegexpFilter{
-		Field:  "URL",
-		Regexp: re,
+	return elasticsearch.QueryFilter{
+		Query: elasticsearch.MatchQuery{
+			Field: "User",
```
mhilton, line 283: Why couldn't we have been consistent and called this filed Owner?

internal/elasticsearch/definitions/entity.json
mhilton, line 17: @mitechie it would find them the same way it does now, using the URL ngrams.

internal/elasticsearch/definitions/entity.json
```diff
@@ -1,7 +1,25 @@
 {
   "entity" : {
     "dynamic" : "false",
-    "properties" : {
+    "properties" : {
```
mhilton: ooh, trailing white-space.

mhilton:
:+1: But please take a look to see if Termfilters can be used, They'll be faster.

internal/charmstore/search.go
urosj: Removed. Just wanted to see them work :)

internal/elasticsearch/definitions/entity.json
urosj, line 17: It would only find ~user/ part of the URL. So, if there is ~rharding/trusty, then filtering with "&owner=rhading" will return all your charms. But in some next PR, we should have better search over extra info, which we don't use at all now. And that would allow us to search by real names of contributors, etc.

internal/charmstore/search.go
urosj, line 270: If I add termFilters, tests fail.

internal/charmstore/search.go
urosj, line 283: It is the user part of the URL. Better question is, why didn't we use user as filter instead. Suggesting it as improvement for next PR.

internal/elasticsearch/definitions/entity.json
urosj: Done :)

urosj:
:shipit:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/532/

PR #216, opened 2014-11-12
Adding sorting to search results.

mhilton:


mhilton:
To QA:
1. Build an elasticsearch database using essync
2. Run charmd
3. Perform a number of searches on charmd using the new sort= parameter (e.g. `sort=name`, `sort=name+desc`, `sort=owner&sort=name+desc`, `sort=series+desc&sort=owner+asc&sort=name`)

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/534/

internal/charmstore/search.go
```diff
@@ -121,6 +121,22 @@ type SearchParams struct {
 	Include []string
 	// Start the the returned items at a specific offset
 	Skip int
+	// Sort the returned items
+	Sort []SortParam
+}
+
+//SortDirection defines the direction in which a field should be sorted.
```
rogpeppe: need a space after the //

internal/charmstore/search.go
```diff
@@ -121,6 +121,22 @@ type SearchParams struct {
...
+type SortDirection int
+
+const (
+	SortAscending = iota
+	SortDescending
+)
+
+// SortParam represents a field and direction on which results should be sorted.
+type SortParam struct {
+	Field     string
```
rogpeppe: It's not clear what kind of thing this field refers to - is it the name of a field in the mongodoc entity struct, for example, or a field as stored in elastic search, or a field as stored in mongo?

Looking at the implementation, it seems to be none of the above, but the name as specified in the API. Perhaps it might be best to enumerate the allowed fields, and translate to those values inside v4. That way it becomes more obvious which fields are currently supported.

mhilton: Yes, they are designed to be the same as the filter name that filters on the same thing.

internal/v4/search.go
```diff
@@ -18,7 +19,7 @@ import (
 
 const maxConcurrency = 20
 
-// GET search[?text=text][&autocomplete=1][&filter=value…][&limit=limit][&include=meta]
+// GET search[?text=text][&autocomplete=1][&filter=value…][&limit=limit][&include=meta][&skip=count][&sort=field[+dir]]
```
rogpeppe, line 21: How about instead of "sort=name asc" we have sort=-name for descending and sort=+name for ascending (with no + or -, the default would be ascending)? 
That's nice and compact, and makes a nice obvious syntax for multiple-level sorting at some point in the future, e.g. sort=series,-publishTime

internal/charmstore/search.go
rogpeppe: I'm not keen on the fact that we have to enumerate all of the possible names in both places (v4 and charmstore).

Perhaps a nice approach might be to add a function to do the translation:

```
func ParseSortField(field string) (SortParam, error)
```

Then we can avoid enumerating the fields specifically inside the API endpoint.
And the SortParam members could be unexported too (it could just hold an elasticsearch.Sort member).

internal/charmstore/search.go
```diff
@@ -366,3 +387,23 @@ func typeFilter(value string) elasticsearch.Filter {
...
+func createSort(s SortParam) elasticsearch.Sort {
+	var sort elasticsearch.Sort
+	switch s.Field {
+	case "name":
+		sort.Field = "Name"
+	case "owner":
+		sort.Field = "User"
+	case "series":
+		sort.Field = "Series"
+	}
```
rogpeppe, line 422: It's not great that when the field is unknown, we fail silently.
See my ParseSortField suggestion above.

internal/v4/search.go
rogpeppe, line 21: Actually probably better without the +, as it may be awkward in URLs.

mhilton, line 21: Nice idea, I'm not sure about + because they tend to get un-escaped into spaces by URL handlers and I wouldn't want to make a client escape an actual '+'. We already have an, evidently non-obvious, syntax for multilevel sorting you just add additional sort fields.

internal/charmstore/search.go
mhilton, line 422: We only fail silently internally. v4/search.go will reject anything that's not the three names it knows about so you get an error if you sak for anything else.

rogpeppe:
Looks really good in general, but I have made a few comments and suggestions above. What do you think?

internal/charmstore/search.go
```diff
@@ -121,6 +121,22 @@ type SearchParams struct {
 	Include []string
 	// Start the the returned items at a specific offset
 	Skip int
+	// Sort the returned items
```
bac: Please use punctuation.  Everywhere.

bac:
Had a quick look.  Will look again after you address Roger's concerns.

internal/charmstore/search.go
mhilton: I've taken the approach of translating at the v4 layer into an esdoc field name and the just using it in the search.

internal/v4/search.go
mhilton, line 21: I've done this using the sort=name for ascending and sort=-name for descending approach.

internal/charmstore/search.go
mhilton, line 422: This part now uses what is supplied. The v4 layer should translate to the correct field name.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/535/

mhilton:
The QA instructions are slightly different after changing the API. Now for step 3 ascending searches are specified by just the search field and descending by `-` followed by the search name. sort orders may also be separated by a `,`. Valid sort queries are:
- `sort=name`
- `sort=-series`
- `sort=name&sort=-owner`
- `sort=-series,name`

internal/v4/search.go
```diff
@@ -18,7 +19,13 @@ import (
 
 const maxConcurrency = 20
 
-// GET search[?text=text][&autocomplete=1][&filter=value…][&limit=limit][&include=meta]
+var fields = map[string]string{
```
rogpeppe: I think this should be in the charmstore package, as the values here are internal details of the stored ES document - knowledge of the ES document details should be kept private to charmstore IMO, so that when it changes we don't need to look for this code too.

rogpeppe:
LGTM with the above comment addressed.

bac:
Code is :+1:

mhilton:
:shipit:

jujugui:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/536/

bac:
QA :+1: 

Thanks Martin!

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/538/

PR #219, opened 2014-11-13
Stop adding charms from deprecated series into search

mhilton:


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/542/

mhilton:
## QA Instructions

This one, again requires a complete replacement of the elasticsearch database. These instructions assume that you have a charmstore with a reasonably full mongodb database.
1. Recreate elasticsearch database.
   
   cd ${GOPATH}/src/github.com/juju/charmstore
   make install
   http DELETE :9200/charmstore
   essync -settings=internal/elasticsearch/definitions/index.json -mapping=internal/elasticsearch/definitions/entity.json &lt;path/to/config.yaml&gt;
2. Check some charms from deprecated series exist.
   
   charmd &lt;path/to/config.yaml&gt; &amp;
   http :8080/v4/wordpress/expand-id

This will return the set of ingested wordpress charms, look out for ones in deprecated series. e.g.

```
[
    {
        "Id": "cs:oneiric/wordpress-0"
    }, 
    {
        "Id": "cs:precise/wordpress-0"
    }, 
    {
        "Id": "cs:precise/wordpress-1"
    },
...
```
1. Perform search.
   
   charmd &lt;path/to/config.yaml&gt; &amp;
   http ":8080/v4/search?name=wordpress&owner="

Notice that the deprecated series do not show up in the search resutls. e.g.

```
{
    "Results": [
        {
            "Id": "cs:precise/wordpress-26"
        }
    ],
...
```

internal/elasticsearch/elasticsearch.go
```diff
@@ -212,9 +212,13 @@ func (db *Database) do(method, url string, body, v interface{}) error {
 		var eserr ElasticSearchError
 		if err = json.Unmarshal(b, &eserr); err != nil {
 			log.Debugf("*** %s", err)
-			eserr.Err = fmt.Sprintf(`elasticsearch status "%s"`, resp.Status)
+		}
+		if eserr.Status == 0 {
```
rogpeppe, line 216: What's going on here?

rogpeppe:
LGTM with one small query.

internal/elasticsearch/elasticsearch.go
mhilton, line 216: elasticsearch returns a nice standard completely not documented json structure if an error occurs, except in the case I happen to care about (Getting a document that isn't present). What this does is unmarshal that structure, if it is present, but if not it tries to fill in the details as best it can.

frankban, line 216: Could you add the above as a comment in the code?

frankban:
:+1: thank you!

mhilton:
:shipit:

jujugui:

urosj:
PS: QA was done. Deleted the index, run essync and then ingestion. Searched for charms with old series using /expand-id endpoint. Check if they showed in ES. Esp. in the early stages of ingestion, when only the old series existed and expand-id showed only old series, but they could not be found in ES.

PR #228, opened 2014-11-20
Implement the charm store logging API.

frankban:
This will allow storing/retrieving ingestion logs.
To activate the functionality, a client branch is also
required.


internal/charmstore/store.go
```diff
@@ -555,6 +558,20 @@ func bundleCharms(data *charm.BundleData) ([]*charm.Reference, error) {
 	return urls, nil
 }
 
+// AddIngestionLog adds an ingestion log to the database.
+func (s *Store) AddIngestionLog(message []byte, urls []*charm.Reference, level mongodoc.IngestionLogLevel) error {
```
rogpeppe: Why use a []byte for the message rather than a string?

params/params.go
```diff
@@ -149,3 +150,35 @@ type IdRevisionResponse struct {
 
 // BzrDigestKey is the extra-info key used to store the Bazaar digest
 const BzrDigestKey = "bzr-digest"
+
+type IngestionLogLevel string
+
+const (
+	IngestionInfo  IngestionLogLevel = "info"
```
rogpeppe: Given that this is probably going to be a relatively high-volume collection, perhaps it would be better to use small integer constants here.

internal/charmstore/store.go
```diff
@@ -580,12 +597,18 @@ func (s StoreDatabase) Entities() *mgo.Collection {
...
+	return s.C("ingestionlogs")
+}
+
 // allCollections holds for each collection used by the charm store a
 // function returns that collection.
 var allCollections = []func(StoreDatabase) *mgo.Collection{
 	StoreDatabase.StatCounters,
 	StoreDatabase.StatTokens,
 	StoreDatabase.Entities,
+	StoreDatabase.IngestionLogs,
```
urosj: Is there a plan to store more then just errors? IngestionLogs suggest that we store info and warnings there as well.

frankban: The plan is to store info like "ingestion started"/"ingestion completed".

internal/charmstore/store.go
frankban: The intention is to make the message a JSON data.

internal/v4/ingestion_test.go
```diff
@@ -0,0 +1,417 @@
...
+			charm.MustParseReference("trusty/django"),
+			charm.MustParseReference("utopic/rails"),
+		},
+	}
+	b, err := json.Marshal(log)
+	if err != nil {
+		panic(err)
+	}
+	return b
+}
```
urosj: Wow! Nice coverage!

params/params.go
```diff
@@ -149,3 +150,42 @@ type IdRevisionResponse struct {
...
+	IngestionError IngestionLogLevel = "error"
+)
+
+// IngestionLog holds the representation of a log message sent while
+// ingesting/uploading charms or bundles.
+type IngestionLog struct {
+	// Message holds the log message.
+	Message []byte
```
urosj: This is the first time we use []byte in CS. Any particular reason?

urosj:
:+1: Some questions, nothing blocking.

internal/v4/ingestion_test.go
```diff
@@ -0,0 +1,417 @@
...
+			Code:    params.ErrUnauthorized,
+		},
+	})
+}
+
+func makeByteLog(message json.RawMessage, urls []*charm.Reference, level params.IngestionLogLevel) []byte {
+	log := &params.IngestionLog{
+		Message: message,
+		Level:   level,
+		URLs: []*charm.Reference{
```
rogpeppe: shouldn't this be using urls ?

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/560/

params/params.go
rogpeppe: Agreed; this seems wrong, as this will be travelling on the wire.
I think we probably want *json.RawMessage here instead.

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/561/

internal/mongodoc/doc.go
```diff
@@ -80,6 +80,30 @@ type Entity struct {
...
+
+	// Time holds the time of the log.
+	Time time.Time
+}
+
+// IngestionLogLevel holds the log level associated with an ingestion log.
+type IngestionLogLevel int
+
+const (
+	IngestionInfo IngestionLogLevel = iota
```
rogpeppe: I'd suggest:
    const (
         _ IngestionLogLevel = iota
         IngestionInfo
         IngestionError
     )

so that it's obvious when a level is accidentally left unset.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/562/

internal/mongodoc/doc.go
```diff
@@ -80,6 +80,30 @@ type Entity struct {
 	Contents map[FileId]ZipFile `json:",omitempty" bson:",omitempty"`
 }
 
+// IngestionLog holds the in-database representation of a log message sent
+// while ingesting/uploading charms or bundles.
+type IngestionLog struct {
+	// Message holds the log message.
+	Message []byte
```
rogpeppe: If the intention is to store JSON here, then let's make that obvious, and call it Data or Attributes or something.

internal/v4/ingestion.go
```diff
@@ -0,0 +1,151 @@
...
+func (h *Handler) getIngestionLogs(w http.ResponseWriter, req *http.Request) error {
+	w.Header().Set("content-type", "application/json")
+	encoder := json.NewEncoder(w)
+
+	// Retrieve values from the query string.
+	limit, err := getPositiveIntValue(req.Form, "limit", 1000)
+	if err != nil {
+		return badRequestf(err, "invalid query")
+	}
+	offset, err := getPositiveIntValue(req.Form, "offset", 0)
```
rogpeppe: If the default is 0, why don't we allow users to specify offset=0 ?

internal/v4/ingestion.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (h *Handler) getIngestionLogs(w http.ResponseWriter, req *http.Request) error {
+	w.Header().Set("content-type", "application/json")
+	encoder := json.NewEncoder(w)
+
+	// Retrieve values from the query string.
+	limit, err := getPositiveIntValue(req.Form, "limit", 1000)
+	if err != nil {
+		return badRequestf(err, "invalid query")
```
rogpeppe: "invalid limit value" ?

internal/v4/ingestion.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	offset, err := getPositiveIntValue(req.Form, "offset", 0)
+	if err != nil {
+		return badRequestf(err, "invalid query")
```
rogpeppe: "invalid offset value" ?

internal/v4/ingestion.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	id := req.Form.Get("id")
+	levelStr := req.Form.Get("level")
+
+	// Build the Mongo query.
+	query := bson.D{}
```
rogpeppe: query := make(bson.D, 0, 2)

to save a little garbage?

internal/v4/ingestion.go
```diff
@@ -0,0 +1,151 @@
...
+			return errgo.Notef(err, "cannot marshal ingestion log")
+		}
+	}
+	if err := iter.Close(); err != nil {
+		return errgo.Notef(err, "cannot retrieve ingestion logs")
+	}
+	return nil
+}
+
+func (h *Handler) postIngestionLog(w http.ResponseWriter, req *http.Request) error {
```
rogpeppe: I wonder if we should allow the client to add several log messages here in one call, making it reasonable for a client to buffer log messages and send a bunch at once.

internal/v4/ingestion.go
```diff
@@ -0,0 +1,151 @@
...
+// value is not set. An error is returned if the value is not valid.
+func getPositiveIntValue(flags url.Values, name string, defaultValue int) (int, error) {
+	if strValue := flags.Get(name); strValue != "" {
+		value, err := strconv.Atoi(strValue)
+		if err != nil || value <= 0 {
+			return 0, errgo.New("invalid value for " + name)
+		}
+		return value, nil
+	}
+	return defaultValue, nil
```
rogpeppe: Perhaps put the test for this first, to save the indent level in the reset of the function?

internal/charmstore/store.go
```diff
@@ -555,6 +558,20 @@ func bundleCharms(data *charm.BundleData) ([]*charm.Reference, error) {
 	return urls, nil
 }
 
+// AddIngestionLog adds an ingestion log to the database.
+func (s *Store) AddIngestionLog(message []byte, urls []*charm.Reference, level mongodoc.IngestionLogLevel) error {
+	log := &mongodoc.IngestionLog{
+		Message: message,
+		Level:   level,
+		URLs:    urls,
```
rogpeppe: Do we want to allow messages to be added with no URLs ?

rogpeppe:
Looks nice, thanks, with a few comments, suggestions and queries above.

I will say one thing again though: if we globally replace the string "Ingestion"
with the word "EntityEvent" in this PR, what we actually have here is
a general mechanism for associating events with entities - there's almost
nothing ingestion specific about it. Given that it's a significant amount of code,
and that clients will inevitably start to rely on it, however much
we say they should not, I would again suggest that we remove
the word "ingestion" from this PR and use it as the general mechanism that it
is.

params/params.go
frankban: Done.

internal/mongodoc/doc.go
frankban: Done.

internal/mongodoc/doc.go
frankban: Done.

internal/v4/ingestion.go
frankban: Done.

internal/v4/ingestion.go
frankban: Done.

internal/v4/ingestion.go
frankban: Done.

internal/v4/ingestion.go
frankban: Done.

internal/v4/ingestion.go
frankban: This function is no longer used.

internal/charmstore/store.go
frankban: Yes, I think we do, for example the ingestion started/completed logs.

internal/v4/ingestion.go
frankban: I am not sure about that. Usually a log is sent asap as a single entity.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/563/

internal/v4/log.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Add the log to the database.
+	if err := h.store.AddLog(log.Data, logLevel, logType, log.URLs); err != nil {
+		return errgo.Notef(err, "cannot add log")
+	}
+	return nil
+}
+
+var (
+	// mongodocLogLevels maps internal mongodoc log levels to API ones.
```
urosj, line 165: Are these really needed?

internal/charmstore/store.go
```diff
@@ -555,6 +559,25 @@ func bundleCharms(data *charm.BundleData) ([]*charm.Reference, error) {
 	return urls, nil
 }
 
+// AddLog adds a log message to the database.
+func (s *Store) AddLog(data *json.RawMessage, logLevel mongodoc.LogLevel, logType mongodoc.LogType, urls []*charm.Reference) error {
+	b, err := json.Marshal(data)
```
rogpeppe, line 565: Alternatively, if we're prepared to trust the no-one is actually going to fabricate an invalid json.RawMessage (which I think isn't unreasonable), we could just use *data below.

internal/mongodoc/doc.go
```diff
@@ -80,6 +80,53 @@ type Entity struct {
 	Contents map[FileId]ZipFile `json:",omitempty" bson:",omitempty"`
 }
 
+// Log holds the in-database representation of a log message sent to the charm
+// store.
+type Log struct {
+	// Data holds the log message.
```
rogpeppe: // Data holds the JSON-encoded log message.

?

internal/v4/log.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		query = append(query, bson.DocElem{"type", logType})
+	}
+
+	// Retrieve the logs.
+	var log mongodoc.Log
+	iter := h.store.DB.Logs().Find(query).Sort("-_id").Skip(offset).Limit(limit).Iter()
+	for iter.Next(&log) {
+		var data json.RawMessage
+		if err := json.Unmarshal(log.Data, &data); err != nil {
```
rogpeppe: No need to unmarshal here - json will make sure it's all valid when marshaling anyway.

```
logResponse = &params.LogResponse{
    Data: (*json.RawMessage)(&log.Data),
    ... etc
```

internal/v4/log.go
```diff
@@ -0,0 +1,171 @@
...
+		return h.getLogs(w, req)
+	case "POST":
+		return h.postLog(w, req)
+	}
+	return errgo.WithCausef(nil, params.ErrMethodNotAllowed, "%s method not allowed", req.Method)
+}
+
+func (h *Handler) getLogs(w http.ResponseWriter, req *http.Request) error {
+	w.Header().Set("content-type", "application/json")
+	encoder := json.NewEncoder(w)
```
rogpeppe, line 40: This is pretty unusual, and I'm not sure that JS clients will be able to read a sequence of JSON messages like this.

However I like the idea of streaming the messages to avoid potentially using up loads of memory.

One alternative could be to manually write the first "[", subsequent comma separators, and the final "]".

That way we can still stream everything, but the client will see the whole thing as a single JSON array.

internal/v4/log.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (h *Handler) postLog(w http.ResponseWriter, req *http.Request) error {
+	// Check the request content type.
+	if ctype := req.Header.Get("Content-Type"); ctype != "application/json" {
+		return badRequestf(nil, "unexpected Content-Type %q; expected 'application/json'", ctype)
+	}
+
+	// Unmarshal the request body.
+	var log params.Log
```
rogpeppe: As I mentioned in my previous review, I'd prefer it if this was a []params.Log so clients could send log messages in bulk.

internal/v4/log.go
```diff
@@ -0,0 +1,171 @@
...
+	// Add the log to the database.
+	if err := h.store.AddLog(log.Data, logLevel, logType, log.URLs); err != nil {
+		return errgo.Notef(err, "cannot add log")
+	}
+	return nil
+}
+
+var (
+	// mongodocLogLevels maps internal mongodoc log levels to API ones.
+	mongodocLogLevels = map[mongodoc.LogLevel]params.LogLevel{
```
rogpeppe, line 166: Slightly more efficient, because we know these are small
integers, could be:

```
 mongodocLogLevels = []params.LogLevel{
    mongodoc.InfoLevel:    params.InfoLevel,
    mongodoc.WarningLevel: params.WarningLevel,
    mongodoc.ErrorLevel:   params.ErrorLevel,
 }
```

You'd then check the array bounds rather than the map
get result. Same thing for mongodocLogTypes.

YMMV.

internal/v4/log.go
```diff
@@ -0,0 +1,171 @@
...
+	mongodocLogTypes = map[mongodoc.LogType]params.LogType{
+		mongodoc.IngestionType: params.IngestionType,
+	}
+	// paramsLogTypes maps API params log types to internal mongodoc ones.
+	paramsLogTypes = map[params.LogType]mongodoc.LogType{
+		params.IngestionType: mongodoc.IngestionType,
+	}
+)
+
+// intValue checks that the given string value is a number greater than the
```
rogpeppe: greater than or equal to.

params/params.go
```diff
@@ -149,3 +150,55 @@ type IdRevisionResponse struct {
 
 // BzrDigestKey is the extra-info key used to store the Bazaar digest
 const BzrDigestKey = "bzr-digest"
+
+// Log holds the representation of a log message.
+// This is used by clients to store log events in the charm store.
+type Log struct {
+	// Data holds the log message as a JSON.
```
rogpeppe: as a JSON-encoded value.

params/params.go
```diff
@@ -149,3 +150,55 @@ type IdRevisionResponse struct {
...
+	Data json.RawMessage
+
+	// Level holds the log level as a string.
+	Level LogLevel
+
+	// Type holds the log type as a string.
+	Type LogType
+
+	// URLs holds a slice of entity URLs associated with the log message.
+	URLs []*charm.Reference
```
rogpeppe: omitempty ?

rogpeppe:
LGTM modulo a few comments and suggestions above. Thanks very much for making the changes. I am much happier with the PR in this form.

internal/v4/log.go
```diff
@@ -0,0 +1,171 @@
...
+func (h *Handler) getLogs(w http.ResponseWriter, req *http.Request) error {
+	w.Header().Set("content-type", "application/json")
+	encoder := json.NewEncoder(w)
+
+	// Retrieve values from the query string.
+	limit, err := intValue(req.Form.Get("limit"), 1, 1000)
+	if err != nil {
+		return badRequestf(err, "invalid limit value")
+	}
+	offset, err := intValue(req.Form.Get("offset"), 0, 0)
```
jrwren: In the search API we settled on "skip" instead of offset. We may want to be consistent here.

jrwren:
LGTM with one comment about skip v. offset.

A more general question: why do this rather than run logstash and kibana and put the logs into logstash?

internal/charmstore/store.go
frankban, line 565: The more explicit approach seems good to me, and not that expensive.

internal/mongodoc/doc.go
frankban: Better, done.

internal/v4/log.go
frankban: Done.

internal/v4/log.go
frankban: Done.

params/params.go
frankban: Done.

params/params.go
frankban: Yes, done.

internal/v4/log.go
frankban: We discussed about this on IRC, I am not sure about this change, but I was not able to convince you otherwise. So done.

internal/v4/log.go
frankban, line 40: Done.

internal/v4/log.go
frankban: Done.

internal/v4/log.go
frankban, line 166: Given this is internal, I'll add a TODO for this to be fixed later, if we want to.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/565/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/566/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/567/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/569/

frankban:
Thank a lot for the great reviews and QA!
:shipit:

jujugui:

PR #231, opened 2014-11-21
Automatically initialise elasticsearch.

mhilton:
Charmstore will now automatically initialise elasticsearch when starting up.
This ensures that elasticsearch will always have the correct index and mapping
configuration required for correct operation.


cmd/essync/main.go
```diff
@@ -22,10 +21,10 @@ import (
 var logger = loggo.GetLogger("essync")
 
 var (
-	index         = flag.String("index", "charmstore", "Name of index to populate.")
+	index         = flag.String("index", "cs", "Name of index to populate.")
 	loggingConfig = flag.String("logging-config", "", "specify log levels for modules e.g. <root>=TRACE")
-	mapping       = flag.String("mapping", "", "File to use to configure the entity mapping.")
-	settings      = flag.String("settings", "", "File to use to configure the index.")
+	mapping       = flag.String("mapping", "", "No longer used.")
+	settings      = flag.String("settings", "", "No longer used.")
```
urosj, line 27: Hm. Now We can't try out "new" mappings without making them part of code? Not sure if I'm fond of this limitation.

mhilton:
# QA Instructions
1. Delete all existing elasticsearch indexes (not strictly necesssary, but it's a good way to know charmd isn't using existing indexes)

`curl -XDELETE http://localhost:9200/*`
1. Start charmd with an ingested mongodb,
2. After a few seconds perform a search.

`curl http://localhost:8080/v4/search?text=mongo`

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/559/

urosj:
@mhilton why after a few seconds? Is essync performed?

cmd/essync/main.go
rogpeppe, line 27: I think this is great. The less flags the better. Especially as these mappings are tied intimately to the code, so only useful to someone that intimately involved in development of the charm store.

If we want to be able to try out new mappings, you'll need to re-run the charmstore, and it's easy to change them in the code, I think.

cmd/essync/main.go
```diff
@@ -22,10 +21,10 @@ import (
 var logger = loggo.GetLogger("essync")
 
 var (
-	index         = flag.String("index", "charmstore", "Name of index to populate.")
+	index         = flag.String("index", "cs", "Name of index to populate.")
```
rogpeppe, line 24: Why the index name change?

cmd/essync/main.go
```diff
@@ -66,58 +65,12 @@ func populate(confPath string) error {
 	}
 	defer session.Close()
 	db := session.DB("juju")
-	store, err := charmstore.NewStore(db, &charmstore.StoreElasticSearch{es.Index(*index)})
+	charmstore.ESSearchIndex = *index
+	charmstore.ESWriteIndex = charmstore.ESSearchIndex + "-write"
+	// Syncing will happen automatically when creating the store
```
bac: .

mhilton:
@urosj Yes the reasoning is thus. charmd cannot know if any particular index already has the correct configuration, so it has to create a new one. If it's creating a new one it has to populate it otherwise searches won't return results.

cmd/essync/main.go
mhilton, line 24: This version uses aliases to make it possible to move between indexes on upgrade in a sensible manner. I changed the index name so that it doesn't clash with any existing charmstore index.

internal/charmstore/elasticsearch.go
```diff
@@ -1,3 +1,52 @@
...
+package charmstore
+
+import "encoding/json"
+
+var (
+	esIndex   interface{}
+	esMapping interface{}
+)
+
+func init() {
```
rogpeppe: or:

```
func mustJSON(data string) interface{} {
    var x interface{}
    if err := json.Unmarshal([]byte(data), &x); err != nil {
        panic(err)
    }
}
var esIndex = mustJSON(`
{ .... }`)

var esMapping = mustJSON(`
{ ....}`)
```

?

To save a bit of runtime space, we could keep them as *json.RawMessage rather than interface{} too, although there may be other reasons why that won't work.

cmd/essync/main.go
```diff
@@ -66,58 +65,12 @@ func populate(confPath string) error {
 	}
 	defer session.Close()
 	db := session.DB("juju")
-	store, err := charmstore.NewStore(db, &charmstore.StoreElasticSearch{es.Index(*index)})
+	charmstore.ESSearchIndex = *index
```
rogpeppe: This mutable global makes me uncomfortable.
Can't we make it a parameter to the charm store instead?

internal/charmstore/search.go
```diff
@@ -54,30 +63,43 @@ var deprecatedSeries = map[string]bool{
...
 		Series: entity.URL.Series,
 	}
-	_, err := ses.PutDocumentVersion(typeName, ses.getID(entity), int64(entity.URL.Revision), doc)
+	_, err := ses.PutDocumentVersionWithType(
+		ESWriteIndex,
+		typeName,
+		ses.getID(entity.URL),
+		int64(entity.URL.Revision),
+		elasticsearch.ExternalGTE,
+		doc)
```
rogpeppe, line 97: This is getting towards the stage where a parameter struct might be appropriate...

internal/charmstore/search.go
```diff
@@ -113,20 +135,64 @@ func (ses *StoreElasticSearch) search(sp SearchParams) (SearchResult, error) {
...
+	l := lock.NewLock(ses.Database, "essync")
+	locked, err := l.Lock(5 * time.Minute)
+	if err != nil {
+		return errgo.Notef(err, "error aquiring lock")
+	}
+	if !locked {
+		// There is another essync in process somewhere else, let that carry on
+		return nil
+	}
+	defer l.Unlock() // Make sure we always unlock
```
rogpeppe: This comment is somewhat redundant.

internal/charmstore/store.go
```diff
@@ -615,3 +623,13 @@ func (r *readerAtSeeker) ReadAt(buf []byte, p int64) (int, error) {
 func ReaderAtSeeker(r io.ReadSeeker) io.ReaderAt {
 	return &readerAtSeeker{r}
 }
+
+// Search searches the store for the given SearchParams.
+// It returns a slice a SearchResult containing the results of the search.
```
rogpeppe: s/a slice//

internal/charmstore/store.go
```diff
@@ -615,3 +623,13 @@ func (r *readerAtSeeker) ReadAt(buf []byte, p int64) (int, error) {
...
+func (store *Store) Search(sp SearchParams) (SearchResult, error) {
+	results, err := store.ES.search(sp)
```
rogpeppe: s/results/result/
?

internal/elasticsearch/elasticsearch.go
```diff
@@ -25,6 +25,12 @@ import (
 	"gopkg.in/errgo.v1"
 )
 
+const (
```
rogpeppe, line 28: could we have a doc comment on these please?

internal/elasticsearch/elasticsearch.go
```diff
@@ -40,6 +46,64 @@ type Database struct {
...
+type Document struct {
+	Found   bool            `json:"found"`
+	Id      string          `json:"_id"`
+	Index   string          `json:"_index"`
+	Type    string          `json:"_type"`
+	Version int64           `json:"_version"`
+	Source  json.RawMessage `json:"_source"`
+}
+
+// Alias creates or updates an alias. An alias a is created,
```
rogpeppe: s/an alias/an index alias/ ?

internal/elasticsearch/elasticsearch.go
```diff
@@ -40,6 +46,64 @@ type Database struct {
...
+// or modified if it already exists, to point to i. See
+// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/indices-aliases.html#indices-aliases
+// for further details.
+func (db *Database) Alias(i, a string) error {
+	indexes, err := db.ListIndexesForAlias(a)
+	if err != nil {
+		return errgo.Notef(err, "error retreiving current aliases")
```
rogpeppe: s/retreiving/retrieving/

rogpeppe: actually, I'd usually say "cannot retrieve current aliases".

internal/elasticsearch/elasticsearch.go
```diff
@@ -40,6 +46,64 @@ type Database struct {
...
+
+// Create document attempts to create a new document at index/type_/id with the
+// contents in doc. If the document already exists then CreateDocument will return false,
+// otherwise a new document is created and CreateDocument returns true.
+// CreateDocument return a non-nil error if any other error occurs.
+// See http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/create-doc.html#create-doc
+// for further details.
+func (db *Database) CreateDocument(index, type_, id string, doc interface{}) (bool, error) {
+	if err := db.put(db.url(index, type_, id, "_create"), doc, nil); err != nil {
+		if eserr, ok := err.(ElasticSearchError); ok && eserr.Status == http.StatusConflict {
```
rogpeppe: FWIW it's conventional to make errors returned as pointer types, so that it's obvious what to check for. So it would conventially be *ElasticSearchError.

internal/elasticsearch/lock/doc.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+// Package lock provides a form of co-operative mutex locking on an elasticsearch
+// database. The locking is based on the existance of documents with particualr names
```
rogpeppe: s/particualr/particular/

urosj:
:+1: Thanks for this improvement!

PS: this does not include QA.

internal/elasticsearch/lock/lock.go
```diff
@@ -0,0 +1,86 @@
...
+	locked   bool
+}
+
+// NewLock creates a Lock with the name n on an elasticsearch database db.
+func NewLock(db *es.Database, n string) *Lock {
+	return &Lock{Database: db, Name: n}
+}
+
+// Lock attempts to acquire the for the maximum duration t. If the lock is acquired
+// Lock will return true, it will return false if it cannot acquire the lock. If there is
```
rogpeppe: Go convention would have:
// Lock reports whether the lock has been acquired.

internal/elasticsearch/lock/lock.go
```diff
@@ -0,0 +1,86 @@
...
+// any error communicating the elasticsearch database then the error will be returned.
+//
+// The timeout t should be set as a maximum resonable amount of time that the client
```
rogpeppe: s/resonable/reasonable/

internal/elasticsearch/lock/lock.go
```diff
@@ -0,0 +1,86 @@
...
+// will require the lock. After the timeout the lock will be automatically released and can
```
rogpeppe: s/lock./lock for./ ?

internal/elasticsearch/lock/lock.go
```diff
@@ -0,0 +1,86 @@
...
+// be acquired by other clients.
+func (l *Lock) Lock(t time.Duration) (bool, error) {
+	lck := lock{Time: time.Now(), Duration: t}
```
rogpeppe: lockDoc ?

internal/elasticsearch/elasticsearch.go
```diff
@@ -40,6 +46,64 @@ type Database struct {
...
+	}
+	actions.Actions = append(actions.Actions, action{Add: &alias{Index: i, Alias: a}})
+	if err := db.post(db.url("_aliases"), actions, nil); err != nil {
+		return errgo.Notef(err, "error updating aliases")
+	}
+	return nil
+}
+
+// Create document attempts to create a new document at index/type_/id with the
+// contents in doc. If the document already exists then CreateDocument will return false,
```
rogpeppe: I think this might be better to just return an error, but say that if the document already exists, it returns ErrAlreadyExists.

Then the sense of the return value is more obvious and the code using it should look more obvious too.

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -246,150 +243,8 @@ func (s *Suite) TestPutMapping(c *gc.C) {
...
-	c.Assert(stored, gc.Equals, true)
-	var result map[string]string
-	err = s.ES.GetDocument(s.TestIndex, "testtype", "a", &result)
-	c.Assert(result["a"], gc.Equals, "c")
-	exists, err = s.ES.EnsureID(s.TestIndex, "testtype", "a")
-	c.Assert(err, gc.IsNil)
-	c.Assert(exists, gc.Equals, true)
-}
-
-func (s *Suite) TestIndexPutVersionUpdateEarlierDocumentVersion(c *gc.C) {
```
rogpeppe, line 339: Why delete these tests?

mhilton, line 339: I deleted the code that they tested.

internal/elasticsearch/lock/lock.go
```diff
@@ -0,0 +1,86 @@
...
+			return false, errgo.Notef(err, "error querying lock")
+		}
+		if !held.Time.Add(held.Duration).Before(time.Now()) {
+			return false, nil
+		}
+		l.locked, err = l.Database.PutDocumentVersion(".lock", "lock", l.Name, d.Version, lck)
+		if err != nil {
+			return false, errgo.Notef(err, "error updating lock")
+		}
+	} else {
```
rogpeppe: Perhaps put this clause first and return early to save a bunch of indentation?

internal/elasticsearch/lock/lock.go
```diff
@@ -0,0 +1,86 @@
...
+	es "github.com/juju/charmstore/internal/elasticsearch"
+)
+
+// lock is the structure of a lock on the elasticsearch database
+type lock struct {
+	Time     time.Time
+	Duration time.Duration
+}
+
+// Lock provides a lock on an elasticsearch database.
```
rogpeppe: Perhaps given the name, it might be worth adding:

// Note that, unlike in memory mutexes, the methods on this
// type are not safe to use concurrently.

?

Alternatively, add a mutex and make it safe (which is probably my preferred option actually)

internal/elasticsearch/lock/lock.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"gopkg.in/errgo.v1"
+
+	es "github.com/juju/charmstore/internal/elasticsearch"
+)
+
+// lock is the structure of a lock on the elasticsearch database
+type lock struct {
+	Time     time.Time
+	Duration time.Duration
```
rogpeppe: I don't see what the Duration is adding here. Couldn't you just store an expiry time?

internal/elasticsearch/lock/lock_test.go
```diff
@@ -0,0 +1,80 @@
...
+	locked, err := l.Lock(time.Second)
+	c.Assert(err, gc.Equals, nil)
+	c.Assert(locked, gc.Equals, true)
+	time.Sleep(time.Second)
+	locked, err = l.Lock(time.Second)
+	c.Assert(err, gc.Equals, nil)
+	c.Assert(locked, gc.Equals, true)
+	err = l.Unlock()
+	c.Assert(err, gc.Equals, nil)
+}
```
rogpeppe: I'd really hope to have some concurrent logic here to test that the locks do actually work when several clients are contesting over them.

internal/elasticsearch/elasticsearch_test.go
```diff
@@ -495,3 +350,32 @@ func (s *Suite) TestEscapeRegexp(c *gc.C) {
...
+	indexes, err = s.ES.ListIndexesForAlias(alias)
+	c.Assert(err, gc.Equals, nil)
+	c.Assert(indexes, gc.HasLen, 1)
+	c.Assert(indexes[0], gc.Equals, index1)
+	err = s.ES.Alias(index2, alias)
+	c.Assert(err, gc.Equals, nil)
+	indexes, err = s.ES.ListIndexesForAlias(alias)
+	c.Assert(err, gc.Equals, nil)
+	c.Assert(indexes, gc.HasLen, 1)
+	c.Assert(indexes[0], gc.Equals, index2)
```
rogpeppe, line 397: A few comments might make this test logic easier to follow (and easier to understand what conditions are being tested for when we come to change the code later)

internal/elasticsearch/lock/lock.go
mhilton: I didn't think of that.

internal/elasticsearch/elasticsearch.go
```diff
@@ -54,24 +118,40 @@ func (db *Database) DeleteIndex(index string) error {
...
 // GetDocument retrieves the document with the given index, type_ and id and
 // unmarshals the json response into v.
 func (db *Database) GetDocument(index, type_, id string, v interface{}) error {
-	if err := db.get(db.url(index, type_, id, "_source"), nil, v); err != nil {
+	d, err := db.GetESDocument(index, type_, id)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	if !d.Found {
+		return ElasticSearchError{Status: http.StatusNotFound, Err: fmt.Sprintf("%s/%s/%s not found", index, type_, id)}
```
rogpeppe: split this across lines?

rogpeppe: alternatively just define:

```
var ErrNotFound = errgo.New("elasticsearch document not found")
```

and do:
     if !d.Found {
         return ErrNotFound
    }

alternatively, if you deem the message important enough:

```
type NotFoundError struct {
    Index string
    Type string
    Id string
 }
 func (e *NotFoundError) Error() string {
      return fmt.Sprintf("%s/%s/%s not found", e.Index, e.Type, e.Id)
 }
```

Also, if you choose to stick with what you've got, the http status is not really appropriate here. Better would be params.ErrNotFound, which is our common currency for errors in the charm store, and can be used to derive http status codes if needed.

internal/charmstore/search.go
```diff
@@ -113,20 +135,64 @@ func (ses *StoreElasticSearch) search(sp SearchParams) (SearchResult, error) {
...
+//
+// A database lock is aquired by sync before creating the new database,
+// this is to prevent a race condition where two charmstores are attempting
+// to synchronise elasticsearch at the same time.
+func (ses *storeElasticSearch) sync() error {
+	if ses == nil || ses.Database == nil {
+		return nil
+	}
+	l := lock.NewLock(ses.Database, "essync")
+	locked, err := l.Lock(5 * time.Minute)
```
rogpeppe: I see that the lock package is tested, but I don't see anywhere that tests this logic in place here.

rogpeppe:
This looks great in general, thanks!
I have bunch of mostly minor suggestions above.
I'm looking forward to the day we can ditch essync entirely!

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/564/

internal/charmstore/search.go
mhilton, line 97: I don't think it's quite there yet and I like to keep it consistent.

internal/charmstore/server.go
```diff
@@ -30,11 +29,11 @@ type ServerParams struct {
 // versions using db to store that charm store data.
 // The key of the versions map is the version name.
 // The handler configuration is provided to all version handlers.
```
rogpeppe, line 33: Perhaps document that si can be nil (assuming it can be) ?

I wonder if all these parameters should move into NewServerParams.

rogpeppe, line 33: The latter is just something to think about in the future BTW - no need to do it now.

internal/charmstore/store.go
```diff
@@ -38,15 +38,24 @@ type Store struct {
...
+func NewStore(db *mgo.Database, si *SearchIndex) (*Store, error) {
 	s := &Store{
 		DB:        StoreDatabase{db},
 		BlobStore: blobstore.New(db, "entitystore"),
-		ES:        ses,
 	}
 	if err := s.ensureIndexes(); err != nil {
 		return nil, errgo.Notef(err, "cannot ensure indexes")
 	}
+	if si != nil {
```
rogpeppe:
```
if si == nil {
    return s, nil
}
```

?

internal/elasticsearch/elasticsearch.go
```diff
@@ -25,8 +25,25 @@ import (
 	"gopkg.in/errgo.v1"
 )
 
+const (
+	// Internal provides elasticsearches "internal" versioning system, as described in
```
rogpeppe: s/elasticsearches/elasticsearch's/ ?

Might it be worth making these typed constants, so it's obvious where they fit in the rest of the API?

mhilton: I didn't want to do that, because I didn't want to put all the possible values in.

internal/charmstore/store.go
mhilton: I like that, done

internal/elasticsearch/elasticsearch.go
```diff
@@ -384,3 +449,30 @@ var regexpReplacer = strings.NewReplacer(
...
+	Alias string `json:"alias"`
+}
+
+// action is an action that can be performed on an alias
+type action struct {
+	Remove *alias `json:"remove,omitempty"`
+	Add    *alias `json:"add,omitempty"`
+}
+
+// getError
```
rogpeppe: // getError derives the local error types from the underlying
// elastic search errors.

?

internal/charmstore/server.go
mhilton, line 33: done

internal/elasticsearch/elasticsearch.go
```diff
@@ -40,13 +57,73 @@ type Database struct {
...
+	}
+	return nil
+}
+
+// Create document attempts to create a new document at index/type_/id with the
+// contents in doc. If the document already exists then CreateDocument will return false,
+// otherwise a new document is created and CreateDocument returns true.
+// CreateDocument return a non-nil error if any other error occurs.
+// See http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/create-doc.html#create-doc
+// for further details.
```
rogpeppe, line 101: It would be worth documenting the possible error returns here and for the functions below, I think.

```
// It returns ErrConflict if there is already a document with the
// given id.
```

internal/elasticsearch/elasticsearch.go
```diff
@@ -25,8 +25,25 @@ import (
...
+
+	// ExternalGTE provides elasticsearches "external_gte" versioning system, as described in
+	// http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-index_.html#_version_types
+	ExternalGTE = "external_gte"
+)
+
 var log = loggo.GetLogger("charmstore.elasticsearch")
 
+var ErrConflict = errgo.New("conflict")
+var ErrNotFound = errgo.New("not found")
```
rogpeppe: Perhaps we could these slightly more descriptive, so that we know what package they've come from at any rate. e.g.

var (
    ErrConflict = errgo.New("elastic search document conflict")
    ErrNotFound = errgo.New("elastic search document not found")
)

?

internal/elasticsearch/elasticsearch.go
```diff
@@ -54,31 +131,47 @@ func (db *Database) DeleteIndex(index string) error {
 // EnsureID tests to see a document of the given index, type_, and id exists
 // in ElasticSearch.
 func (db *Database) EnsureID(index, type_, id string) (bool, error) {
```
rogpeppe: I know this isn't a change that's been made in this PR, but "Ensure" is often used to mean "make it so if it isn't already". I'd be inclined to rename this to IdExists or something like that.

internal/elasticsearch/elasticsearch.go
```diff
@@ -91,6 +184,20 @@ func (db *Database) ListAllIndexes() ([]string, error) {
 	return indexes, nil
 }
 
+// ListIndexesForAlias retreieves the list of all indexes in the elasticsearch database
+// that have the alias a.
+func (db *Database) ListIndexesForAlias(a string) ([]string, error) {
+	var result map[string]interface{}
```
rogpeppe: Perhaps struct{} instead of interface{} here, so we don't bother unmarshalling all the contents?

internal/elasticsearch/elasticsearch_test.go
rogpeppe, line 339: But PutDocumentVersion still exists, right? Aren't these tests still relevant, even if only as regression tests?

internal/charmstore/search.go
```diff
@@ -54,30 +65,46 @@ var deprecatedSeries = map[string]bool{
 	"saucy":   true,
 }
 
-// Put inserts the mongodoc.Entity into elasticsearch if elasticsearch
+// put inserts the mongodoc.Entity into elasticsearch if elasticsearch
```
jrwren: The doc needs to be updated to reflect the changes.

internal/elasticsearch/elasticsearch_test.go
mhilton, line 339: Yes, but these tests were for the PutDocumentVersion function on the elasticsearch.Index data type which I have removed because the application is now maintaining it's own indexes. The (almost identical) test functions for the PutDocumentVersion function on the Database data type are still present.

internal/charmstore/search.go
```diff
@@ -113,14 +140,91 @@ func (ses *StoreElasticSearch) search(sp SearchParams) (SearchResult, error) {
...
+
+// sync populates an elasticsearch index with all the data currently stored in
+// mongodb.
+func (ses *storeElasticSearch) sync() error {
+	if ses == nil || ses.SearchIndex == nil {
+		return nil
+	}
 	var result mongodoc.Entity
-	iter := store.DB.Entities().Find(nil).Iter()
+	iter := ses.DB.Entities().Find(nil).Select(bson.M{"_id": 1}).Iter()
```
jrwren, line 262: It is kind of strange to query just ID here, only to query again inside of put for the entire document? I see that it is an optimization to prevent pulling the entire database when deprecated series are being ignored. Can we add a comment explaining this?

jrwren:
QA: I followed exactly and it worked perfectly.

internal/elasticsearch/elasticsearch.go
mhilton: done

internal/elasticsearch/elasticsearch.go
mhilton: I've changed it to HasDocument

internal/elasticsearch/elasticsearch.go
mhilton: done

internal/charmstore/search.go
mhilton: done.

internal/charmstore/search.go
mhilton, line 262: done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/571/

internal/elasticsearch/elasticsearch.go
```diff
@@ -111,37 +218,45 @@ func (db *Database) PostDocument(index, type_ string, doc interface{}) (string,
...
+// parameter is the same as the currently stored version. The type_ parameter
+// controls how the document will be indexed. PutDocumentVersion returns a
+// bool indicating if the data was stored and an error. If the data was not stored
+// due to a version conflict the returned value will be false, nil.
+// See http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-index_.html#index-versioning
+// and http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-index_.html#_version_types for more information.
+func (db *Database) PutDocumentVersionWithType(
+	index, type_, id string,
+	version int64,
+	version_type string,
```
rogpeppe: s/version_type/versionType/

internal/elasticsearch/elasticsearch.go
rogpeppe: I guess we wouldn't _need_ to put all the possible values in. I understand the
reluctance though. At least we should document where it might be appropriate
to use these constants. I _presume_ from inspection of the code that they're
intended for the version type argument to PutDocumentVersionWithType
but it would be nice to say so.

internal/elasticsearch/elasticsearch.go
rogpeppe: This comment needs updating to fit the new signature - it doesn't return a bool any more.

internal/charmstore/store.go
```diff
@@ -38,15 +38,24 @@ type Store struct {
...
+		s.ES = &storeElasticSearch{
+			SearchIndex: si,
+			DB:          s.DB,
+		}
+		if err := si.ensureIndexes(false); err != nil {
+			return nil, errgo.Notef(err, "cannot ensure elasticsearch indexes")
+		}
+		go s.ES.sync()
```
rogpeppe: It would be polite to produce a log message if sync returns an error, I think.

internal/elasticsearch/elasticsearch.go
mhilton: done

internal/charmstore/store.go

internal/elasticsearch/elasticsearch_test.go
mhilton, line 397: done

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/572/

internal/charmstore/search_test.go
```diff
@@ -470,6 +454,60 @@ func (s *StoreSearchSuite) TestBoosting(c *gc.C) {
...
+	c.Assert(indexes, gc.HasLen, 1)
+	c.Assert(indexes[0], gc.Equals, index)
+}
+
+func (s *StoreSearchSuite) TestEnsureConcurrent(c *gc.C) {
+	s.store.ES.Index = s.TestIndex + "-ensure-index-conc"
+	indexes, err := s.ES.ListIndexesForAlias(s.store.ES.Index)
+	c.Assert(err, gc.Equals, nil)
+	c.Assert(indexes, gc.HasLen, 0)
+	go func() {
```
rogpeppe, line 486: I'm not sure this is enough concurrency to actually tickle the possible problems.

Also, we shouldn't call Assert in goroutines (use Check instead),
and we should make sure that the goroutine has completed
by the time the test ends.

How about testing the scenario that will actually happen in practice
(ensureIndexes(false) followed by sync()), adding a bunch of stuff to the
database, and ramping up the concurrency quite a bit (say 5-10 concurrent
goroutines)?

Then run the test and see if you get reasonable coverage of the ensureIndexes
method.

internal/charmstore/search.go
```diff
@@ -113,14 +141,93 @@ func (ses *StoreElasticSearch) search(sp SearchParams) (SearchResult, error) {
...
+		if !force && old.Version >= esSettingsVersion {
+			return nil
+		}
+	}
+	index, err := si.newIndex()
+	if err != nil {
+		return errgo.Notef(err, "cannot create index")
+	}
+	new := version{Version: esSettingsVersion, Index: index}
+	updateVersion := func(d elasticsearch.Document, v version) error {
```
rogpeppe: I'm not sure that using this inline function is better than just doing
the obvious:

```
// Update the document version
if d.Found {
   err = si.PutDocumentVersion(".versions", "version", si.Index, d.Version, v)
} else {
   err = si.CreateDocument(".versions", "version", si.Index, v)
}
if errgo.Cause(err) == elasticsearch.ErrConflict {
    // ...
    return nil
}
if err != nil {
}
```

internal/charmstore/search_test.go
mhilton, line 486: I did check that the coverage of this test covers the code path I expected to test, But I'm happy to ramp it up If you think that's necessary.

internal/charmstore/search.go
```diff
@@ -113,14 +141,93 @@ func (ses *StoreElasticSearch) search(sp SearchParams) (SearchResult, error) {
...
+		if !d.Found {
+			return si.CreateDocument(".versions", "version", si.Index, v)
+		}
+		return si.PutDocumentVersion(".versions", "version", si.Index, d.Version, v)
+	}
+	if err := updateVersion(d, new); err != nil {
+		if err == elasticsearch.ErrConflict {
```
rogpeppe: please use errgo.Cause to check for error causes, to allow for error wrapping:

   if errgo.Cause(err) == elasticsearch.ErrConflict {

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/573/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/574/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/575/

mhilton:
:shipit:

jujugui:

PR #235, opened 2014-11-26
Add viewBox attribute to icons that lack it

makyo:
This will resolve the issues with icons not being sized properly when embedded in other SVGs such as diagrams and the GUI.


makyo:
This branch is admittedly messy and may need to be taken over while I'm out of town, however it should still serve as a good starting point to adding the correct attribute to the SVG.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/584/

PR #236, opened 2014-11-26
Denormalize entity ids in Mongo.

frankban:
When adding charms or bundles, store their URL parts
in separate fields.
Also implement a simple migration strategy.

### DIFF
--- a/internal/charmstore/elasticsearch.go
+++ b/internal/charmstore/elasticsearch.go
@@ -52,24 +52,6 @@ const esMappingJSON = `
   "entity" : {
     "dynamic" : "false",
     "properties" : {
-      "Name" : {
-        "type" : "string",
-        "index" : "not_analyzed",
-        "omit_norms" : true,
-        "index_options" : "docs"
-      },
-      "Series" : {
-        "type" : "string",
-        "index" : "not_analyzed",
-        "omit_norms" : true,
-        "index_options" : "docs"
-      },
-      "User" : {
-        "type" : "string",
-        "index" : "not_analyzed",
-        "omit_norms" : true,
-        "index_options" : "docs"
-      },  
       "URL" : {
         "type" : "multi_field",
         "fields" : {
@@ -89,6 +71,28 @@ const esMappingJSON = `
       "BaseURL" : {
         "type" : "string",
         "index": "not_analyzed",
+        "index_options" : "docs"
+      },
+      "User" : {
+        "type" : "string",
+        "index" : "not_analyzed",
+        "omit_norms" : true,
+        "index_options" : "docs"
+      },
+      "Name" : {
+        "type" : "string",
+        "index" : "not_analyzed",
+        "omit_norms" : true,
+        "index_options" : "docs"
+      },
+      "Revision" : {
+        "type" : "integer",
+        "index" : "not_analyzed"
+      },
+      "Series" : {
+        "type" : "string",
+        "index" : "not_analyzed",
+        "omit_norms" : true,
         "index_options" : "docs"
       },
       "BlobHash" : {
--- a/internal/charmstore/elasticsearch.go
+++ b/internal/charmstore/elasticsearch.go
@@ -10,7 +10,7 @@ var (
 	esMapping = mustParseJSON(esMappingJSON)
 )
 
-const esSettingsVersion = 1
+const esSettingsVersion = 2
 
 func mustParseJSON(s string) interface{} {
 	var j json.RawMessage
--- a/internal/charmstore/migrations.go
+++ b/internal/charmstore/migrations.go
@@ -5,25 +5,85 @@ package charmstore
 
 import (
 	"gopkg.in/errgo.v1"
+	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 
 	"github.com/juju/charmstore/internal/mongodoc"
 )
 
 // migrations holds all the migration functions that are executed in the order
-// they are defined when the charm store server is started.
-// To introduce a new database migration, just add the corresponding migration
-// function to this list.
-var migrations = []func(db StoreDatabase) error{
-	denormalizeEntityIds,
+// they are defined when the charm store server is started. Each migration is
+// associated with a name that is used to check whether the migration has been
+// already run. To introduce a new database migration, just add the
+// corresponding migration name and function to this list. Note that names must
+// be unique across the list.
+var migrations = []migration{{
+	name:    "entity ids denormalization",
+	migrate: denormalizeEntityIds,
+}}
+
+type migration struct {
+	name    string
+	migrate func(StoreDatabase) error
 }
 
 // Migrate starts the migration process using the given database.
 func migrate(db StoreDatabase) error {
-	for _, f := range migrations {
-		if err := f(db); err != nil {
+	// Retrieve already executed migrations.
+	executed, err := getExecuted(db)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+
+	// Execute required migrations.
+	for _, m := range migrations {
+		if executed[m.name] {
+			logger.Debugf("skipping already executed migration: %s", m.name)
+			continue
+		}
+		logger.Infof("starting migration: %s", m.name)
+		if err := m.migrate(db); err != nil {
+			return errgo.Notef(err, "error executing migration: %s", m.name)
+		}
+		if err := setExecuted(db, m.name); err != nil {
 			return errgo.Mask(err)
 		}
+		logger.Infof("migration completed: %s", m.name)
+	}
+	return nil
+}
+
+func getExecuted(db StoreDatabase) (map[string]bool, error) {
+	// Retrieve the already executed migration names.
+	executed := make(map[string]bool)
+	var doc mongodoc.Migration
+	if err := db.Migrations().Find(nil).Select(bson.D{{"executed", 1}}).One(&doc); err != nil {
+		if err == mgo.ErrNotFound {
+			return executed, nil
+		}
+		return nil, errgo.Notef(err, "cannot retrieve executed migrations")
+	}
+
+	names := make(map[string]bool, len(migrations))
+	for _, m := range migrations {
+		names[m.name] = true
+	}
+	for _, name := range doc.Executed {
+		// Check that the already executed migrations are known.
+		if !names[name] {
+			return nil, errgo.Newf("unexpected already executed migration: %s", name)
+		}
+		// Collect the name of the executed migration.
+		executed[name] = true
+	}
+	return executed, nil
+}
+
+func setExecuted(db StoreDatabase, name string) error {
+	if _, err := db.Migrations().Upsert(nil, bson.D{{
+		"$addToSet", bson.D{{"executed", name}},
+	}}); err != nil {
+		return errgo.Notef(err, "cannot add %s to executed migrations", name)
 	}
 	return nil
 }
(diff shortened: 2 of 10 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/586/

internal/mongodoc/doc.go
```diff
@@ -18,6 +18,18 @@ type Entity struct {
 	// e.g. cs:wordpress, cs:~user/foo
 	BaseURL *charm.Reference
 
+	// User holds the user part of the entity URL (for instance, "joe").
```
mhilton, line 21: Now these are in the mongodoc can we remove the special esDoc that does the same thing for elasticsearch?

frankban, line 21: I presume we can, could you please add a card for that?

urosj, line 21: We must and also update the entity doc if needed.

internal/charmstore/migrations.go
```diff
@@ -0,0 +1,62 @@
...
+func denormalizeEntityIds(db StoreDatabase) error {
+	logger.Debugf("starting entity ids migration")
+	entities := db.Entities()
+	var entity mongodoc.Entity
+	iter := entities.Find(bson.D{{
+		// Use the name field to collect not migrated entities.
+		"name", bson.D{{"$exists", false}},
+	}}).Select(bson.D{{"_id", 1}}).Iter()
+
+	for iter.Next(&entity) {
```
rogpeppe, line 105: Rather than running this every time, I wonder if we could just set a flag at the end of the migration to signify that it's been done. Then we'd only need to check that flag to see if we needed to do the migration. Since we make migration functions idempotent, when we start for the first time, we'll run all the migrations (all will be no-ops) and set all the current flags.

We could use a separate collection to hold the flags (just a simple string->bool mapping).

This way, we can potentially end up with a fair number of migrations but not pay
a significant overhead when adding charmstore units (this also makes it reasonable to
have costly migrations that lock the whole database).

What do you think?

internal/charmstore/migrations.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+	}
+	return nil
+}
+
+// denormalizeEntityIds adds the user, name, revision and series fields to
+// entities where those fields are missing.
+// This function is not supposed to be called directly.
+func denormalizeEntityIds(db StoreDatabase) error {
+	logger.Debugf("starting entity ids migration")
```
mhilton: perhaps info level?

internal/v4/archive.go
```diff
@@ -386,7 +386,9 @@ func (h *Handler) bundleCharms(ids []string) (map[string]charm.Charm, error) {
 }
 
 // entityCharm implements charm.Charm.
-type entityCharm mongodoc.Entity
+type entityCharm struct {
```
rogpeppe, line 389: Why change this?

mhilton:
I'd like to see the elasticsearch part updated too so we don't have double denormalised data.

internal/v4/archive.go
frankban, line 389: In the previous impl, Revision is both a method and a field.

internal/charmstore/migrations.go
frankban: Done.

mhilton:
For elasticsearch we need to make sure that the esDoc is removed if all of it's fields are now in mongodoc.Entity (I believe they are). Also the mapping in github.com/juju/charmstore/internal/charmstore/elasticsearch.go called esMappingJSON needs to be checked that it includes all the fields that are now in the mongodoc.Entity (a brief look suggests that Revision might need to be added).

internal/charmstore/migrations.go
frankban, line 105: Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/588/

internal/mongodoc/doc.go
frankban, line 21: Done.

internal/mongodoc/doc.go
```diff
@@ -80,6 +92,34 @@ type Entity struct {
 	Contents map[FileId]ZipFile `json:",omitempty" bson:",omitempty"`
 }
 
+type FileId string
```
frankban, line 95: This code is just moved.

internal/mongodoc/doc.go
```diff
@@ -127,30 +167,8 @@ const (
...
-	Size int64
-}
-
-// Valid reports whether f is a valid (non-zero) reference to
-// a zip file.
-func (f ZipFile) IsValid() bool {
-	// Note that no valid zip files can start at offset zero,
-	// because that's where the zip header lives.
-	return f != ZipFile{}
+// Migration holds information about the database migration.
```
frankban, line 170: New code.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/590/

internal/charmstore/migrations_test.go
```diff
@@ -0,0 +1,350 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
```
mhilton, line 2: I'm not sure, but why is some code AGPL and some LGPL?

mhilton:
:+1: Looks good, thanks for sorting out the elasticsearch. Note I haven't done any QA.

internal/charmstore/migrations_test.go
frankban, line 2: This is a really good question. Looking at the files we use both, and some files don't even have the license header. I'd be inclined to say this should be AGPL, but IANAL.
We need a branch to fix all the license headers.

internal/charmstore/elasticsearch.go
```diff
@@ -91,6 +73,28 @@ const esMappingJSON = `
         "index": "not_analyzed",
         "index_options" : "docs"
       },
+      "User" : {
```
mhilton, line 76: One other thing to do with this file is increment the esSettingsVersion number on line 13. Thanks

frankban, line 76: Done.

internal/charmstore/migrations.go
```diff
@@ -0,0 +1,121 @@
...
+	"gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+
+	"github.com/juju/charmstore/internal/mongodoc"
+)
+
+// migrations holds all the migration functions that are executed in the order
+// they are defined when the charm store server is started. Each migration is
+// associated with a name that is used to check whether the migration has been
+// already run. To introduce a new database migration, just add the
```
rogpeppe: s/just//

jujugui:
Test FAILed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/587/

internal/charmstore/migrations.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	names := make(map[string]bool, len(migrations))
+	for _, m := range migrations {
+		names[m.name] = true
+	}
+	for _, name := range doc.Executed {
+		// Check that the already executed migrations are known.
+		if !names[name] {
+			return nil, errgo.Newf("unexpected already executed migration: %s", name)
```
rogpeppe: "found unknown migration %q; running old charm store code on newer charm store database?"

?

internal/charmstore/migrations.go
```diff
@@ -0,0 +1,121 @@
...
+	// Retrieve already executed migrations.
+	executed, err := getExecuted(db)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+
+	// Execute required migrations.
+	for _, m := range migrations {
+		if executed[m.name] {
+			logger.Debugf("skipping already executed migration: %s", m.name)
```
rogpeppe, line 43: Infof ?

internal/charmstore/migrations.go
```diff
@@ -0,0 +1,121 @@
...
+// migration holds a migration function with its corresponding name.
+type migration struct {
+	name    string
+	migrate func(StoreDatabase) error
+}
+
+// Migrate starts the migration process using the given database.
+func migrate(db StoreDatabase) error {
+	// Retrieve already executed migrations.
+	executed, err := getExecuted(db)
```
rogpeppe, line 35: I wonder if this should be inside the loop, so if two units are doing this at the
same time, they don't necessarily both do all the migrations. I guess that's
more of an issue if we have migrations that are mutually exclusive though,
and that's another bag of worms.

internal/charmstore/migrations.go
```diff
@@ -0,0 +1,121 @@
...
+// denormalizeEntityIds adds the user, name, revision and series fields to
+// entities where those fields are missing.
+// This function is not supposed to be called directly.
+func denormalizeEntityIds(db StoreDatabase) error {
+	entities := db.Entities()
+	var entity mongodoc.Entity
+	iter := entities.Find(bson.D{{
+		// Use the name field to collect not migrated entities.
+		"name", bson.D{{"$exists", false}},
+	}}).Select(bson.D{{"_id", 1}}).Iter()
```
rogpeppe, line 102: defer iter.Close()

internal/charmstore/migrations_test.go
```diff
@@ -0,0 +1,350 @@
...
+			name: name,
+			migrate: func(StoreDatabase) error {
+				s.executed = append(s.executed, name)
+				return nil
+			},
+		}
+	}
+	s.patchMigrations(c, ms)
+}
+
```
rogpeppe, line 71: I think we should probably have a test that all the migrations are
idempotent and work OK when run concurrently, because that
may well happen when two units are bootstrapped at the same time.

internal/charmstore/store.go
```diff
@@ -84,6 +84,9 @@ func (s *Store) ensureIndexes() error {
 	}, {
 		s.DB.Logs(),
 		mgo.Index{Key: []string{"urls"}},
+	}, {
+		s.DB.Migrations(),
+		mgo.Index{Key: []string{"executed"}},
```
rogpeppe: I really don't think it's worth having an index on a single-document collection :)

rogpeppe:
LGTM with a few minor suggestions, thanks!

internal/charmstore/migrations.go
frankban: Done.

internal/charmstore/migrations.go
frankban: Done.

internal/charmstore/migrations.go
frankban, line 43: Done.

internal/charmstore/migrations.go
frankban, line 102: Done.

internal/charmstore/migrations.go
frankban, line 35: I'll leave this as is for now.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/588/

internal/charmstore/migrations_test.go
frankban, line 71: Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/589/

frankban:
Thanks for the great reviews!
:shipit:

jujugui:

PR #237, opened 2014-11-26
Add boosting of bundles.

mhilton:
Bundles are now boosted to be just above the highest series.


mhilton:
To QA run charmd on a server with elasticsearch running and perform some queries. If all other things are equal bundles should come higher than trusty which comes higher than precise. But promulgated and other things and matches should still trump those.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/587/

mhilton:
:shipit:

jujugui:

PR #240, opened 2014-11-28
Added last ingestion times to debug/status output.

mhilton:
Debug status now has an "ingestion" entry with details of the last ingestion.

### DIFF
--- a/internal/v4/api_test.go
+++ b/internal/v4/api_test.go
@@ -1275,6 +1275,20 @@ func (s *APISuite) TestStatus(c *gc.C) {
 	}
 	now := time.Now()
 	s.PatchValue(v4.StartTime, now)
+	start := now.Add(-2 * time.Hour)
+	s.addLog(c, &mongodoc.Log{
+		Data:  []byte(`"ingestion started"`),
+		Level: mongodoc.InfoLevel,
+		Type:  mongodoc.IngestionType,
+		Time:  start,
+	})
+	end := now.Add(-1 * time.Hour)
+	s.addLog(c, &mongodoc.Log{
+		Data:  []byte(`"ingestion completed"`),
+		Level: mongodoc.InfoLevel,
+		Type:  mongodoc.IngestionType,
+		Time:  end,
+	})
 	storetesting.AssertJSONCall(c, storetesting.JSONCallParams{
 		Handler:      s.srv,
 		URL:          storeURL("debug/status"),
@@ -1300,6 +1314,11 @@ func (s *APISuite) TestStatus(c *gc.C) {
 				Value:  now.String(),
 				Passed: true,
 			},
+			"ingestion": {
+				Name:   "Ingestion",
+				Value:  "started: " + start.Format(time.RFC3339) + ", completed: " + end.Format(time.RFC3339),
+				Passed: true,
+			},
 		},
 	})
 }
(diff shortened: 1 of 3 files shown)


jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/599/

internal/v4/status.go
```diff
@@ -103,6 +111,33 @@ func (h *Handler) checkEntities() (string, bool) {
...
 }
 
+func (h *Handler) checkIngestion() (string, bool) {
+	var start time.Time
+	var end time.Time
+	var log mongodoc.Log
+	iter := h.store.DB.Logs().Find(bson.D{{"level", mongodoc.InfoLevel}}).Sort("-time").Iter()
+	for iter.Next(&log) {
+		var msg string
+		if err := json.Unmarshal(log.Data, &msg); err != nil {
```
rogpeppe, line 123: Hmm, I was expecting the data to be a JSON object, not just a json string, which would make it easy to expand later if desired. I wonder if we should fix that sooner rather than later.

internal/v4/status.go
```diff
@@ -103,6 +111,33 @@ func (h *Handler) checkEntities() (string, bool) {
...
+			end = log.Time
+		}
+		if !start.IsZero() && !end.IsZero() {
+			break
+		}
+	}
+	if err := iter.Close(); err != nil {
+		return "Cannot query ingestion logs: " + err.Error(), false
+	}
+	return fmt.Sprintf("started: %s, completed: %s", start.Format(timeFormat), end.Format(timeFormat)), !(start.IsZero() || end.IsZero())
```
rogpeppe: I wonder if, when there's an ingestion in progress, we should just say:
"started %s, not yet complete".

Alternatively, I think we would like to know how long the ingestions are
taking even if there's one in progress, so how about these messages?

```
started 2015-01-01T01:01:01; previous duration 5m4.456s
finished 2015-01-01T01:01:0; duration 5m4.456s
```

(that would require a little more info gathering, but perhaps worth it)

internal/v4/status.go
```diff
@@ -14,6 +16,8 @@ import (
 	"github.com/juju/charmstore/params"
 )
 
+const timeFormat = "2006-01-02 15:04:05.000 -0700 MST"
```
rogpeppe: Not just RFC3339 as usual?

internal/v4/status.go
```diff
@@ -103,6 +111,33 @@ func (h *Handler) checkEntities() (string, bool) {
 	return fmt.Sprintf("%d charms; %d bundles; %d promulgated", charms, bundles, promulgated), true
 }
 
+func (h *Handler) checkIngestion() (string, bool) {
+	var start time.Time
+	var end time.Time
+	var log mongodoc.Log
+	iter := h.store.DB.Logs().Find(bson.D{{"level", mongodoc.InfoLevel}}).Sort("-time").Iter()
```
frankban: I'd further filter logs here using type=mongodoc.IngestionType.

rogpeppe: This should also select by type=mongodoc.IngestionType

internal/v4/status.go
frankban: What if an ingestion started but it's not yet completed?
Should we check complete time > start time?

internal/v4/api_test.go
```diff
@@ -1275,6 +1275,20 @@ func (s *APISuite) TestStatus(c *gc.C) {
 	}
 	now := time.Now()
 	s.PatchValue(v4.StartTime, now)
+	start := now.Add(-2 * time.Hour)
```
frankban, line 1278: Could you test a scenario in which the first ingestion started but it's not yet done?

internal/v4/status.go
frankban, line 123: My first implementation just uses JSON strings because the goal was to output human readable errors.
But given we now need to parse those messages (e.g. "ingestion started" in message) I'd be +1 on using an object.

urosj:
QA done.
- used existing CS DB, dates were set as described
- used completely new, the dates are set on the 1.1.1 at midnight (still tempted to add a comment that ingestion will start on the 7th day :D :D)
- interrupted ingestion, start date was set as stated, end date was not set - good!

QA-wise :+1:

internal/v4/status.go
mhilton: RFC3339 is a little less readable and it isn't as consistent with the charmstore started time. I'm happy to change it if people would prefer ISO8601.

frankban:
:+1: with changes, thank you!

internal/v4/status.go
mhilton: Done

internal/v4/status.go
mhilton: Done

internal/v4/api_test.go
mhilton, line 1278: done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/602/

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/603/

mhilton:
:shipit:

jujugui:

PR #244, opened 2014-12-02
Re-implement the charm-event legacy API endpoint.

frankban:
Only support what strictly required by "juju publish".
This is based on Jay's work.

### DIFF
--- a/internal/legacy/api.go
+++ b/internal/legacy/api.go
@@ -258,7 +258,7 @@ func (h *Handler) serveCharmEvent(_ http.Header, req *http.Request) (interface{}
 		// Retrieve the charm.
 		entity, err := h.store.FindEntity(id, "_id", "uploadtime", "extrainfo")
 		if err != nil {
-			if err == params.ErrNotFound {
+			if errgo.Cause(err) == params.ErrNotFound {
 				// The old API actually returned "entry not found"
 				// on *any* error, but it seems reasonable to be
 				// a little more descriptive for other errors.
--- a/internal/legacy/api.go
+++ b/internal/legacy/api.go
@@ -180,10 +180,9 @@ func (h *Handler) serveCharmInfo(_ http.Header, req *http.Request) (interface{},
 			c.CanonicalURL = curl.String()
 			c.Sha256 = entity.BlobHash256
 			c.Revision = curl.Revision
-			if digest, found := entity.ExtraInfo[params.BzrDigestKey]; found {
-				if err := json.Unmarshal(digest, &c.Digest); err != nil {
-					c.Errors = append(c.Errors, "cannot unmarshal digest: "+err.Error())
-				}
+			c.Digest, err = entityBzrDigest(&entity)
+			if err != nil {
+				c.Errors = append(c.Errors, err.Error())
 			}
 			h.store.IncCounterAsync(charmStatsKey(curl, params.StatsCharmInfo))
 		} else {
@@ -275,12 +274,23 @@ func (h *Handler) serveCharmEvent(_ http.Header, req *http.Request) (interface{}
 		c.Kind = "published"
 		c.Revision = id.Revision
 		c.Time = entity.UploadTime.UTC().Format(time.RFC3339)
-		if digest, found := entity.ExtraInfo[params.BzrDigestKey]; found {
-			if err := json.Unmarshal(digest, &c.Digest); err != nil {
-				c.Errors = []string{"cannot unmarshal digest: " + err.Error()}
-			}
+		c.Digest, err = entityBzrDigest(entity)
+		if err != nil {
+			c.Errors = []string{err.Error()}
 		}
 		h.store.IncCounterAsync(charmStatsKey(id, params.StatsCharmEvent))
 	}
 	return response, nil
+}
+
+func entityBzrDigest(entity *mongodoc.Entity) (string, error) {
+	value, found := entity.ExtraInfo[params.BzrDigestKey]
+	if !found {
+		return "", nil
+	}
+	var digest string
+	if err := json.Unmarshal(value, &digest); err != nil {
+		return "", errgo.Notef(err, "cannot unmarshal digest")
+	}
+	return digest, nil
 }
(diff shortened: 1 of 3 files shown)


mhilton:
:+1: LGTM. No QA.

jrwren:
I would like to add one additional test which adds two of the same charms with different revision and assert that the digest with the highest revision is used for the response.

jrwren:
QA Done.

I ran charmd with this patch and did /etc/hosts and apache ssl things so that juju publish would use this charmd.

There is a race condition, which I don't think we care about. juju publish currently does not delay between retries when polling the charm store /charm-event and so juju publish can see the new revision of the charm, but with no digest and so it fails with an error "charm changed but not to local charm digest;
publishing race?". Running the exact same juju publish command again gives correct output as if the first command worked. This should be mitigated by the pending 500ms between retries juju-core patch.

:+1:

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/594/

internal/legacy/api.go
```diff
@@ -221,3 +223,61 @@ func (h *Handler) updateEntitySHA256(curl *charm.Reference) (string, error) {
 
 	return sum256, nil
 }
+
+func (h *Handler) serveCharmEvent(_ http.Header, req *http.Request) (interface{}, error) {
```
rogpeppe, line 229: Let's add a comment here describing the endpoint, so that we know what the code is trying to deal with.

internal/legacy/api.go
```diff
@@ -221,3 +223,61 @@ func (h *Handler) updateEntitySHA256(curl *charm.Reference) (string, error) {
...
+	response := make(map[string]*charm.EventResponse)
+	for _, url := range req.Form["charms"] {
+		c := &charm.EventResponse{}
+
+		// Ignore the digest part of the request.
+		if i := strings.Index(url, "@"); i != -1 {
+			url = url[:i]
+		}
+		// Intentionally not supporting long_keys query parameter which
```
rogpeppe: Full sentences are good...

```
// We intentionally do not implement the long_keys query parameter that the
// legacy charm store supported, as "juju publish" does not use it.
```

internal/legacy/api.go
```diff
@@ -221,3 +223,61 @@ func (h *Handler) updateEntitySHA256(curl *charm.Reference) (string, error) {
...
+			continue
+		}
+
+		// Retrieve the charm.
+		entity, err := h.store.FindEntity(id, "_id", "uploadtime", "extrainfo")
+		if err != nil {
+			if errgo.Cause(err) == params.ErrNotFound {
+				// The old API actually returned "entry not found"
+				// on *any* error, but it seems reasonable to be
+				// a little more descriptive for other errors.
```
rogpeppe, line 266: +1

internal/legacy/api.go
```diff
@@ -221,3 +223,61 @@ func (h *Handler) updateEntitySHA256(curl *charm.Reference) (string, error) {
...
+			}
+			c.Errors = []string{err.Error()}
+			continue
+		}
+
+		// Prepare the response part for this charm.
+		c.Kind = "published"
+		c.Revision = id.Revision
+		c.Time = entity.UploadTime.UTC().Format(time.RFC3339)
+		if digest, found := entity.ExtraInfo[params.BzrDigestKey]; found {
```
rogpeppe: Rather than duplicating the code from serveCharmInfo, perhaps
factor out the code to extract a digest into a separate function?

rogpeppe:
LGTM which a few minor suggestions. but it's a pity about the race. We could mitigate that ourselves by providing the capability to add extra-info in the charm upload PUT request and making charmload use that to add the digest.

Alternatively (and probably an easier solution for the time being) we could return a not-found error if there's no bzr digest found. I suggest that we do that, as the only reason we're serving charm-event is to provide this info for publish.

No QA.

internal/legacy/api.go
frankban, line 229: Done.

internal/legacy/api.go
frankban: Done.

internal/legacy/api.go
frankban: Done.

jujugui:
Test PASSed.
Refer to this link for build results: http://ci.jujugui.org:8080/job/charmstore/596/

internal/legacy/api.go
```diff
@@ -221,3 +222,94 @@ func (h *Handler) updateEntitySHA256(curl *charm.Reference) (string, error) {
...
+			// There are two possible reasons why an entity is found without a
+			// digest:
+			// 1) the entity has been recently added in the ingestion process,
+			//    but the extra-info has not been sent yet by "charmload";
+			// 2) there was an error while ingesting the entity.
+			// If the entity has been recently published, we assume case 1),
+			// and therefore we return a not found error, forcing
+			// "juju publish" to keep retrying and possibly succeed later.
+			// Otherwise, we return an error so that "juju publish" exits with
+			// an error and avoids an infinite loop.
```
urosj, line 287: Thank you.

internal/legacy/api_test.go
```diff
@@ -410,3 +410,225 @@ func (s *APISuite) addCharm(c *gc.C, charmName, curl string) (*charm.Reference,
...
+	s.updateUploadTime(c, url, time.Now().Add(-121*time.Second))
+	storetesting.AssertJSONCall(c, storetesting.JSONCallParams{
+		Handler:      s.srv,
+		URL:          "/charm-event?charms=cs:trusty/wordpress",
+		ExpectStatus: http.StatusOK,
+		ExpectBody: map[string]charm.EventResponse{
+			"cs:trusty/wordpress": {
+				Errors: []string{"digest not found: this can be due to an error while ingesting the entity"},
+			},
+		},
```
urosj, line 585: Thanks :)

urosj:
:+1: Did not do QA.

jrwren:
QA with same procedure as yesterday and did not encounter the race condition.
:+1:

frankban:
:shipit:

frankban:
Thanks a lot for the reviews! And thanks to Roger for the great idea for fixing the race condition.

jujugui:

PR #286, opened 2015-01-28
Add HTTP Basic Authorisation to search.

mhilton:
If a search call is performed with valid Basic Authorisation credentials then
the search will be in admin mode and will not check ACLs.

### DIFF
--- a/internal/v4/api_test.go
+++ b/internal/v4/api_test.go
@@ -23,6 +23,7 @@ import (
 	gc "gopkg.in/check.v1"
 	"gopkg.in/errgo.v1"
 	"gopkg.in/juju/charm.v4"
+	"gopkg.in/macaroon-bakery.v0/bakery"
 	"gopkg.in/macaroon-bakery.v0/bakery/checkers"
 	"gopkg.in/macaroon-bakery.v0/bakerytest"
 	"gopkg.in/macaroon-bakery.v0/httpbakery"
@@ -78,7 +79,7 @@ func (s *APISuite) SetUpTest(c *gc.C) {
 
 func newServer(c *gc.C, session *mgo.Session, si *charmstore.SearchIndex, config charmstore.ServerParams) (http.Handler, *charmstore.Store) {
 	db := session.DB("charmstore")
-	store, err := charmstore.NewStore(db, si, nil)
+	store, err := charmstore.NewStore(db, si, &bakery.NewServiceParams{})
 	c.Assert(err, gc.IsNil)
 	srv, err := charmstore.NewServer(db, si, config, map[string]charmstore.NewAPIHandlerFunc{"v4": v4.NewAPIHandler})
 	c.Assert(err, gc.IsNil)
--- a/internal/v4/auth.go
+++ b/internal/v4/auth.go
@@ -45,29 +45,16 @@ func (h *Handler) authorize(req *http.Request, acl []string) error {
 		}
 	}
 
-	// If basic auth credentials are presented, use them.
-	user, passwd, err := parseCredentials(req)
-	if err == nil {
-		if user != h.config.AuthUsername || passwd != h.config.AuthPassword {
-			return errgo.WithCausef(nil, params.ErrUnauthorized, "invalid user name or password")
-		}
-		return nil
-	}
-	if err != errNoCreds || h.store.Bakery == nil || h.config.IdentityLocation == "" {
-		return errgo.WithCausef(err, params.ErrUnauthorized, "authentication failed")
-	}
-
-	// No basic auth presented: use macaroon third-party caveat verification.
-	attrs, verr := httpbakery.CheckRequest(h.store.Bakery, req, nil, checkers.New())
+	auth, verr := h.checkRequest(req)
 	if verr == nil {
-		logger.Infof("authenticated with attrs %q", attrs)
-		if err := h.checkACLMembership(attrs, acl); err != nil {
+		logger.Infof("authenticated with auth: %q", auth)
+		if err := h.checkACLMembership(auth, acl); err != nil {
 			return errgo.WithCausef(err, params.ErrUnauthorized, "")
 		}
 		return nil
 	}
 	if _, ok := errgo.Cause(verr).(*bakery.VerificationError); !ok {
-		return errgo.Mask(verr)
+		return errgo.Mask(verr, errgo.Is(params.ErrUnauthorized))
 	}
 
 	// Macaroon verification failed: mint a new macaroon.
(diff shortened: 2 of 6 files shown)


mhilton:
## QA Instructions

Add some ACLs on charms in the charmstore and then perform a search on those charms prividing valid admin credentials and make sure they're shown.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/712/
Test PASSed.

internal/v4/api_test.go
```diff
@@ -78,7 +79,7 @@ func (s *APISuite) SetUpTest(c *gc.C) {
 
 func newServer(c *gc.C, session *mgo.Session, si *charmstore.SearchIndex, config charmstore.ServerParams) (http.Handler, *charmstore.Store) {
 	db := session.DB("charmstore")
-	store, err := charmstore.NewStore(db, si, nil)
+	store, err := charmstore.NewStore(db, si, &bakery.NewServiceParams{})
```
rogpeppe, line 82: I wonder if it would be better if NewStore took a value rather than a pointer.
If we keep it as a pointer, why is this change necessary?

internal/v4/auth.go
```diff
@@ -78,6 +79,32 @@ func (h *Handler) authorize(req *http.Request, acl []string) error {
 	return httpbakery.NewDischargeRequiredError(m, verr)
 }
 
+// checkRequest performs an httpbakery.CheckRequest and then filters out any
+// declared caveat attributes that were not requested.
+func (h *Handler) checkRequest(req *http.Request) (map[string]string, error) {
```
rogpeppe: Let's make this return a struct:

```
type authDeclaredAttrs struct {
    Username string
    Groups []string
}
```

That way we know exactly what we've got, and this function
becomes simpler too:

```
return &authDeclaredAttrs{
    Username: attrs[usernameAttr],
    Groups: strings.Fields(attrs[groupsAttr]),
}
```

internal/v4/auth.go
```diff
@@ -78,6 +79,32 @@ func (h *Handler) authorize(req *http.Request, acl []string) error {
...
+	attrs, err := httpbakery.CheckRequest(h.store.Bakery, req, nil, checkers.New())
+	if err != nil {
+		return nil, errgo.Mask(err, errgo.Any)
+	}
+	for k := range attrs {
```
frankban: A comment about why we are doing this sanity check?

internal/v4/search_test.go
```diff
@@ -51,39 +48,27 @@ var exportTestBundles = map[string]string{
...
-		},
-		URL:      storeURL("trusty/riak-67/meta/perm/read"),
-		Username: serverParams.AuthUsername,
-		Password: serverParams.AuthPassword,
-	})
-	err := s.ES.RefreshIndex(s.TestIndex)
+	// hide the riak charm
+	err := s.store.DB.BaseEntities().UpdateId(
+		charm.MustParseReference("cs:riak"),
+		bson.D{{"$set", map[string]map[string][]string{
```
frankban: Use mongodoc.ACL instead?

frankban:
:+1: with a couple of minor suggestions. Thank you!

internal/v4/search.go
```diff
@@ -137,17 +135,24 @@ func parseSearchParams(req *http.Request) (charmstore.SearchParams, error) {
 	return sp, nil
 }
 
+// groupsFromRequest finds the Read ACL users and groups that should be added to the
+// SearchParams.Groups list. It will also determine if the requestee has full admin rights.
 func (h *Handler) groupsFromRequest(sp *charmstore.SearchParams, req *http.Request) {
```
rogpeppe: I kind of feel that this doesn't need to be tied so closely to SearchParams.
It could be groupsFromRequest(req *http.Request) (isAdmin bool, groups []string, err error)
and which is closer to the contract implied by the name.

But when you look at it that way, it's not really much different from checkRequest,
which takes a request and returns some user info.

So perhaps better still might be to change checkRequest to check for admin credentials,
and add Admin to authDeclaredAttrs (authInfo might be a better name for that).

Then checkBasicAuth only needs to be called in one place.

rogpeppe:
Really nice stuff, thanks. I have a couple of suggestions above. What do you think?

internal/v4/api_test.go
mhilton, line 82: If the pointer is nil then a bakery is not created, where as if it is empty parameters it is created with default configuration. My tests require the latter.

internal/v4/search_test.go
rogpeppe: +1

internal/v4/auth.go
mhilton: done. I like that.

internal/v4/search_test.go
mhilton: done. thanks, code's much simpler.

internal/v4/search.go
mhilton: done. This has really simplified things.

internal/v4/auth.go
mhilton: This has changed quite a bit, but I've expanded the comment.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/714/
Test PASSed.

internal/v4/auth.go
```diff
@@ -82,6 +72,34 @@ func (h *Handler) authorize(req *http.Request, acl []string) error {
 	return httpbakery.NewDischargeRequiredError(m, cookiePath, verr)
 }
 
+// checkRequest checks for any authorisation tokens in the request either HTTP
```
rogpeppe: I'm not sure we need to describe all the implementation in this comment.
I'd limit it to describing the purpose of the function and how to use it.

```
// checkRequest checks the authorization provided by the
// given request. On success, it reports whether the request has admin
// rights; without admin rights it returns information on the authenticated
// user.
```

This is slightly more awkward to phrase because Admin isn't part of
the struct.

internal/v4/auth.go
```diff
@@ -45,29 +45,19 @@ func (h *Handler) authorize(req *http.Request, acl []string) error {
...
-		return errgo.WithCausef(err, params.ErrUnauthorized, "authentication failed")
-	}
-
-	// No basic auth presented: use macaroon third-party caveat verification.
-	attrs, verr := httpbakery.CheckRequest(h.store.Bakery, req, nil, checkers.New())
+	admin, attrs, verr := h.checkRequest(req)
 	if verr == nil {
-		logger.Infof("authenticated with attrs %q", attrs)
+		logger.Infof("authenticated with admin: %q, attrs %q", admin, attrs)
+		if admin {
```
rogpeppe: This is another place where having Admin in the struct returned by checkRequest
would make things simpler (the check could be done in checkACLMembership)

internal/v4/search.go
```diff
@@ -28,7 +25,14 @@ func (h *Handler) serveSearch(_ http.Header, req *http.Request) (interface{}, er
 	if err != nil {
 		return "", err
 	}
-	h.groupsFromRequest(&sp, req)
+	admin, attrs, err := h.checkRequest(req)
```
rogpeppe: check error here.

there should probably be a test for that.

rogpeppe:
LGTM with the error check fixed, and one suggestion about putting Admin inside the struct (if you did, authInfo or clientAuthorizationInfo would probably be a better name for it). Thanks!

internal/v4/auth.go
mhilton: I've updated this so admin is part of the struct.

internal/v4/auth.go
mhilton: done

internal/v4/search.go
mhilton: I don't actually want to error here, just carry on as if there is no logged in user, which there probably isn't, I have added a comment explaining this.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/715/
Test PASSed.

internal/v4/auth.go
```diff
@@ -119,16 +131,26 @@ const (
 	groupsAttr   = "groups"
 )
 
-func (h *Handler) checkACLMembership(attrs map[string]string, acl []string) error {
-	user := attrs[usernameAttr]
-	if user == "" {
+// authorizartion conatains authorization information extracted from the request.
```
rogpeppe: s/authorizartion/authorization/
s/the request/an HTTP request/ ?

internal/v4/search.go
```diff
@@ -28,7 +25,14 @@ func (h *Handler) serveSearch(_ http.Header, req *http.Request) (interface{}, er
 	if err != nil {
 		return "", err
 	}
-	h.groupsFromRequest(&sp, req)
+	// Ignore the error from checkRequest and carry on assuming no authorization was
+	// sent.
+	auth, _ := h.checkRequest(req)
```
rogpeppe: Seems like a good approach in general, but I'd
be inclined to log the error (Info level?)
e.g.
log.Infof("authorization failed on search request, granting no privileges: %v", err)

?

rogpeppe:
LGTM with one trivial suggestion, thanks!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/716/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/717/
Test PASSed.

mhilton:
:shipit:

jujugui:

PR #288, opened 2015-01-29
Introduce write ACL for entities.

frankban:
An entity can be written only by the user/group who
uploaded it.

Write the corresponding database migration.

Also update the meta/perm API endpoints accordingly.

### DIFF
--- a/internal/charmstore/migrations.go
+++ b/internal/charmstore/migrations.go
@@ -28,6 +28,9 @@ var migrations = []migration{{
 }, {
 	name:    "read acl creation",
 	migrate: populateReadACL,
+}, {
+	name:    "write acl creation",
+	migrate: populateWriteACL,
 }}
 
 // migration holds a migration function with its corresponding name.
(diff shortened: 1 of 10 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/718/
Test PASSed.

mhilton:
:+1: LGTM

internal/charmstore/migrations.go
```diff
@@ -181,7 +184,35 @@ func populateReadACL(db StoreDatabase) error {
 		if err := baseEntities.UpdateId(entity.URL, bson.D{{
 			"$set", bson.D{{"acls.read", readPerm}},
 		}}); err != nil {
-			return errgo.Notef(err, "cannot populate ACL for base entity %s", entity.URL)
+			return errgo.Notef(err, "cannot populate read ACL for base entity %s", entity.URL)
+		}
+		counter++
+	}
+	if err := iter.Close(); err != nil {
```
jrwren, line 191: won't this close conflict with the one above on line 176? It is now a double close.

Nevermind, I see from the docs: "Close is idempotent. That means it can be called repeatedly and will return the same result every time."

internal/charmstore/migrations.go
```diff
@@ -181,7 +184,35 @@ func populateReadACL(db StoreDatabase) error {
...
+func populateWriteACL(db StoreDatabase) error {
+	baseEntities := db.BaseEntities()
+	var entity mongodoc.BaseEntity
+	iter := baseEntities.Find(bson.D{{
+		"acls.write", bson.D{{"$exists", false}},
+	}, {
+		"user", bson.D{{"$ne", ""}},
+	}}).Select(bson.D{{"_id", 1}}).Iter()
+
+	defer iter.Close()
```
jrwren, line 208: Maybe log possible error from Close() like in the above populateReadACL function?

jrwren:
:+1: :+1:  A couple of comments, take them or leave them.

rogpeppe:
LGTM, thanks!

internal/charmstore/migrations.go
frankban, line 208: We already do that, don't we?

PR #291, opened 2015-02-12
Added migrations for the new promulgation model

mhilton:
Migrations will now populate the new PromulgatedURL in Entity and the
Promulgated flag in BaseEntity.

Note: Once the rest of the charmstore uses these fields additional
migrations should be added to remove the unnecessary Entities and
BaseEntities.

[Review on Reviewable](https://reviewable.io/reviews/juju/charmstore/291)

### DIFF
--- a/internal/charmstore/migrations.go
+++ b/internal/charmstore/migrations.go
@@ -31,6 +31,12 @@ var migrations = []migration{{
 }, {
 	name:    "write acl creation",
 	migrate: populateWriteACL,
+}, {
+	name:    "promulgated url creation",
+	migrate: populatePromulgatedURL,
+}, {
+	name:    "base entity promulgated flag creation",
+	migrate: populateBaseEntityPromulgated,
 }}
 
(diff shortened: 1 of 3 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/722/
Test PASSed.

urosj:
PS: the code should not land without extensive QA!

urosj:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/juju/charmstore/291)

---

<sup>**[internal/charmstore/migrations.go, line 39 [r1]](https://reviewable.io:443/reviews/juju/charmstore/291#-JkppGj6CwrijPmgqx7S)** ([raw file](https://github.com/juju/charmstore/blob/63b4f79fd6f0a18aaaeed8332777e117e5ec6ea6/internal/charmstore/migrations.go#L39)):</sup>
We should probably deprecate this migration as we've redone the ingestion model.

---

PR #292, opened 2015-02-12
Add store method to detemine promulgated url.

mhilton:
Added a function to the store to determine if an uploaded charm should
have a promulgated URL, and if so what that should be.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -188,6 +188,14 @@ type AddParams struct {
 	// Contents holds references to files inside the
 	// entity's archive blob.
 	Contents map[mongodoc.FileId]mongodoc.ZipFile
+
+	// PromulgatedURL holds the promulgated URL of the entity. If the entity
+	// is not promulgated this should be set to nil.
+	PromulgatedURL *charm.Reference
+
+	// PromulgatedRevision holds the revision number from the promulgated URL.
+	// If the entity is not promulgated this should be set to -1.
+	PromulgatedRevision int
 }
 
 // AddCharm adds a charm entities collection with the given
@@ -213,6 +221,8 @@ func (s *Store) AddCharm(c charm.Charm, p AddParams) (err error) {
 		CharmProvidedInterfaces: interfacesForRelations(c.Meta().Provides),
 		CharmRequiredInterfaces: interfacesForRelations(c.Meta().Requires),
 		Contents:                p.Contents,
+		PromulgatedURL:          p.PromulgatedURL,
+		PromulgatedRevision:     p.PromulgatedRevision,
 	}
 
 	// Check that we're not going to create a charm that duplicates
@@ -252,6 +262,7 @@ func (s *Store) insertEntity(entity *mongodoc.Entity) (err error) {
 			Read:  readPerm,
 			Write: writePerm,
 		},
+		Promulgated: entity.PromulgatedURL != nil,
 	}
 	err = s.DB.BaseEntities().Insert(baseEntity)
 	if err != nil && !mgo.IsDup(err) {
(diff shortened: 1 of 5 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/723/
Test PASSed.

internal/charmstore/store.go
```diff
@@ -188,6 +188,10 @@ type AddParams struct {
 	// Contents holds references to files inside the
 	// entity's archive blob.
 	Contents map[mongodoc.FileId]mongodoc.ZipFile
+
+	PromulgatedURL *charm.Reference
```
frankban, line 194: Please add docstrings for the two new fields.

internal/charmstore/store.go
```diff
@@ -213,6 +217,8 @@ func (s *Store) AddCharm(c charm.Charm, p AddParams) (err error) {
 		CharmProvidedInterfaces: interfacesForRelations(c.Meta().Provides),
 		CharmRequiredInterfaces: interfacesForRelations(c.Meta().Requires),
 		Contents:                p.Contents,
+		PromulgatedURL:          p.PromulgatedURL,
```
frankban, line 224: Do we need to add those two in AddBundle too?

frankban, line 224: It's done, nevermind.

internal/charmstore/store.go
```diff
@@ -252,6 +258,7 @@ func (s *Store) insertEntity(entity *mongodoc.Entity) (err error) {
 			Read:  readPerm,
 			Write: writePerm,
 		},
+		Promulgated: false,
```
frankban: Has been time since I red the spec, but shouldn't this be set to true if the given entity has a promulgated URL?

frankban: If this will be done later, please add a TODO, if I am wrong, just ignore me.

internal/charmstore/store.go
```diff
@@ -788,3 +797,44 @@ func (s *Store) SynchroniseElasticsearch() error {
 	}
 	return nil
 }
+
+//GetPromulgatedURL finds the promulgatedURL that should be used for
```
frankban: Please add a space at the beginning of each comment line.

internal/charmstore/store_test.go
```diff
@@ -1432,3 +1432,125 @@ func (s *StoreSuite) TestAddBundleArchiveIndexed(c *gc.C) {
 	c.Assert(err, gc.IsNil)
 	s.checkAddBundle(c, bundleArchive, true)
 }
+
+var getPromulgatedURLTests = []struct {
+	about    string
```
frankban: It would be nice to have a test with more than one previously promulgated charms, so that we check the last revision is used to calculate the new one.

internal/mongodoc/doc.go
```diff
@@ -124,6 +131,10 @@ type BaseEntity struct {
 	// the base entity. The permissions apply to all
 	// revisions.
 	ACLs ACL
+
+	// Promulgated indicates that charms with this base entity should be
```
frankban: ... that charms or bundles ...

internal/charmstore/store.go
frankban: ... finds the promulgated URL ...

frankban:
:+1: with minor suggestions.
Thanks for this good start on the promulgation work.

internal/charmstore/store.go
```diff
@@ -788,3 +797,44 @@ func (s *Store) SynchroniseElasticsearch() error {
...
+//this newly uploaded charm, if the charm should be promulgated,
+//othewise it returns nil. An error is returned if there is a problem
+//communicating with the storage.
+func (s *Store) GetPromulgatedURL(id *charm.Reference) (*charm.Reference, error) {
```
rogpeppe: I'm not entirely sure that this is the best place for this method. It seems like something that's probably only single use (by the archive POST code), and the current logic for incrementing the revision lives in v4/archive.go. What do you think about moving it there?

internal/charmstore/store.go
```diff
@@ -788,3 +797,44 @@ func (s *Store) SynchroniseElasticsearch() error {
...
+			// If there is no base entity then by definition it cannot be promulgated.
+			return nil, nil
+		}
+		return nil, errgo.Mask(err)
+	}
+	if !baseEntity.Promulgated {
+		return nil, nil
+	}
+	var entity mongodoc.Entity
+	err = s.DB.Entities().Find(bson.D{{"name", id.Name}, {"series", id.Series}, {"promulgated-url", bson.D{{"$exists", true}}}}).Sort("-promulgated-revision").One(&entity)
```
rogpeppe: Perhaps select the only fields that are required?

Also, I think I'd find this line easier to read if it was broken up a little.

internal/charmstore/store.go
```diff
@@ -788,3 +797,44 @@ func (s *Store) SynchroniseElasticsearch() error {
...
+	if err != nil {
```
rogpeppe: Perhaps:

```
promulgatedURL := *id
promulgatedURL.User = ""
promulgatedURL.Revision = 0
if err == nil {
    promulgatedURL.Revision = entity.PromulgatedRevision + 1
} else if err != mgo.ErrNotFound {
    return nil, errgo.Mask(err)
}
return &promulgatedURL, nil
```

?

internal/charmstore/store.go
rogpeppe, line 194: +1. Please mention the relevant invariants in those docstrings too.

internal/charmstore/store.go
rogpeppe: +1

internal/mongodoc/doc.go
rogpeppe: s/that/whether/
?

rogpeppe:
LGTM modulo frankban's comments and a couple of suggestions above. Thanks.

internal/charmstore/store.go
mhilton: I suspect it will be rare for us to be adding a base entity that is already promulgated, but I've put in the change anyway.

internal/charmstore/store.go

internal/charmstore/store.go
mhilton: I wasn't sure where to put it and chose on the store as it does mostly store things. I've moved it.

internal/charmstore/store.go

internal/charmstore/store.go

internal/charmstore/store_test.go
mhilton: done.

internal/mongodoc/doc.go
mhilton: done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/727/
Test PASSed.

frankban:
:+1: thank you!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/728/
Test PASSed.

mhilton:
:shipit:

jujugui:

PR #293, opened 2015-02-13
csclient package

frankban:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/724/
Test PASSed.

csclient/archive.go
```diff
@@ -0,0 +1,105 @@
...
+
+// ReadSeekCloser implements io.ReadSeeker and io.Closer.
+type ReadSeekCloser interface {
+	io.ReadSeeker
+	io.Closer
+}
+
+// openArchive is used to turn the current charm or bundle implementations
+// into ReadClosers for their corresponding archive.
+// It returns the corresponding archive ReadCloser, its SHA384 hash and size.
```
mhilton: It returns a ReadSeekCloser does the comment need to be updated?

csclient/archive.go
```diff
@@ -0,0 +1,105 @@
...
+	return &removeOnCloseFile{file}, nil
+}
+
+// removeOnCloseFile represents a file which is removed when closed.
+type removeOnCloseFile struct {
+	*os.File
+}
+
+func (r *removeOnCloseFile) Close() error {
+	r.File.Close()
```
mhilton, line 103: We are ignoring any error from close here, is that OK?

csclient/csclient.go
```diff
@@ -0,0 +1,487 @@
...
+const (
+	apiVersion = "v4"
+
+	// ServerURL holds the default location of the global charm store.
+	// An alternate location can be configured by changing the URL
+	// field in the Params struct.
+	// For live testing or QAing the application, a different charm store
+	// location should be used, for instance
+	// "https://api.staging.jujucharms.com".
+	ServerURL = "https://api.jujucharms.com"
```
mhilton: I believe the canonical location should be https://api.jujucharms.com/charmstore

csclient/csclient.go
```diff
@@ -0,0 +1,487 @@
...
+// Client represents the client side of a charm store.
+type Client struct {
+	params Params
+}
+
+// Params holds parameters for creating a new charm store client.
+type Params struct {
+	// URL holds the root endpoint URL of the charmstore,
+	// with no trailing slash, not including the version.
+	// For example http://charms.ubuntu.com
```
mhilton: Possibly update this example with the address of the new charmstore?

csclient/csclient.go
```diff
@@ -0,0 +1,487 @@
...
+	URL string
+
+	// User and Password hold the authentication credentials
+	// for the client. If User is empty, no credentials will be
+	// sent.
+	User     string
+	Password string
+
+	// HTTPClient holds the HTTP client to use when making
+	// requests to the store. If nil, http.DefaultClient will
```
mhilton: If this is nil httpbakery.NewHTTPClient() is used.

csclient/csclient.go
mitechie: +1

csclient/csclient.go
```diff
@@ -0,0 +1,487 @@
...
+	}
+	resp.Body.Close()
+	return nil
+}
+
+// Meta fetches metadata on the charm or bundle with the
+// given id. The result value provides a value
+// to be filled in with the result, which must be
+// a pointer to a struct containing members corresponding
+// to possible metadata include parameters (see http://tinyurl.com/nysdjly).
```
mhilton, line 219: The link is to a private document, do we want it in the public client API?

csclient/csclient.go
```diff
@@ -0,0 +1,487 @@
...
+//		ArchiveSize params.ArchiveSizeResponse
+//		ArchiveUploadTime params.ArchiveUploadTimeResponse
+//		Digest string `csclient:"extra-info/digest"`
+//	}
+//	id, err := client.Meta(id, &result)
+func (c *Client) Meta(id *charm.Reference, result interface{}) (*charm.Reference, error) {
+	if result == nil {
+		return nil, fmt.Errorf("expected valid result pointer, not nil")
+	}
+	resultv := reflect.ValueOf(result)
```
mhilton, line 244: could you use httprequest.Unmarshal to do this operation, to save having too much reflection in the system. It might change the semantics slightly though.

mhilton, line 244: On second thoughts that might not work exactly. It would be nice if httprequest had a method that did what we need here though. A lot of this code is similar to a lot in httprequest. There is probably a good argument for extracting the Field finding parts of httprequest into a package of it's own that could be used in both places.

mhilton:
:+1: with some review of comments.

csclient/archive.go
rogpeppe: maybe just s/ReadClosers/readers/ and let the type speak for itself.

csclient/archive.go
rogpeppe, line 103: I think that's OK. I don't think there's anything useful we can do if the file close fails.

csclient/archive.go
frankban: Done.

csclient/archive.go
frankban, line 103: I think that's probably fine.

csclient/csclient.go
frankban: Done.

csclient/csclient.go
frankban: Done.

csclient/csclient.go
frankban: Good catch, fixed.

csclient/csclient.go
frankban, line 219: The link is to the charm store API v4 spec. All the charmstore endpoints already have those links.
We should update all of them in a branch and point to the spec included in this repository in github.
But that's a separate task.

csclient/csclient.go
```diff
@@ -0,0 +1,487 @@
...
+		includes = append(includes, "include="+apiName)
+		results[apiName] = resultv.FieldByName(field.Name).Addr()
+	}
+	// We unmarshal into rawResult, then unmarshal each field
+	// separately into its place in the final result value.
+	// Note that we can't use params.MetaAnyResponse because
+	// that will unpack all the values inside the Meta field,
+	// but we want to keep them raw so that we can unmarshal
+	// them ourselves.
+	var rawResult struct {
```
rogpeppe, line 289: For the record, with httprequest this might look something like:

```
type metaAnyRequest struct {
    Include []string `httprequest:"include,form"`
    Result struct {
        Id *charm.Reference
        Meta map[string] json.RawMessage
    }       `httprequest:",body"`,
}
```

csclient/csclient_test.go
```diff
@@ -0,0 +1,1052 @@
...
+				},
+			},
+		})
+		var result interface{}
+		err := cl.Get("/foo", &result)
+		c.Assert(err, gc.ErrorMatches, test.expectError)
+	}
+}
+
+var hyphenateTests = []struct {
```
rogpeppe, line 694: These tests should be next to TestHyphenate, I think.

csclient/csclient.go
rogpeppe, line 244: +1 to extracting the fields function from httprequest.

I don't think that it's reasonable to make httprequest support this kind of request directly though,
unless it's made much more general (which is a possibility I'm still considering)

csclient/csclient.go
rogpeppe, line 219: +1

rogpeppe:
LGTM with a few suggestions above. Thanks!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/725/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/726/
Test PASSed.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

PR #299, opened 2015-02-23
Change to promulgation model.

mhilton:
Changes to use the new promulgation model. This includes:
- Updated insertion logic
- Updated query logic
- Updates to tests where necessary

### DIFF
--- a/csclient/csclient_test.go
+++ b/csclient/csclient_test.go
@@ -97,6 +97,7 @@ func (s *suite) TearDownTest(c *gc.C) {
 func (s *suite) TestDefaultServerURL(c *gc.C) {
 	// Add a charm used for tests.
 	err := s.store.AddCharmWithArchive(
+		charm.MustParseReference("~charmers/vivid/testing-wordpress-42"),
 		charm.MustParseReference("vivid/testing-wordpress-42"),
 		storetesting.Charms.CharmDir("wordpress"))
 	c.Assert(err, gc.IsNil)
@@ -148,8 +149,9 @@ var getTests = []struct {
 
 func (s *suite) TestGet(c *gc.C) {
 	ch := storetesting.Charms.CharmDir("wordpress")
-	url := charm.MustParseReference("utopic/wordpress-42")
-	err := s.store.AddCharmWithArchive(url, ch)
+	url := charm.MustParseReference("~charmers/utopic/wordpress-42")
+	purl := charm.MustParseReference("utopic/wordpress-42")
+	err := s.store.AddCharmWithArchive(url, purl, ch)
 	c.Assert(err, gc.IsNil)
 
 	for i, test := range getTests {
@@ -189,8 +191,9 @@ func (s *suite) TestGetArchive(c *gc.C) {
 	r, expectHash, expectSize := archiveHashAndSize(c, ch.Path)
 	r.Close()
 
-	url := charm.MustParseReference("utopic/wordpress-42")
-	err := s.store.AddCharmWithArchive(url, ch)
+	url := charm.MustParseReference("~charmers/utopic/wordpress-42")
+	purl := charm.MustParseReference("utopic/wordpress-42")
+	err := s.store.AddCharmWithArchive(url, purl, ch)
 	c.Assert(err, gc.IsNil)
 
 	rb, id, hash, size, err := s.client.GetArchive(url)
@@ -318,24 +321,26 @@ func (s *suite) TestUploadArchiveWithCharm(c *gc.C) {
 	path := storetesting.Charms.CharmArchivePath(c.MkDir(), "wordpress")
 
 	// Post the archive.
-	s.checkUploadArchive(c, path, "utopic/wordpress", "cs:utopic/wordpress-0")
+	s.checkUploadArchive(c, path, "~charmers/utopic/wordpress", "cs:~charmers/utopic/wordpress-0")
 
 	// Posting the same archive a second time does not change its resulting id.
-	s.checkUploadArchive(c, path, "utopic/wordpress", "cs:utopic/wordpress-0")
+	s.checkUploadArchive(c, path, "~charmers/utopic/wordpress", "cs:~charmers/utopic/wordpress-0")
 
 	// Posting a different archive to the same URL increases the resulting id
 	// revision.
 	path = storetesting.Charms.CharmArchivePath(c.MkDir(), "mysql")
-	s.checkUploadArchive(c, path, "utopic/wordpress", "cs:utopic/wordpress-1")
+	s.checkUploadArchive(c, path, "~charmers/utopic/wordpress", "cs:~charmers/utopic/wordpress-1")
 }
 
 func (s *suite) prepareBundleCharms(c *gc.C) {
 	// Add the charms required by the wordpress-simple bundle to the store.
 	err := s.store.AddCharmWithArchive(
+		charm.MustParseReference("~charmers/utopic/wordpress-42"),
 		charm.MustParseReference("utopic/wordpress-42"),
 		storetesting.Charms.CharmArchive(c.MkDir(), "wordpress"))
 	c.Assert(err, gc.IsNil)
 	err = s.store.AddCharmWithArchive(
+		charm.MustParseReference("~charmers/utopic/mysql-47"),
 		charm.MustParseReference("utopic/mysql-47"),
 		storetesting.Charms.CharmArchive(c.MkDir(), "mysql"))
 	c.Assert(err, gc.IsNil)
@@ -345,7 +350,7 @@ func (s *suite) TestUploadArchiveWithBundle(c *gc.C) {
 	s.prepareBundleCharms(c)
 	path := storetesting.Charms.BundleArchivePath(c.MkDir(), "wordpress-simple")
 	// Post the archive.
-	s.checkUploadArchive(c, path, "bundle/wordpress-simple", "cs:bundle/wordpress-simple-0")
+	s.checkUploadArchive(c, path, "~charmers/bundle/wordpress-simple", "cs:~charmers/bundle/wordpress-simple-0")
 }
 
 var uploadArchiveWithBadResponseTests = []struct {
@@ -422,7 +427,7 @@ func (s *suite) TestUploadArchiveWithServerError(c *gc.C) {
 	defer body.Close()
 
 	// Send an invalid hash so that the server returns an error.
-	url := charm.MustParseReference("trusty/wordpress")
+	url := charm.MustParseReference("~charmers/trusty/wordpress")
 	id, err := csclient.UploadArchive(s.client, url, body, hash+"mismatch", size)
 	c.Assert(id, gc.IsNil)
 	c.Assert(err, gc.ErrorMatches, "cannot post archive: cannot put archive blob: hash mismatch")
@@ -460,17 +465,17 @@ func archiveHashAndSize(c *gc.C, path string) (r csclient.ReadSeekCloser, hash s
 
 func (s *suite) TestUploadCharmDir(c *gc.C) {
 	ch := storetesting.Charms.CharmDir("wordpress")
-	id, err := s.client.UploadCharm(charm.MustParseReference("utopic/wordpress"), ch)
+	id, err := s.client.UploadCharm(charm.MustParseReference("~charmers/utopic/wordpress"), ch)
 	c.Assert(err, gc.IsNil)
-	c.Assert(id.String(), gc.Equals, "cs:utopic/wordpress-0")
+	c.Assert(id.String(), gc.Equals, "cs:~charmers/utopic/wordpress-0")
 	s.checkUploadCharm(c, id, ch)
 }
 
 func (s *suite) TestUploadCharmArchive(c *gc.C) {
 	ch := storetesting.Charms.CharmArchive(c.MkDir(), "wordpress")
-	id, err := s.client.UploadCharm(charm.MustParseReference("trusty/wordpress"), ch)
+	id, err := s.client.UploadCharm(charm.MustParseReference("~charmers/trusty/wordpress"), ch)
 	c.Assert(err, gc.IsNil)
-	c.Assert(id.String(), gc.Equals, "cs:trusty/wordpress-0")
+	c.Assert(id.String(), gc.Equals, "cs:~charmers/trusty/wordpress-0")
 	s.checkUploadCharm(c, id, ch)
 }
 
@@ -479,10 +484,10 @@ func (s *suite) TestUploadCharmErrorUploading(c *gc.C) {
 	// Note that the possible upload errors are already extensively exercised
 	// as part of the client.uploadArchive tests.
 	id, err := s.client.UploadCharm(
-		charm.MustParseReference("trusty/wordpress-42"),
+		charm.MustParseReference("~charmers/trusty/wordpress-42"),
 		storetesting.Charms.CharmDir("wordpress"),
 	)
-	c.Assert(err, gc.ErrorMatches, `revision specified in "cs:trusty/wordpress-42", but should not be specified`)
+	c.Assert(err, gc.ErrorMatches, `revision specified in "cs:~charmers/trusty/wordpress-42", but should not be specified`)
 	c.Assert(id, gc.IsNil)
 }
 
@@ -491,7 +496,7 @@ func (s *suite) TestUploadCharmErrorUnknownType(c *gc.C) {
 	unknown := struct {
 		charm.Charm
 	}{ch}
-	id, err := s.client.UploadCharm(charm.MustParseReference("trusty/wordpress"), unknown)
+	id, err := s.client.UploadCharm(charm.MustParseReference("~charmers/trusty/wordpress"), unknown)
 	c.Assert(err, gc.ErrorMatches, `cannot open charm archive: cannot get the archive for entity type .*`)
 	c.Assert(id, gc.IsNil)
 }
@@ -537,9 +542,9 @@ func (s *suite) checkUploadCharm(c *gc.C, id *charm.Reference, ch charm.Charm) {
 func (s *suite) TestUploadBundleDir(c *gc.C) {
 	s.prepareBundleCharms(c)
 	b := storetesting.Charms.BundleDir("wordpress-simple")
-	id, err := s.client.UploadBundle(charm.MustParseReference("bundle/wordpress-simple"), b)
+	id, err := s.client.UploadBundle(charm.MustParseReference("~charmers/bundle/wordpress-simple"), b)
 	c.Assert(err, gc.IsNil)
-	c.Assert(id.String(), gc.Equals, "cs:bundle/wordpress-simple-0")
+	c.Assert(id.String(), gc.Equals, "cs:~charmers/bundle/wordpress-simple-0")
 	s.checkUploadBundle(c, id, b)
 }
 
@@ -548,9 +553,9 @@ func (s *suite) TestUploadBundleArchive(c *gc.C) {
 	path := storetesting.Charms.BundleArchivePath(c.MkDir(), "wordpress-simple")
 	b, err := charm.ReadBundleArchive(path)
 	c.Assert(err, gc.IsNil)
-	id, err := s.client.UploadBundle(charm.MustParseReference("bundle/wp"), b)
+	id, err := s.client.UploadBundle(charm.MustParseReference("~charmers/bundle/wp"), b)
 	c.Assert(err, gc.IsNil)
-	c.Assert(id.String(), gc.Equals, "cs:bundle/wp-0")
+	c.Assert(id.String(), gc.Equals, "cs:~charmers/bundle/wp-0")
 	s.checkUploadBundle(c, id, b)
 }
 
@@ -559,10 +564,10 @@ func (s *suite) TestUploadBundleErrorUploading(c *gc.C) {
 	// Note that the possible upload errors are already extensively exercised
 	// as part of the client.uploadArchive tests.
 	id, err := s.client.UploadBundle(
-		charm.MustParseReference("wordpress-simple"),
+		charm.MustParseReference("~charmers/wordpress-simple"),
 		storetesting.Charms.BundleDir("wordpress-simple"),
 	)
-	c.Assert(err, gc.ErrorMatches, `no series specified in "cs:wordpress-simple"`)
+	c.Assert(err, gc.ErrorMatches, `no series specified in "cs:~charmers/wordpress-simple"`)
 	c.Assert(id, gc.IsNil)
 }
 
@@ -591,8 +596,9 @@ func (s *suite) checkUploadBundle(c *gc.C, id *charm.Reference, b charm.Bundle) 
 func (s *suite) TestDoAuthorization(c *gc.C) {
 	// Add a charm to be deleted.
 	ch := storetesting.Charms.CharmDir("wordpress")
-	url := charm.MustParseReference("utopic/wordpress-42")
-	err := s.store.AddCharmWithArchive(url, ch)
+	url := charm.MustParseReference("~charmers/utopic/wordpress-42")
+	purl := charm.MustParseReference("utopic/wordpress-42")
+	err := s.store.AddCharmWithArchive(url, purl, ch)
 	c.Assert(err, gc.IsNil)
 
 	// Check that when we use incorrect authorization,
@@ -604,12 +610,12 @@ func (s *suite) TestDoAuthorization(c *gc.C) {
 	})
 	req, err := http.NewRequest("DELETE", "", nil)
 	c.Assert(err, gc.IsNil)
-	_, err = client.Do(req, "/utopic/wordpress-42/archive")
+	_, err = client.Do(req, "/~charmers/utopic/wordpress-42/archive")
 	c.Assert(err, gc.ErrorMatches, "invalid user name or password")
 	c.Assert(errgo.Cause(err), gc.Equals, params.ErrUnauthorized)
 
 	// Check that it's still there.
-	err = client.Get("/utopic/wordpress-42/expand-id", nil)
+	err = client.Get("/~charmers/utopic/wordpress-42/expand-id", nil)
 	c.Assert(err, gc.IsNil)
 
 	// Then check that when we use the correct authorization,
@@ -621,7 +627,7 @@ func (s *suite) TestDoAuthorization(c *gc.C) {
 	})
 	req, err = http.NewRequest("DELETE", "", nil)
 	c.Assert(err, gc.IsNil)
-	resp, err := client.Do(req, "/utopic/wordpress-42/archive")
+	resp, err := client.Do(req, "/~charmers/utopic/wordpress-42/archive")
 	c.Assert(err, gc.IsNil)
 	resp.Body.Close()
 
@@ -749,8 +755,9 @@ func (s *suite) TestDo(c *gc.C) {
 	// Do is tested fairly comprehensively (but indirectly)
 	// in TestGet, so just a trivial smoke test here.
 	ch := storetesting.Charms.CharmDir("wordpress")
-	url := charm.MustParseReference("utopic/wordpress-42")
-	err := s.store.AddCharmWithArchive(url, ch)
+	url := charm.MustParseReference("~charmers/utopic/wordpress-42")
+	purl := charm.MustParseReference("utopic/wordpress-42")
+	err := s.store.AddCharmWithArchive(url, purl, ch)
 	c.Assert(err, gc.IsNil)
 	err = s.client.PutExtraInfo(url, map[string]interface{}{
 		"foo": "bar",
@@ -795,8 +802,9 @@ type embed struct{}
 
 func (s *suite) TestMeta(c *gc.C) {
 	ch := storetesting.Charms.CharmDir("wordpress")
-	url := charm.MustParseReference("utopic/wordpress-42")
-	err := s.store.AddCharmWithArchive(url, ch)
+	url := charm.MustParseReference("~charmers/utopic/wordpress-42")
+	purl := charm.MustParseReference("utopic/wordpress-42")
+	err := s.store.AddCharmWithArchive(url, purl, ch)
 	c.Assert(err, gc.IsNil)
 
 	// Put some extra-info.
@@ -898,15 +906,16 @@ func (s *suite) TestMeta(c *gc.C) {
 			continue
 		}
 		c.Assert(err, gc.IsNil)
-		c.Assert(id, jc.DeepEquals, url)
+		c.Assert(id, jc.DeepEquals, purl)
 		c.Assert(result, jc.DeepEquals, test.expectResult)
 	}
 }
 
 func (s *suite) TestPutExtraInfo(c *gc.C) {
 	ch := storetesting.Charms.CharmDir("wordpress")
-	url := charm.MustParseReference("utopic/wordpress-42")
-	err := s.store.AddCharmWithArchive(url, ch)
+	url := charm.MustParseReference("~charmers/utopic/wordpress-42")
+	purl := charm.MustParseReference("utopic/wordpress-42")
+	err := s.store.AddCharmWithArchive(url, purl, ch)
 	c.Assert(err, gc.IsNil)
 
 	// Put some info in.
@@ -1011,9 +1020,9 @@ func (s *suite) TestLog(c *gc.C) {
 
 func (s *suite) TestMacaroonAuthorization(c *gc.C) {
 	ch := storetesting.Charms.CharmArchive(c.MkDir(), "wordpress")
-
-	curl := charm.MustParseReference("utopic/wordpress-42")
-	err := s.store.AddCharmWithArchive(curl, ch)
+	curl := charm.MustParseReference("~charmers/utopic/wordpress-42")
+	purl := charm.MustParseReference("utopic/wordpress-42")
+	err := s.store.AddCharmWithArchive(curl, purl, ch)
 	c.Assert(err, gc.IsNil)
 
 	req, err := http.NewRequest("PUT", "", nil)
@@ -1023,7 +1032,7 @@ func (s *suite) TestMacaroonAuthorization(c *gc.C) {
 	body, err := json.Marshal([]string{"bob"})
 	c.Assert(err, gc.IsNil)
 
-	resp, err := s.client.DoWithBody(req, "/"+curl.Path()+"/meta/perm/read", httpbakery.SeekerBody(bytes.NewReader(body)))
+	resp, err := s.client.DoWithBody(req, "/"+purl.Path()+"/meta/perm/read", httpbakery.SeekerBody(bytes.NewReader(body)))
 	c.Assert(err, gc.IsNil)
 	defer resp.Body.Close()
 	c.Assert(resp.StatusCode, gc.Equals, http.StatusOK)
@@ -1034,7 +1043,7 @@ func (s *suite) TestMacaroonAuthorization(c *gc.C) {
 
 	var result struct{ IdRevision struct{ Revision int } }
 	// TODO 2015-01-23: once supported, rewrite the test using POST requests.
-	_, err = client.Meta(curl, &result)
+	_, err = client.Meta(purl, &result)
 	c.Assert(err, gc.ErrorMatches, `cannot get "/utopic/wordpress-42/meta/any\?include=id-revision": cannot get discharge from ".*": cannot discharge: no discharge`)
 
 	s.discharge = func(cond, arg string) ([]checkers.Caveat, error) {
@@ -1062,7 +1071,7 @@ func (s *suite) TestMacaroonAuthorization(c *gc.C) {
 			return fmt.Errorf("stopping interaction")
 		}})
 
-	_, err = client.Meta(curl, &result)
+	_, err = client.Meta(purl, &result)
 	c.Assert(err, gc.ErrorMatches, `cannot get "/utopic/wordpress-42/meta/any\?include=id-revision": cannot get discharge from ".*": cannot start interactive session: stopping interaction`)
 	c.Assert(result.IdRevision.Revision, gc.Equals, curl.Revision)
 }
--- a/internal/charmstore/elasticsearch.go
+++ b/internal/charmstore/elasticsearch.go
@@ -10,7 +10,7 @@ var (
 	esMapping = mustParseJSON(esMappingJSON)
 )
 
-const esSettingsVersion = 4
+const esSettingsVersion = 5
 
 func mustParseJSON(s string) interface{} {
 	var j json.RawMessage
@@ -67,6 +67,11 @@ const esMappingJSON = `
             "include_in_all" : false
           }
         }
+      },
+      "PromulgatedURL" : {
+        "type" : "string",
+        "index": "not_analyzed",
+        "index_options" : "docs"
       },
       "BaseURL" : {
         "type" : "string",
--- a/internal/charmstore/search.go
+++ b/internal/charmstore/search.go
@@ -77,8 +77,22 @@ func (s *Store) UpdateSearch(r *charm.Reference) error {
 	if deprecatedSeries[r.Series] {
 		return nil
 	}
+	var query *mgo.Query
+	if r.User == "" {
+		query = s.DB.Entities().Find(bson.D{
+			{"name", r.Name},
+			{"series", r.Series},
+			{"promulgated-url", bson.D{{"$exists", true}}},
+		}).Sort("-promulgated-revision")
+	} else {
+		query = s.DB.Entities().Find(bson.D{
+			{"user", r.User},
+			{"name", r.Name},
+			{"series", r.Series},
+		}).Sort("-revision")
+	}
 	var entity mongodoc.Entity
-	if err := s.DB.Entities().Find(bson.M{"user": r.User, "name": r.Name, "series": r.Series}).Sort("-revision").One(&entity); err != nil {
+	if err := query.One(&entity); err != nil {
 		if err == mgo.ErrNotFound {
 			return errgo.WithCausef(nil, params.ErrNotFound, "entity not found %s", r)
 		}
@@ -173,7 +187,7 @@ func (si *SearchIndex) search(sp SearchParams) (SearchResult, error) {
 		return SearchResult{}, nil
 	}
 	q := createSearchDSL(sp)
-	q.Fields = append(q.Fields, "URL")
+	q.Fields = append(q.Fields, "URL", "PromulgatedURL")
 	esr, err := si.Search(si.Index, typeName, q)
 	if err != nil {
 		return SearchResult{}, errgo.Mask(err)
@@ -184,7 +198,11 @@ func (si *SearchIndex) search(sp SearchParams) (SearchResult, error) {
 		Results:    make([]*charm.Reference, 0, len(esr.Hits.Hits)),
 	}
 	for _, h := range esr.Hits.Hits {
-		ref, err := charm.ParseReference(h.Fields.GetString("URL"))
+		url := h.Fields.GetString("PromulgatedURL")
+		if url == "" {
+			url = h.Fields.GetString("URL")
+		}
+		ref, err := charm.ParseReference(url)
 		if err != nil {
 			return SearchResult{}, errgo.Notef(err, "invalid result %q", h.Fields.GetString("URL"))
 		}
@@ -580,6 +598,9 @@ func nameFilter(value string) elasticsearch.Filter {
 // ownerFilter generates a filter that will match against the
 // owner taken from the URL.
 func ownerFilter(value string) elasticsearch.Filter {
+	if value == "" {
+		return elasticsearch.ExistsFilter("PromulgatedURL")
+	}
 	return elasticsearch.QueryFilter{
 		Query: elasticsearch.MatchQuery{
 			Field: "User",
--- a/internal/charmstore/search_test.go
+++ b/internal/charmstore/search_test.go
@@ -67,8 +67,7 @@ var charmDownloadCounts = map[string]int{
 
 func (s *StoreSearchSuite) TestSuccessfulExport(c *gc.C) {
 	for name, ref := range exportTestCharms {
-		var entity *mongodoc.Entity
-		err := s.store.DB.Entities().FindId(ref).One(&entity)
+		entity, err := s.store.FindEntity(charm.MustParseReference(ref))
 		c.Assert(err, gc.IsNil)
 		var actual json.RawMessage
 		err = s.store.ES.GetDocument(s.TestIndex, typeName, s.store.ES.getID(entity.URL), &actual)
@@ -77,6 +76,8 @@ func (s *StoreSearchSuite) TestSuccessfulExport(c *gc.C) {
 		readACLs := []string{params.Everyone}
 		if r.User != "" {
 			readACLs = append(readACLs, r.User)
+		} else {
+			readACLs = append(readACLs, "charmers")
 		}
 		if r.Name == "riak" {
 			readACLs = []string{"quux"}
@@ -92,18 +93,18 @@ func (s *StoreSearchSuite) TestSuccessfulExport(c *gc.C) {
 
 func (s *StoreSearchSuite) TestNoExportDeprecated(c *gc.C) {
 	charmArchive := storetesting.Charms.CharmDir("mysql")
-	url := charm.MustParseReference("cs:saucy/mysql-4")
-	err := s.store.AddCharmWithArchive(url, charmArchive)
+	url := charm.MustParseReference("cs:~charmers/saucy/mysql-4")
+	err := s.store.AddCharmWithArchive(url, nil, charmArchive)
 	c.Assert(err, gc.IsNil)
 
 	var entity *mongodoc.Entity
-	err = s.store.DB.Entities().FindId("cs:trusty/mysql-7").One(&entity)
+	err = s.store.DB.Entities().FindId("cs:~charmers/trusty/mysql-7").One(&entity)
 	c.Assert(err, gc.IsNil)
 	present, err := s.store.ES.HasDocument(s.TestIndex, typeName, s.store.ES.getID(entity.URL))
 	c.Assert(err, gc.IsNil)
 	c.Assert(present, gc.Equals, true)
 
-	err = s.store.DB.Entities().FindId("cs:saucy/mysql-4").One(&entity)
+	err = s.store.DB.Entities().FindId("cs:~charmers/saucy/mysql-4").One(&entity)
 	c.Assert(err, gc.IsNil)
 	present, err = s.store.ES.HasDocument(s.TestIndex, typeName, s.store.ES.getID(entity.URL))
 	c.Assert(err, gc.IsNil)
@@ -112,25 +113,25 @@ func (s *StoreSearchSuite) TestNoExportDeprecated(c *gc.C) {
 
 func (s *StoreSearchSuite) TestExportOnlyLatest(c *gc.C) {
 	charmArchive := storetesting.Charms.CharmDir("wordpress")
-	url := charm.MustParseReference("cs:precise/wordpress-24")
-	err := s.store.AddCharmWithArchive(url, charmArchive)
+	url := charm.MustParseReference("cs:~charmers/precise/wordpress-24")
+	err := s.store.AddCharmWithArchive(url, nil, charmArchive)
 	c.Assert(err, gc.IsNil)
 	var expected, old *mongodoc.Entity
 	var actual json.RawMessage
-	err = s.store.DB.Entities().FindId("cs:precise/wordpress-23").One(&old)
+	err = s.store.DB.Entities().FindId("cs:~charmers/precise/wordpress-23").One(&old)
 	c.Assert(err, gc.IsNil)
-	err = s.store.DB.Entities().FindId("cs:precise/wordpress-24").One(&expected)
+	err = s.store.DB.Entities().FindId("cs:~charmers/precise/wordpress-24").One(&expected)
 	c.Assert(err, gc.IsNil)
 	err = s.store.ES.GetDocument(s.TestIndex, typeName, s.store.ES.getID(old.URL), &actual)
 	c.Assert(err, gc.IsNil)
-	doc := SearchDoc{Entity: expected, ReadACLs: []string{params.Everyone}}
+	doc := SearchDoc{Entity: expected, ReadACLs: []string{params.Everyone, "charmers"}}
 	c.Assert(string(actual), jc.JSONEquals, doc)
 }
 
 func (s *StoreSearchSuite) TestExportSearchDocument(c *gc.C) {
 	var entity *mongodoc.Entity
 	var actual json.RawMessage
-	err := s.store.DB.Entities().FindId("cs:precise/wordpress-23").One(&entity)
+	err := s.store.DB.Entities().FindId("cs:~charmers/precise/wordpress-23").One(&entity)
 	c.Assert(err, gc.IsNil)
 	doc := SearchDoc{Entity: entity, TotalDownloads: 4000}
 	err = s.store.ES.update(&doc)
@@ -144,8 +145,14 @@ func (s *StoreSearchSuite) addCharmsToStore(c *gc.C, store *Store) {
 	for name, ref := range exportTestCharms {
 		charmArchive := storetesting.Charms.CharmDir(name)
 		url := charm.MustParseReference(ref)
+		var purl *charm.Reference
+		if url.User == "" {
+			purl = new(charm.Reference)
+			*purl = *url
+			url.User = "charmers"
+		}
 		charmArchive.Meta().Categories = strings.Split(name, "-")
-		err := store.AddCharmWithArchive(url, charmArchive)
+		err := store.AddCharmWithArchive(url, purl, charmArchive)
 		c.Assert(err, gc.IsNil)
 		for i := 0; i < charmDownloadCounts[name]; i++ {
 			err := store.IncrementDownloadCounts(url)
@@ -155,8 +162,14 @@ func (s *StoreSearchSuite) addCharmsToStore(c *gc.C, store *Store) {
 	for name, ref := range exportTestBundles {
 		bundleArchive := storetesting.Charms.BundleDir(name)
 		url := charm.MustParseReference(ref)
+		var purl *charm.Reference
+		if url.User == "" {
+			purl = new(charm.Reference)
+			*purl = *url
+			url.User = "charmers"
+		}
 		bundleArchive.Data().Tags = strings.Split(name, "-")
-		err := store.AddBundleWithArchive(url, bundleArchive)
+		err := store.AddBundleWithArchive(url, purl, bundleArchive)
 		c.Assert(err, gc.IsNil)
 		for i := 0; i < charmDownloadCounts[name]; i++ {
 			err := store.IncrementDownloadCounts(url)
@@ -166,9 +179,10 @@ func (s *StoreSearchSuite) addCharmsToStore(c *gc.C, store *Store) {
 	baseEntity, err := store.FindBaseEntity(charm.MustParseReference("cs:riak"))
 	c.Assert(err, gc.IsNil)
 	baseEntity.ACLs.Read = []string{"quux"}
-	err = store.DB.BaseEntities().UpdateId("cs:riak", baseEntity)
+	err = store.DB.BaseEntities().UpdateId(baseEntity.URL, baseEntity)
 	c.Assert(err, gc.IsNil)
-	store.UpdateSearch(charm.MustParseReference(exportTestCharms["riak"]))
+	err = store.UpdateSearch(charm.MustParseReference(exportTestCharms["riak"]))
+	c.Assert(err, gc.IsNil)
 }
 
 var searchTests = []struct {
@@ -436,8 +450,9 @@ func (s *StoreSearchSuite) TestLimitTestSearch(c *gc.C) {
 
 func (s *StoreSearchSuite) TestPromulgatedRank(c *gc.C) {
 	charmArchive := storetesting.Charms.CharmDir("varnish")
-	url := charm.MustParseReference("cs:trusty/varnish-1")
-	s.store.AddCharmWithArchive(url, charmArchive)
+	url := charm.MustParseReference("cs:~charmers/trusty/varnish-1")
+	purl := charm.MustParseReference("cs:trusty/varnish-1")
+	s.store.AddCharmWithArchive(url, purl, charmArchive)
 	s.store.ES.Database.RefreshIndex(s.TestIndex)
 	sp := SearchParams{
 		Filters: map[string][]string{
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -71,11 +71,6 @@ func NewStore(db *mgo.Database, si *SearchIndex, bakeryParams *bakery.NewService
 		}
 		s.Bakery = bsvc
 	}
-	go func() {
-		if err := s.syncSearch(); err != nil {
-			logger.Errorf("Cannot populate elasticsearch: %v", err)
-		}
-	}()
 	return s, nil
 }
 
@@ -95,6 +90,9 @@ func (s *Store) ensureIndexes() error {
 	}, {
 		s.DB.Entities(),
 		mgo.Index{Key: []string{"uploadtime"}},
+	}, {
+		s.DB.Entities(),
+		mgo.Index{Key: []string{"promulgated-url"}, Unique: true, Sparse: true},
 	}, {
 		s.DB.BaseEntities(),
 		mgo.Index{Key: []string{"public"}},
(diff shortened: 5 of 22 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/756/
Test FAILed.

csclient/csclient_test.go
```diff
@@ -1011,9 +1020,9 @@ func (s *suite) TestLog(c *gc.C) {
 
 func (s *suite) TestMacaroonAuthorization(c *gc.C) {
 	ch := storetesting.Charms.CharmArchive(c.MkDir(), "wordpress")
-
-	curl := charm.MustParseReference("utopic/wordpress-42")
-	err := s.store.AddCharmWithArchive(curl, ch)
+	curl := charm.MustParseReference("~charmers/utopic/wordpress-42")
```
rogpeppe, line 1023: Any particular reason that this test needs a promulgated charm URL?

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,73 @@ func populateWriteACL(db StoreDatabase) error {
...
+		{"baseurl", 1},
+		{"blobhash", 1},
+		{"extrainfo.bzr-owner", 1},
+	}).Sort("revision").Iter()
+	var entity mongodoc.Entity
+	for iter.Next(&entity) {
+		logger.Debugf("URL: %s", entity.URL)
+		var user string
+		if err := json.Unmarshal([]byte(entity.ExtraInfo["bzr-owner"]), &user); err != nil {
+			return errgo.Notef(err, "cannot unmarshal user from extra-info")
```
rogpeppe: I wonder if we should really abort the migration if we can't find the bzr owner. Perhaps print a log message and use "unknown" (or something else suitable) as the user in this case, perhaps?

csclient/csclient_test.go
mhilton, line 1023: Only because we can't add a charm without one.

internal/charmstore/migrations.go
mhilton: If we use unknown then there won't be a matching owned entity so the update will just fail a bit later.

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,73 @@ func populateWriteACL(db StoreDatabase) error {
...
+			},
+			bson.D{{"$set", bson.D{
+				{"promulgated-url", entity.URL},
+				{"promulgated-revision", entity.URL.Revision},
+			}}},
+		)
+		if err != nil {
+			return errgo.Notef(err, "cannot update entity for promulgated charm or bundle %q", entity.URL)
+		}
+		_, err = db.BaseEntities().UpdateAll(
```
rogpeppe: I'm not sure I believe that this will work, as discussed online.

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +831,692 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+			"bzr-owner": []byte(`"who"`),
+		},
+	})
+	c.Assert(err, gc.IsNil)
+
+	s.insertBaseEntity(c, baseURL(id1), &mongodoc.ACL{})
+	s.insertBaseEntity(c, baseURL(id2), &mongodoc.ACL{})
+	s.insertBaseEntity(c, baseURL(id3), &mongodoc.ACL{})
+	s.insertBaseEntity(c, baseURL(id4), &mongodoc.ACL{})
+	s.insertBaseEntity(c, baseURL(id5), &mongodoc.ACL{})
```
rogpeppe: I suspect if you change insertBaseEntity to unset the promulgated field, then the tests will fail.

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +831,692 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+		Name:     id2.Name,
+		User:     id2.User,
+		Series:   id2.Series,
+		Revision: id2.Revision,
+		ExtraInfo: map[string][]byte{
+			"bzr-owner": []byte(`"who"`),
+		},
+	})
+	c.Assert(err, gc.IsNil)
+	err = s.db.Entities().Insert(&entity2{
```
rogpeppe: Perhaps we could have a helper function for filling out the URL-derived fields in an entity so we don't have to duplicate so much similar code?

something like:

```
 func newEntity2(e entity2) *entity2 {
     e.BaseURL = baseURL(e.URL)
     e.Name = e.URL.Name
     etc
     return &e
}
```

(yeah, I know "new" isn't quite right here, but I can't think of a better name)

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/757/
Test FAILed.

internal/charmstore/search_test.go
```diff
@@ -77,6 +76,8 @@ func (s *StoreSearchSuite) TestSuccessfulExport(c *gc.C) {
 		readACLs := []string{params.Everyone}
 		if r.User != "" {
```
rogpeppe, line 77: Why not just have ~charmers in the exportTestCharms URLs ?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/758/
Test PASSed.

internal/charmstore/search_test.go
mhilton, line 77: I wanted to reduce the amount of changes to existing tests, this way they could be left alone in many places. Also it gives a nice way to specify which ones should be promulgated. exportTestCharms is used in a lot of places.

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,96 @@ func populateWriteACL(db StoreDatabase) error {
 	logger.Infof("%d base entities updated", counter)
 	return nil
 }
+
+func populatePromulgatedEntities(db StoreDatabase) error {
+	baseEntities := db.BaseEntities()
+	entities := db.Entities()
+
+	// 1. Update All Base Entities to have a promulgated value of the correct form.
```
rogpeppe: s/All Base Entities/all base entities/

csclient/csclient_test.go
mhilton, line 1023: OK, I get you now. I wanted to leave existing tests testing the thing they already tested. In this case it might have been better to change it as it makes little difference.

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,96 @@ func populateWriteACL(db StoreDatabase) error {
...
+	_, err := baseEntities.UpdateAll(
+		bson.D{{"promulgated", true}},
+		bson.D{{"$set", bson.D{{"promulgated", 1}}}},
+	)
```
rogpeppe: Since we're updating all promulgated entries later, why not just set promulgated to -1 for all base entities (ignoring all current values)?

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,96 @@ func populateWriteACL(db StoreDatabase) error {
...
+			bson.D{{"promulgated", false}},
+			bson.D{{"promulgated", bson.D{{"$exists", false}}}},
+		}}},
+		bson.D{{"$set", bson.D{{"promulgated", -1}}}},
+	)
+	if err != nil {
+		return errgo.Notef(err, "cannot set promulgated to initial value")
+	}
+
+	// 2. Update entities with users with their promulgated URL.
```
rogpeppe: // Note that we order the query results by revision, so when we've traversed all the entities,
// owner[name] will hold the owner of the latest revision of each promulgated
// charm with the given name.

?

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,96 @@ func populateWriteACL(db StoreDatabase) error {
...
+			bson.D{
+				{"user", user},
+				{"name", entity.URL.Name},
+				{"series", entity.URL.Series},
+				{"blobhash", entity.BlobHash},
+			},
+			bson.D{{"$set", bson.D{
+				{"promulgated-url", entity.URL},
+				{"promulgated-revision", entity.URL.Revision},
+			}}},
```
rogpeppe: Nice.

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,96 @@ func populateWriteACL(db StoreDatabase) error {
...
+		var user string
+		if err := json.Unmarshal([]byte(entity.ExtraInfo["bzr-owner"]), &user); err != nil {
+			return errgo.Notef(err, "cannot unmarshal user from extra-info")
+		}
+		if user == "" {
+			return errgo.Newf("no user for %q", entity.URL)
+		}
+		owners[entity.URL.Name] = user
+		err := entities.Update(
+			bson.D{
```
rogpeppe: Do we have to worry about several entities with different revisions but the same blobhash here? If that happened, ISTM that this might cause a revision to be lost, as we'll be updating twice with the same query parameters, and the second update will change the same document. Perhaps this won't happen in practice, but maybe we could mitigate the possibility by only updating entries that have an unset promulgated-url field?

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,96 @@ func populateWriteACL(db StoreDatabase) error {
...
+	// 1. Update All Base Entities to have a promulgated value of the correct form.
+	_, err := baseEntities.UpdateAll(
+		bson.D{{"promulgated", true}},
+		bson.D{{"$set", bson.D{{"promulgated", 1}}}},
+	)
+	if err != nil {
+		return errgo.Notef(err, "cannot set promulgated to initial value")
+	}
+	_, err = baseEntities.UpdateAll(
+		bson.D{{"$or", []interface{}{
```
frankban: Can't this be just a []bson.D?

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,96 @@ func populateWriteACL(db StoreDatabase) error {
...
+			bson.D{{"promulgated", false}},
+			bson.D{{"promulgated", bson.D{{"$exists", false}}}},
+		}}},
+		bson.D{{"$set", bson.D{{"promulgated", -1}}}},
+	)
+	if err != nil {
+		return errgo.Notef(err, "cannot set promulgated to initial value")
```
frankban: cannot set promulgated to initial false value?

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,96 @@ func populateWriteACL(db StoreDatabase) error {
...
+	baseEntities := db.BaseEntities()
+	entities := db.Entities()
+
+	// 1. Update All Base Entities to have a promulgated value of the correct form.
+	_, err := baseEntities.UpdateAll(
+		bson.D{{"promulgated", true}},
+		bson.D{{"$set", bson.D{{"promulgated", 1}}}},
+	)
+	if err != nil {
+		return errgo.Notef(err, "cannot set promulgated to initial value")
```
frankban: cannot set promulgated to initial true value?

internal/charmstore/migrations.go
frankban: +1

internal/charmstore/migrations.go
mhilton: I agree and it's fixed.

internal/charmstore/migrations_test.go
mhilton: You were right, It no longer does though.

internal/charmstore/migrations_test.go
mhilton: Perhaps, there are functions for doing similar things already. As we get more migrations we will probably need a more general approach.

internal/charmstore/migrations.go
mhilton: I think the migration should be idempotent as much as possible, although they'll be changed before the existings entities are deleted so that will probably work.

internal/charmstore/migrations.go
mhilton: good idea, I'll add that.

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,96 @@ func populateWriteACL(db StoreDatabase) error {
...
+	_, err = entities.RemoveAll(bson.D{{"user", ""}})
+	if err != nil {
+		return errgo.Notef(err, "cannot remove old entities")
+	}
+
+	// 6. Delete the redundant base entities.
+	_, err = baseEntities.RemoveAll(bson.D{{"user", ""}})
+	if err != nil {
+		return errgo.Notef(err, "cannot remove old base entities")
+	}
```
frankban: Thanks for the comments, this is very readable now!

internal/charmstore/migrations.go
mhilton: Yeah, to err on the side a caution I might do that, if it does happen then it will error with the second update, so it won't fail silently.

internal/charmstore/migrations.go
mhilton: yes good idea.

internal/charmstore/search_test.go
rogpeppe, line 77: Hmm, I think I'd prefer it to be more direct and hold a slice of structs each holding a URL and whether it's promulgated.

internal/charmstore/migrations_test.go
frankban: +1
makeEntity2?

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +831,743 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+		Public:      true,
+		Promulgated: 0,
+	})
+	c.Assert(err, gc.IsNil)
+
+	// Start the server.
+	err = s.newServer(c)
+	c.Assert(err, gc.IsNil)
+
+	// Ensure entities have been updated correctly.
```
frankban: What updates do we expect here?

internal/charmstore/migrations.go
mhilton: done, along with @rogpeppe suggestion.

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +831,743 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+		Revision:  id1.Revision,
+		ExtraInfo: map[string][]byte{},
+	})
+	c.Assert(err, gc.IsNil)
+
+	err = s.newServer(c)
+	c.Assert(err, gc.ErrorMatches, "database migration failed: error executing migration: populate promulgated entities: cannot unmarshal user from extra-info: unexpected end of JSON input")
+}
+
+func (s *migrationsSuite) TestPopulatePromulgatedEntitiesNoBlanklBzrOwner(c *gc.C) {
```
frankban: TestPopulatePromulgatedEntitiesBlanklBzrOwner

internal/charmstore/migrations.go
mhilton: no longer needed after @rogpeppe suggestion.

internal/charmstore/migrations.go
mhilton: The code has been removed, so it isn't needed anymore.

internal/charmstore/migrations.go
mhilton: done.

internal/charmstore/migrations.go
mhilton: done.

internal/charmstore/migrations.go
mhilton: done.

internal/charmstore/migrations.go
mhilton: It turns out this isn't a great idea, If there are no updates left and it runs the migration we lose all the promulgations. So we need to be a bit selective.

internal/charmstore/migrations.go
rogpeppe: I'm not sure it will, will it? The promulgated revision will be different, so I don't think it'll trigger any duplicate index errors.

internal/charmstore/migrations.go
rogpeppe: Ooh, good point. Perhaps just set it to -1 if promulgated is unset (or true or false too if we're worrying about that)

internal/charmstore/migrations.go
rogpeppe: +1

internal/charmstore/store.go
```diff
@@ -133,17 +136,23 @@ func (s *Store) putArchive(archive blobstore.ReadSeekCloser) (blobName, blobHash
 // also adds the charm archive to the blob store.
 // This method is provided principally so that
 // tests can easily create content in the store.
-func (s *Store) AddCharmWithArchive(url *charm.Reference, ch charm.Charm) error {
+func (s *Store) AddCharmWithArchive(url, purl *charm.Reference, ch charm.Charm) error {
```
rogpeppe, line 137: Please document the purl parameter.

internal/charmstore/store.go
```diff
@@ -151,17 +160,23 @@ func (s *Store) AddCharmWithArchive(url *charm.Reference, ch charm.Charm) error
 // also adds the charm archive to the blob store.
 // This method is provided principally so that
 // tests can easily create content in the store.
-func (s *Store) AddBundleWithArchive(url *charm.Reference, b charm.Bundle) error {
+func (s *Store) AddBundleWithArchive(url, purl *charm.Reference, b charm.Bundle) error {
```
rogpeppe, line 164: Ditto

internal/charmstore/store.go
```diff
@@ -259,6 +274,10 @@ func (s *Store) insertEntity(entity *mongodoc.Entity) (err error) {
 		readPerm = []string{params.Everyone, entity.User}
 		writePerm = []string{entity.User}
 	}
+	promulgated := 0
```
rogpeppe: s/0/-1/ ?

internal/charmstore/store.go
```diff
@@ -270,7 +289,7 @@ func (s *Store) insertEntity(entity *mongodoc.Entity) (err error) {
 			Read:  readPerm,
 			Write: writePerm,
 		},
-		Promulgated: entity.PromulgatedURL != nil,
+		Promulgated: promulgated,
```
rogpeppe: I'm almost tempted to suggest something like:

```
type IntBool bool
func (b IntBool) GetBSON() (interface{}, error) { 
    if b {
        return 1, nil
    }
    return 0, nil
}
func (b *IntBool) SetBSON(raw bson.Raw) error {
     var x int
     if err := raw.Unmarshal(&x); err != nil { 
        etc
}
```

So we can use normal conditions in the Go code.

internal/charmstore/store.go
```diff
@@ -325,29 +344,92 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
 	}
-	last := 0
-	for _, doc := range docs {
-		if matchURL(doc.URL, url) {
-			docs[last] = doc
-			last++
+	return docs, nil
+}
+
+//FindBestEntity finds the entity that provides the preferred match to
```
rogpeppe: Please add a space after the //s.

internal/charmstore/store.go
```diff
@@ -325,29 +344,92 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
+	"lucid":   1000,
+	"precise": 1001,
+	"trusty":  1002,
+	"quantal": 1,
+	"raring":  2,
+	"saucy":   3,
+	"utopic":  4,
+}
+
+//QueryEntity creates a mgo.Query object that can be used to get
```
rogpeppe: Spaces after // please.

internal/charmstore/store.go
```diff
@@ -325,29 +344,92 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
+//entities matching the given URL.
+func (s *Store) QueryEntity(url *charm.Reference) *mgo.Query {
+	if url.User != "" && url.Series != "" && url.Revision != -1 {
+		return s.DB.Entities().FindId(url)
+	}
+	if url.Series != "" && url.Revision != -1 {
+		return s.DB.Entities().Find(bson.D{{"promulgated-url", url}})
+	}
+	q := bson.D{{"name", url.Name}}
```
rogpeppe: Perhaps save a few allocations in this fairly common case by preallocating the slice?

   q := make(bson.D, 0, 4)
   q = append(q, bson.DocElem{"name", url.Name})

internal/charmstore/store.go
```diff
@@ -325,29 +344,92 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
 // If any fields are specified, only those fields will be
 // populated in the returned entities.
 func (s *Store) FindEntities(url *charm.Reference, fields ...string) ([]*mongodoc.Entity, error) {
```
rogpeppe, line 346: Please document that if url.User is empty, it searches for promulgated charms.

rogpeppe, line 346: (and similar for FindBestEntity and QueryEntity below)

internal/charmstore/store.go
```diff
@@ -325,29 +344,92 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
+	"trusty":  1002,
+	"quantal": 1,
+	"raring":  2,
+	"saucy":   3,
+	"utopic":  4,
+}
+
+//QueryEntity creates a mgo.Query object that can be used to get
+//entities matching the given URL.
+func (s *Store) QueryEntity(url *charm.Reference) *mgo.Query {
```
rogpeppe: I wonder whether EntityQuery might be better here as it's returning a query, not querying an entity as such. Or QueryEntities perhaps.

internal/charmstore/store.go
```diff
@@ -378,32 +468,60 @@ func selectFields(query *mgo.Query, fields []string) *mgo.Query {
 	return query
 }
 
+//UpdateEntity applies the provided update to the entity described by url.
```
rogpeppe: space after //

internal/v4/archive.go
```diff
@@ -133,13 +136,20 @@ func (h *Handler) servePostArchive(id *charm.Reference, w http.ResponseWriter, r
 		})
 	}
 
+	// Find the ID for the promulgated version of this charm if it should have one.
```
rogpeppe: I don't understand this comment.

internal/v4/archive.go
```diff
@@ -162,7 +175,21 @@ func (h *Handler) servePutArchive(id *charm.Reference, w http.ResponseWriter, re
 	if req.ContentLength == -1 {
 		return badRequestf(nil, "Content-Length not specified")
 	}
-	if err := h.addBlobAndEntity(id, req.Body, hash, req.ContentLength); err != nil {
+	if err := req.ParseForm(); err != nil {
```
rogpeppe: Don't we call ParseForm as a matter of course somewhere else?

internal/v4/relations.go
```diff
@@ -134,8 +135,12 @@ func (h *Handler) getRelatedIfaceResponses(
 					return nil, err
 				}
 				// Build the response.
+				url := entity.URL
+				if entity.PromulgatedURL != nil {
```
rogpeppe: I had to think quite hard about this. Perhaps a comment might be justified:

```
// Always use a promulgated URL if one is available.
```

?

internal/v4/relations.go
```diff
@@ -176,8 +181,8 @@ func (h *Handler) metaBundlesContaining(entity *mongodoc.Entity, id *charm.Refer
 	var entities []mongodoc.Entity
 	if err := h.store.DB.Entities().
 		Find(bson.D{{"bundlecharms", &searchId}}).
-		Select(bson.D{{"_id", 1}, {"bundlecharms", 1}}).
-		Sort("baseurl", "series", "-revision").
+		Select(bson.D{{"_id", 1}, {"bundlecharms", 1}, {"promulgated-url", 1}}).
+		Sort("name", "-promulgated-revision", "-revision").
```
rogpeppe, line 181: We're sorting by promulgated-revision just for consistency of results?

internal/v4/relations.go
```diff
@@ -222,8 +227,12 @@ func (h *Handler) metaBundlesContaining(entity *mongodoc.Entity, id *charm.Refer
 		if err != nil {
 			return nil, errgo.Notef(err, "cannot retrieve bundle metadata")
 		}
+		url := e.URL
+		if e.PromulgatedURL != nil {
```
rogpeppe: I've seen this quite a bit.
How about a method on Entity?

```
// PreferredURL returns the preferred form of the URL to
// to refer to the given entity    
func (e *Entity) PreferredURL() *charm.Reference {
     if e.PromulgatedURL != nil { 
         return e.PromulgatedURL
     }
     return e.URL
 }
```

rogpeppe:
This is heroic stuff. Looks fantastic from where I'm sitting now, modulo some
remarks above. I skimmed a lot of the tests though - I'll try to have another
look when my eyeballs aren't falling out.

internal/v4/relations.go
mhilton: I had one of those and took it out. often the URL that is chosen is based on the request that is made rather than what is available.

internal/charmstore/store.go
mhilton: This turned out to be a great idea, it showed up a few potential bugs in the process.

internal/charmstore/store.go
mhilton: This should have been -1 (there were a couple of these) they are sorted now.

internal/charmstore/migrations_test.go
mhilton: done

internal/charmstore/migrations.go
mhilton: It turns out that to cope with parallel executions we have to ignore any case where it cannot find what to update, but this one won't overwrite anything once it's set.

internal/charmstore/migrations_test.go
mhilton: none, it was a copy and paste error. fixed now.

internal/charmstore/store.go
mhilton, line 137: done.

internal/charmstore/store.go
mhilton, line 164: done.

internal/charmstore/store.go

internal/charmstore/store.go

internal/charmstore/store.go
mhilton: done, except I used 3 because if you had all 4 you wouldn't be going this way.

internal/charmstore/store.go
mhilton, line 346: done.

internal/charmstore/store.go
mhilton: I went with EntitiesQuery, done.

internal/charmstore/store.go
mhilton: done.

internal/v4/archive.go
mhilton: I have attempted to clarify.

internal/v4/archive.go
mhilton: It turns out we do, also we were using the result higher up in the function. Redundant code removed.

internal/v4/relations.go
mhilton: I have added a function for this as suggested below.

internal/v4/relations.go
mhilton, line 181: yes, it also boosts promulgated charms to the front of the list.

internal/v4/relations.go
mhilton: I have added a similar method.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/761/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/762/
Test PASSed.

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +848,600 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+		Series:    id1.Series,
+		Revision:  id1.Revision,
+		ExtraInfo: map[string][]byte{},
+	}, entity2...)
+
+	err := s.newServer(c)
+	c.Assert(err, gc.ErrorMatches, "database migration failed: error executing migration: populate promulgated entities: cannot unmarshal user from extra-info: unexpected end of JSON input")
+}
+
+func (s *migrationsSuite) TestPopulatePromulgatedEntitiesNoBlankBzrOwner(c *gc.C) {
```
frankban: I still see NoBlankBzrOwner here.

internal/charmstore/migrations_test.go
```diff
@@ -822,49 +1471,104 @@ func (s *migrationsSuite) checkBaseEntitiesCount(c *gc.C, expectCount int) {
...
 
-func (s *migrationsSuite) insertBaseEntity(c *gc.C, id *charm.Reference, acls *mongodoc.ACL) {
-	entity := &mongodoc.BaseEntity{
-		URL:    id,
-		Name:   id.Name,
-		User:   id.User,
-		Public: true,
+// Standard entity versions. These are lists of fields that can be used with
+// insertEntity to create database documents from previous deployments.
+var (
```
frankban: This is cool, I wonder if it would be more clear if those were named entity3fields, entity2fields etc.
Same for the baseEntityX below.

internal/charmstore/store.go
```diff
@@ -323,31 +350,97 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
+	"quantal": 1,
+	"raring":  2,
+	"saucy":   3,
+	"utopic":  4,
+}
+
+// EntitiesQuery creates a mgo.Query object that can be used to get
+// entities matching the given URL. If the given URL has no user then
+// the produced query will only match promulgated entities.
+func (s *Store) EntitiesQuery(url *charm.Reference) *mgo.Query {
```
frankban, line 410: Cool! This seems to handle all the cases.
Perhaps some comments/examples would help the reader here.

internal/charmstore/store.go
```diff
@@ -323,31 +350,97 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
+	if url.User != "" && url.Series != "" && url.Revision != -1 {
+		return s.DB.Entities().FindId(url)
```
frankban, line 413: Find a specific owned entity, for instance ~who/utopic/django-42.

internal/charmstore/store.go
```diff
@@ -323,31 +350,97 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
+	}
+	if url.Series != "" && url.Revision != -1 {
+		return s.DB.Entities().Find(bson.D{{"promulgated-url", url}})
```
frankban, line 417: Find a specific promulgated entity, for instance utopic/django-42.

internal/charmstore/store.go
```diff
@@ -323,31 +350,97 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
+	}
+	q := make(bson.D, 0, 3)
+	q = append(q, bson.DocElem{"name", url.Name})
+	if url.User != "" {
```
frankban, line 422: Find all the entities matching the URL.
If the URL user is empty, only search the promulgated entities.
?

internal/charmstore/store.go
```diff
@@ -323,31 +350,97 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
+	return docs, nil
+}
+
+// FindBestEntity finds the entity that provides the preferred match to
+// the given URL. If any fields are specified, only those fields will be
+// populated in the returned entities. If the given URL has no user then
+// only promulgated entities will be queried.
+func (s *Store) FindBestEntity(url *charm.Reference, fields ...string) (*mongodoc.Entity, error) {
+	if len(fields) > 0 {
+		// make sure we have all the fields we need to make a decision
```
frankban: // Make sure we have all the fields we need to make a decision.

internal/charmstore/store.go
```diff
@@ -355,7 +448,15 @@ func (s *Store) FindEntities(url *charm.Reference, fields ...string) ([]*mongodo
 // If any fields are specified, only those fields will be populated in the
 // returned base entity.
 func (s *Store) FindBaseEntity(url *charm.Reference, fields ...string) (*mongodoc.BaseEntity, error) {
-	query := s.DB.BaseEntities().FindId(baseURL(url))
+	var query *mgo.Query
+	if url.User == "" {
+		query = s.DB.BaseEntities().Find(bson.D{{"name", url.Name}, {"promulgated", 1}})
+	} else {
+		baseId := *url
```
frankban: Why not using baseUrl here?

``` go
query := s.DB.BaseEntities().FindId(baseURL(url))
```

internal/charmstore/store.go
```diff
@@ -378,32 +479,60 @@ func selectFields(query *mgo.Query, fields []string) *mgo.Query {
...
+	return nil
+}
+
+// UpdateBaseEntity applies the provided update to the base entity of url.
+func (s *Store) UpdateBaseEntity(url *charm.Reference, update interface{}) error {
+	var q bson.D
+	if url.User == "" {
+		q = bson.D{{"name", url.Name}, {"promulgated", 1}}
+	} else {
+		baseId := *url
```
frankban: baseUrl here too?

internal/charmstore/store.go
```diff
@@ -495,12 +624,9 @@ func (s *Store) OpenBlob(id *charm.Reference) (r blobstore.ReadSeekCloser, size
 // for the entity with the given id and its hash. It returns a params.ErrNotFound
 // error if the entity does not exist.
 func (s *Store) BlobNameAndHash(id *charm.Reference) (name, hash string, err error) {
-	var entity mongodoc.Entity
-	if err := s.DB.Entities().
-		FindId(id).
-		Select(bson.D{{"blobname", 1}, {"blobhash", 1}}).
-		One(&entity); err != nil {
-		if err == mgo.ErrNotFound {
+	entity, err := s.FindEntity(id, "blobname", "blobhash")
```
frankban, line 616: Thanks for this clean up.

internal/charmstore/store_test.go
```diff
@@ -44,16 +45,17 @@ func (s *StoreSuite) checkAddCharm(c *gc.C, ch charm.Charm, addToES bool) {
...
 	store, err := NewStore(s.Session.DB("juju_test"), &SearchIndex{s.ES, s.TestIndex}, nil)
 	c.Assert(err, gc.IsNil)
-	url := charm.MustParseReference("cs:precise/wordpress-23")
+	url := charm.MustParseReference("cs:~charmers/precise/wordpress-23")
+	purl := charm.MustParseReference("cs:precise/wordpress-23")
 
 	// Add the charm to the store.
 	beforeAdding := time.Now()
-	err = store.AddCharmWithArchive(url, ch)
+	err = store.AddCharmWithArchive(url, purl, ch)
```
frankban, line 58: Should we also test adding charms/bundles without a promulgated URL?

internal/charmstore/store_test.go
```diff
@@ -300,7 +311,13 @@ func (s *StoreSuite) testURLFinding(c *gc.C, check func(store *Store, expand *ch
 		c.Assert(err, gc.IsNil)
 		urls := mustParseReferences(test.inStore)
 		for _, url := range urls {
-			err := store.AddCharmWithArchive(url, wordpress)
+			var purl *charm.Reference
+			if url.User == "" {
+				purl = new(charm.Reference)
```
frankban, line 332: Is this line required?

internal/charmstore/store_test.go
```diff
@@ -313,24 +330,42 @@ func (s *StoreSuite) testURLFinding(c *gc.C, check func(store *Store, expand *ch
 
 func (s *StoreSuite) TestFindEntities(c *gc.C) {
 	s.testURLFinding(c, func(store *Store, expand *charm.Reference, expect []*charm.Reference) {
-		// check FindEntities works when just retrieving the id.
-		gotEntities, err := store.FindEntities(expand, "_id")
+		// check FindEntities works when just retrieving the id, and promulgated id.
```
frankban: // Check FindEntities works when just retrieving the id and promulgated id.

internal/charmstore/store_test.go
```diff
@@ -1436,3 +1516,208 @@ func (s *StoreSuite) TestAddBundleArchiveIndexed(c *gc.C) {
...
+		Series:              "trusty",
+		Name:                "wordpress",
+		Revision:            13,
+		PromulgatedURL:      charm.MustParseReference("trusty/wordpress-13"),
+		PromulgatedRevision: 13,
+	})
+	c.Assert(err, gc.IsNil)
+
+	for i, test := range findBestEntityTests {
+		c.Logf("test %d. %s", i, test.url)
```
frankban: test %d: %s [shrug]

internal/charmstore/store_test.go
mhilton, line 332: yes because otherwise it is nil.

internal/charmstore/store_test.go
```diff
@@ -1436,3 +1516,208 @@ func (s *StoreSuite) TestAddBundleArchiveIndexed(c *gc.C) {
...
+	expectURL: "~mickey/precise/wordpress-12",
+}, {
+	url:       "mysql",
+	expectErr: "entity not found",
+}}
+
+func (s *StoreSuite) TestFindBestEntity(c *gc.C) {
+	store, err := NewStore(s.Session.DB("juju_test"), nil, nil)
+	c.Assert(err, gc.IsNil)
+	err = store.DB.Entities().Insert(&mongodoc.Entity{
```
frankban, line 1614: I'd add a promulgated precise wordpress here, to check wordpress still returns the trusty one, and that precise/wordpress returns the precise one.
I'd also add some tests for bundles.
Exercising the not found error would be nice too here.

frankban, line 1614: Ignore the request for testing the not found error case: it's already done.

internal/charmstore/store_test.go
```diff
@@ -1436,3 +1516,208 @@ func (s *StoreSuite) TestAddBundleArchiveIndexed(c *gc.C) {
...
+var updateEntityTests = []struct {
+	url       string
+	expectErr string
+}{{
+	url: "~charmers/trusty/wordpress-10",
+}, {
+	url: "trusty/wordpress-4",
+}, {
+	url:       "~charmers/precise/wordpress-10",
+	expectErr: "cannot update \"cs:~charmers/precise/wordpress-10\": not found",
```
frankban: Let's use ` and avoid escaping quotes.

internal/charmstore/store_test.go
```diff
@@ -1436,3 +1516,208 @@ func (s *StoreSuite) TestAddBundleArchiveIndexed(c *gc.C) {
...
+var updateBaseEntityTests = []struct {
+	url       string
+	expectErr string
+}{{
+	url: "~charmers/trusty/wordpress-10",
+}, {
+	url: "trusty/wordpress-4",
+}, {
+	url:       "~charmers/precise/mysql-10",
+	expectErr: "cannot update base entity for \"cs:~charmers/precise/mysql-10\": not found",
```
frankban: Let's use ` and avoid escaping quotes.

internal/v4/api.go
```diff
@@ -266,36 +263,39 @@ func (h *Handler) entityExists(id *charm.Reference, req *http.Request) (bool, er
...
 }
 
 func (h *Handler) entityQuery(id *charm.Reference, selector map[string]int, req *http.Request) (interface{}, error) {
-	var val mongodoc.Entity
-	err := h.store.DB.Entities().
-		Find(bson.D{{"_id", id}}).
-		Select(selector).
-		One(&val)
-	if err == mgo.ErrNotFound {
+	fields := make([]string, 0, len(selector))
```
frankban: What do you think about adding a fieldsFromSelector helper function?

internal/v4/archive.go
```diff
@@ -113,6 +113,9 @@ func (h *Handler) servePostArchive(id *charm.Reference, w http.ResponseWriter, r
 	if id.Revision != -1 {
 		return badRequestf(nil, "revision specified, but should not be specified")
 	}
+	if id.User == "" {
```
frankban, line 116: So we need a follow up branch to fix ingestion, correct?

internal/v4/archive.go
```diff
@@ -133,13 +136,21 @@ func (h *Handler) servePostArchive(id *charm.Reference, w http.ResponseWriter, r
 		})
 	}
 
+	// Get the promulgated URL for this entity, If the user is not the
```
frankban: I am not sure I can parse this comment.

internal/v4/archive.go
```diff
@@ -162,7 +176,18 @@ func (h *Handler) servePutArchive(id *charm.Reference, w http.ResponseWriter, re
 	if req.ContentLength == -1 {
 		return badRequestf(nil, "Content-Length not specified")
 	}
-	if err := h.addBlobAndEntity(id, req.Body, hash, req.ContentLength); err != nil {
+	promulgatedURL := req.Form.Get("promulgated")
```
frankban, line 183: Why do we have this logic in PUT and not in POST?
Also, why a promulgated bool is not sufficient in this case?
Could you add a comment with answers to the questions above?

frankban:
This is great stuff Martin, thank you!
I have some comments and questions, but it looks very nice in general.
Of course this need to be intensely QAed.

internal/charmstore/elasticsearch.go
```diff
@@ -68,6 +68,11 @@ const esMappingJSON = `
           }
         }
       },
+      "PromulgatedURL" : {
+        "type" : "string",
+        "index": "not_analyzed",
+        "index_options" : "docs"
+      },
```
urosj, line 75: The URL supports ngrams. It's probably ok to not include ngrams in PromulgatedURL as we'll find the name by the URL and boost up due to promulgation. Just a thought on the schema.

internal/charmstore/migrations_test.go
```diff
@@ -822,49 +1471,104 @@ func (s *migrationsSuite) checkBaseEntitiesCount(c *gc.C, expectCount int) {
...
+//     Name
+//     Revision
+//     Series
+//     User
+//     PromulgatedRevision
+//
+// After insertion any name in fields will be removed from the inserted
+// document.
+func (s *migrationsSuite) insertEntity(c *gc.C, entity *mongodoc.Entity, fields ...string) {
+	if entity.BaseURL == nil {
```
rogpeppe: We could probably just set these values unconditionally. Is there anywhere we want to knowingly insert an internally inconsistent entity?

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +848,600 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+	err := s.newServer(c)
+	c.Assert(err, gc.ErrorMatches, "database migration failed: error executing migration: populate promulgated entities: cannot unmarshal user from extra-info: unexpected end of JSON input")
+}
+
+func (s *migrationsSuite) TestPopulatePromulgatedEntitiesNoBlankBzrOwner(c *gc.C) {
+	s.patchMigrations(c, getMigrations("populate promulgated entities"))
+	id1 := charm.MustParseReference("trusty/django-42")
+	s.insertEntity(c, &mongodoc.Entity{
+		URL:      id1,
+		BaseURL:  baseURL(id1),
```
rogpeppe: d

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +848,600 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+		BlobHash: "django-1",
+		Name:     id1.Name,
+		User:     id1.User,
+		Series:   id1.Series,
+		Revision: id1.Revision,
```
rogpeppe: delete the above 4 lines.

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +848,600 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+		Public:      true,
+		ACLs:        mongodoc.ACL{},
+		Promulgated: mongodoc.False,
+	})
+}
+
+func (s *migrationsSuite) TestPopulatePromulgatedEntitiesNoBzrOwner(c *gc.C) {
+	s.patchMigrations(c, getMigrations("populate promulgated entities"))
+	id1 := charm.MustParseReference("trusty/django-42")
+	s.insertEntity(c, &mongodoc.Entity{
```
rogpeppe: delete redundant lines

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +848,600 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+		User:                id5.User,
+		Name:                id5.Name,
+		Revision:            id5.Revision,
+		Series:              id5.Series,
+		PromulgatedRevision: -1,
+		ExtraInfo: map[string][]byte{
+			"bzr-owner": []byte(`"ace"`),
+		},
+	})
+	s.checkEntity(c, &mongodoc.Entity{
```
rogpeppe: If, instead of putting the auto-field-fill logic inside insertEntity, we put it into
a mkEntity function, then these places could benefit from it too.

e.g.
   func mkEntity(e mongodoc.Entity) *mongodoc.Entity

```
s.checkEntity(c, mkEntity(mongodoc.Entity{
    URL: id6,
    BlobHash: "django-2", 
    PromulgatedURL: id7,
    ExtraInfo: map[string][]byte{ ...},
```

   }))

internal/charmstore/migrations_test.go
```diff
@@ -371,10 +408,13 @@ func (s *migrationsSuite) TestDenormalizeEntityIdsSomeUpdates(c *gc.C) {
...
 	})
 	s.checkEntity(c, &mongodoc.Entity{
-		URL:     id2,
-		BaseURL: baseURL(id2),
-		Name:    "django",
-		Size:    2,
+		URL:      id2,
+		BaseURL:  baseURL(id2),
+		Name:     "django",
+		User:     "",
```
rogpeppe: Remove the zero-valued fields (User, Revision, Series) here?

rogpeppe: BTW why don't we see user=dalek, revision=47 here?

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,93 @@ func populateWriteACL(db StoreDatabase) error {
...
+	iter := entities.Find(bson.D{{"user", ""}}).Select(bson.D{
+		{"_id", 1},
+		{"baseurl", 1},
+		{"blobhash", 1},
+		{"extrainfo.bzr-owner", 1},
+	}).Sort("name", "revision").Iter()
+	var entity mongodoc.Entity
+	for iter.Next(&entity) {
+		var user string
+		if err := json.Unmarshal([]byte(entity.ExtraInfo["bzr-owner"]), &user); err != nil {
```
urosj: I am quite sure there are some charms that don't have this due to the bugs in December. So we will never iterate over all entities.

internal/charmstore/migrations.go
urosj: Agree, nice mongo fu :)

internal/charmstore/migrations_test.go
urosj: Nice, Martin. +1 on both, this being really cool solution and adding "fields" postfix.

internal/charmstore/search.go
```diff
@@ -77,8 +77,22 @@ func (s *Store) UpdateSearch(r *charm.Reference) error {
...
+	} else {
+		query = s.DB.Entities().Find(bson.D{
+			{"user", r.User},
+			{"name", r.Name},
+			{"series", r.Series},
+		}).Sort("-revision")
+	}
 	var entity mongodoc.Entity
-	if err := s.DB.Entities().Find(bson.M{"user": r.User, "name": r.Name, "series": r.Series}).Sort("-revision").One(&entity); err != nil {
+	if err := query.One(&entity); err != nil {
```
urosj, line 95: +1 on this pattern.

internal/charmstore/search.go
```diff
@@ -580,6 +598,9 @@ func nameFilter(value string) elasticsearch.Filter {
 // ownerFilter generates a filter that will match against the
 // owner taken from the URL.
 func ownerFilter(value string) elasticsearch.Filter {
```
urosj, line 600: We should probably add another filter in search query, ie: promulgated=1 - not removing this one for compatibility. "owner=" semantics is somehow "broken" with new way of looking at promulgated charms.

internal/charmstore/store.go
```diff
@@ -133,17 +136,26 @@ func (s *Store) putArchive(archive blobstore.ReadSeekCloser) (blobName, blobHash
 // also adds the charm archive to the blob store.
 // This method is provided principally so that
 // tests can easily create content in the store.
-func (s *Store) AddCharmWithArchive(url *charm.Reference, ch charm.Charm) error {
+//
+// If purl is not nil then the charm will also be
+// available at the promulgated url specified.
+func (s *Store) AddCharmWithArchive(url, purl *charm.Reference, ch charm.Charm) error {
```
urosj, line 137: And at this point we are 100% sure that promulgated in base entity is set to 1, right?

internal/charmstore/store.go
```diff
@@ -259,6 +280,10 @@ func (s *Store) insertEntity(entity *mongodoc.Entity) (err error) {
 		readPerm = []string{params.Everyone, entity.User}
 		writePerm = []string{entity.User}
 	}
+	promulgated := mongodoc.False
```
rogpeppe: I don't think you need this variable - you should be able do:

```
Promulgated: entity.PromulgatedURL != nil,
```

below.

internal/charmstore/store.go
```diff
@@ -323,31 +350,97 @@ func (s *Store) FindEntity(url *charm.Reference, fields ...string) (*mongodoc.En
...
+// entities matching the given URL. If the given URL has no user then
+// the produced query will only match promulgated entities.
+func (s *Store) EntitiesQuery(url *charm.Reference) *mgo.Query {
+	if url.User != "" && url.Series != "" && url.Revision != -1 {
+		return s.DB.Entities().FindId(url)
+	}
+	if url.Series != "" && url.Revision != -1 {
+		return s.DB.Entities().Find(bson.D{{"promulgated-url", url}})
+	}
+	q := make(bson.D, 0, 3)
```
rogpeppe, line 420: I still think that 4 is a good number here - I still want this
function to be efficient even if we are specifying all the fields.

internal/charmstore/migrations_test.go
```diff
@@ -793,6 +848,600 @@ func (s *migrationsSuite) TestPopulateWriteACLSomeUpdates(c *gc.C) {
...
+		ACLs:        mongodoc.ACL{},
+		Promulgated: mongodoc.False,
+	})
+	s.checkBaseEntity(c, &mongodoc.BaseEntity{
+		URL:         baseURL(id3),
+		User:        id3.User,
+		Name:        id3.Name,
+		Public:      true,
+		ACLs:        mongodoc.ACL{},
+		Promulgated: mongodoc.True,
```
rogpeppe: or just "true".

internal/charmstore/store_test.go
```diff
@@ -211,12 +217,17 @@ func assertBaseEntity(c *gc.C, store *Store, url *charm.Reference) {
...
-		User:   url.User,
-		Name:   url.Name,
-		Public: true,
-		ACLs:   expectACLs,
+		URL:         url,
+		User:        url.User,
+		Name:        url.Name,
+		Public:      true,
+		ACLs:        expectACLs,
+		Promulgated: p,
```
rogpeppe: Promulgated: mongodoc.IntBool(promulgated),

internal/v4/archive.go
mhilton, line 116: Ingestion will need to change, yes.

internal/mongodoc/doc.go
```diff
@@ -230,3 +242,35 @@ type Migration struct {
...
+		*b = True
+	case -1:
+		*b = False
+	default:
+		return errgo.Newf("invalid value %q", x)
+	}
+	return nil
+}
+
+var (
```
rogpeppe: I'm not sure I'd bother with these tbh.
Just mongodoc.IntBool(true) would seem OK. Most of the places
you've used mongodoc.True and mongodoc.False can just
be "true" and "false" constants.

If you really think they do pull their weight then at least they should
be constants and not variables.

internal/mongodoc/doc.go
```diff
@@ -230,3 +242,35 @@ type Migration struct {
...
+	if err := raw.Unmarshal(&x); err != nil {
+		return errgo.Notef(err, "cannot unmarshal value")
+	}
+	switch x {
+	case 1:
+		*b = True
+	case -1:
+		*b = False
+	default:
+		return errgo.Newf("invalid value %q", x)
```
rogpeppe: s/%q/%d/  (x isn't a string)

internal/charmstore/migrations_test.go
mhilton: Yes I see, done.

internal/charmstore/migrations_test.go
mhilton: sounds like a good idea, done

internal/mongodoc/doc_test.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (s *DocSuite) TestIntBoolSetBSON(c *gc.C) {
+	test := bson.D{{"true", 1}, {"false", -1}}
+	b, err := bson.Marshal(test)
+	c.Assert(err, gc.IsNil)
+	result := make(map[string]mongodoc.IntBool, 2)
+	err = bson.Unmarshal(b, &result)
+	c.Assert(err, gc.IsNil)
+	c.Assert(result["true"], gc.Equals, mongodoc.True)
```
rogpeppe: c.Assert(result, jc.DeepEquals, map[string]mongodoc.IntBool{"true": true, "false": false})
?

internal/mongodoc/doc_test.go
```diff
@@ -0,0 +1,77 @@
...
+	c.Assert(err, gc.IsNil)
+	c.Assert(result["true"], gc.Equals, 1)
+	c.Assert(result["false"], gc.Equals, -1)
+}
+
+func (s *DocSuite) TestIntBoolSetBSON(c *gc.C) {
+	test := bson.D{{"true", 1}, {"false", -1}}
+	b, err := bson.Marshal(test)
+	c.Assert(err, gc.IsNil)
+	result := make(map[string]mongodoc.IntBool, 2)
```
rogpeppe: var result map[string]mongodoc.IntBool

no need to make it.

internal/v4/api.go
```diff
@@ -131,14 +131,15 @@ func ResolveURL(store *charmstore.Store, url *charm.Reference) error {
 	if url.Series != "" && url.Revision != -1 {
 		return nil
 	}
-	urls, err := store.ExpandURL(url)
-	if err != nil {
-		return errgo.Notef(err, "cannot expand URL")
+	entity, err := store.FindBestEntity(url, "_id", "promulgated-url")
+	if err != nil && errgo.Cause(err) != params.ErrNotFound {
```
rogpeppe, line 135: I think the NotFound error returned by FindBestEntity is probably fine for use here,
so we could just do:

```
 if err != nil { 
       return errgo.Mask(err, params.Is(params.ErrNotFound))
 }
```

internal/charmstore/store.go
mhilton, line 413: done.

internal/charmstore/store.go
mhilton, line 417: done.

internal/v4/api.go
```diff
@@ -266,36 +263,39 @@ func (h *Handler) entityExists(id *charm.Reference, req *http.Request) (bool, er
...
-	var val mongodoc.BaseEntity
-	id1 := *id
-	id1.Series = ""
-	id1.Revision = -1
-	err := h.store.DB.BaseEntities().
-		Find(bson.D{{"_id", &id1}}).
-		Select(selector).
-		One(&val)
-	if err == mgo.ErrNotFound {
+	fields := make([]string, 0, len(selector))
```
rogpeppe, line 266: It's kind of ironic that we've got a map of the right form here, which we're converting to a slice only for FindBaseEntity to convert it back into something suitable for use with mgo again.

internal/charmstore/store.go
mhilton, line 422: done.

internal/charmstore/store.go
mhilton: done.

internal/charmstore/store.go
mhilton: done.

internal/charmstore/store.go
mhilton: done.

internal/v4/api.go
rogpeppe: or i guess we could make FindBestEntity take a field selection interface{}, losing the type safety, but meaning we could just pass the selector directly.

internal/v4/api_test.go
```diff
@@ -1074,7 +1138,7 @@ var resolveURLTests = []struct {
 	url:      "foo",
 	notFound: true,
 }, {
-	url:      "trusty/bigdata",
+	url:      "rusty/bigdata",
```
rogpeppe: rusty? really?

internal/v4/archive.go
rogpeppe: Perhaps:
// Find the promulgated URL for the entity. Note: if the entity is not promulgated,
// getPromulgatedURL returns nil.

?

internal/v4/archive.go
```diff
@@ -162,7 +176,18 @@ func (h *Handler) servePutArchive(id *charm.Reference, w http.ResponseWriter, re
...
+	var pid *charm.Reference
+	if promulgatedURL != "" {
+		pid, err = charm.ParseReference(promulgatedURL)
+		if err != nil {
+			return badRequestf(err, "cannot parse promulgated url")
+		}
+		if pid.User != "" {
+			return badRequestf(nil, "promulgated URL cannot have a user")
+		}
```
rogpeppe, line 192: Check that the promulgated URL name is the same as id.Name?

internal/v4/archive.go
```diff
@@ -456,27 +488,21 @@ func setArchiveCacheControl(h http.Header, idFullySpecified bool) {
...
-	if !baseEntity.Promulgated {
+	if baseEntity == nil || !baseEntity.Promulgated {
 		return nil, nil
 	}
+	query := h.store.EntitiesQuery(&charm.Reference{
+		Series:   id.Series,
+		Name:     id.Name,
+		Revision: -1,
+	})
+	logger.Debugf("%v", query)
```
rogpeppe: Either delete this or make it a more informative log message (I think I'm more in favour of the former FWIW)

rogpeppe:
LGTM modulo some comments and suggestions above, and assuming QA goes OK. Thanks, really good stuff.

internal/charmstore/store_test.go
mhilton, line 58: Yes I have added explicit tests.

internal/charmstore/migrations.go
mhilton: If that's the case we should fix that first, lets make the data good rather than the code complicated.

internal/charmstore/search.go
mhilton, line 600: I'm not so sure it's much of an issue, but it's a different PR in either case.

internal/charmstore/store.go
mhilton, line 420: If you specify all the fields you go through line 420, or 423. you don't get this far.

internal/charmstore/store_test.go
mhilton: done.

internal/charmstore/store_test.go
mhilton: done.

internal/charmstore/store_test.go
mhilton, line 1614: done.

internal/charmstore/store_test.go
mhilton: done

internal/charmstore/store_test.go
mhilton: done.

internal/v4/api.go
mhilton: I think if we did that we'd have to change all of the store methods to be the same, for consistency. I think I prefer a function, although it would only be used here I think.

mhilton: done.

internal/v4/archive.go
mhilton: done.

internal/v4/archive.go
mhilton, line 183: done.

internal/charmstore/migrations_test.go
mhilton: There are some examples of deliberately inconsistent entities.

internal/charmstore/migrations_test.go
mhilton: done

internal/charmstore/migrations_test.go
mhilton: done

internal/charmstore/migrations_test.go
mhilton: I'm not sure why @frankban might know but It checks that we don't update those fields if name is already set.

internal/charmstore/store.go
rogpeppe, line 420: Ah, good point, thanks.

internal/charmstore/migrations_test.go
rogpeppe: In which case I'd suggest we go with my mkEntity suggestion - then you can even insert entities with an empty baseURL if you want...

internal/charmstore/migrations_test.go
mhilton: That is a stonkingly good idea it turns out. done.

internal/charmstore/migrations_test.go
mhilton: I've left these to make it explicit that we expect these to be unusual.

internal/charmstore/store.go
mhilton, line 137: Well no, but this function if really only used by tests. If the entity has been PUT then we don't necessarily want the base entity to be set to Promulgated (it might not be the latest), if the entity is POSTed then it will only have a promulgated URL if the base entity is promulgated.

internal/charmstore/store.go
mhilton: done.

internal/charmstore/migrations_test.go
mhilton: done.

internal/charmstore/store_test.go
mhilton: done.

internal/mongodoc/doc.go
mhilton: removed.

internal/mongodoc/doc.go
mhilton: done.

internal/mongodoc/doc_test.go
mhilton: done.

internal/mongodoc/doc_test.go
mhilton: done.

internal/v4/api.go
mhilton, line 135: This leads to inconsistent error messages depending on how the ID was queried. I'm going to leave this.

internal/v4/api.go
mhilton, line 266: Yes, we could possibly clean that up, but it will require a change of interface somewhere. I'm going to leave this for another PR.

internal/v4/api_test.go
mhilton: oops, done.

internal/v4/archive.go
mhilton, line 192: Good idea, done.

internal/v4/archive.go
mhilton: That should never have been left there in the first place, done.

internal/charmstore/migrations.go
```diff
@@ -222,3 +227,93 @@ func populateWriteACL(db StoreDatabase) error {
...
+				{"blobhash", entity.BlobHash},
+				{"promulgated-url", bson.D{{"$exists", false}}},
+			},
+			bson.D{{"$set", bson.D{
+				{"promulgated-url", entity.URL},
+				{"promulgated-revision", entity.URL.Revision},
+			}}},
+		)
+		// Ignore not found errors as these will occur if the update is happening in parallel.
+		if err != nil && err != mgo.ErrNotFound {
```
rogpeppe: It seems to me that the "not found" case is quite important here, as this is where we can end up deleting entities if there happen to be any promulgated entities that don't have equivalent user charms. Perhaps if there's a not-found error, we could make sure that we can find the user charm, and abort if we can't, because in that case there's a real problem.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/768/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/769/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/770/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/771/
Test PASSed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/773/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/776/
Test PASSed.

mhilton:
:shipit:

jujugui:

PR #308, opened 2015-03-10
internal/v4: implement StatsEnabled

rogpeppe:
Implement StatsEnabled, use it where we already check for stats-enabled,
and also allow stats disabling in the legacy API.

### DIFF
--- a/internal/v4/stats.go
+++ b/internal/v4/stats.go
@@ -110,3 +110,12 @@ func (h *Handler) serveStatsCounter(_ http.Header, r *http.Request) (interface{}
 
 	return items, nil
 }
+
+// StatsEnabled reports whether statistics should be gathered for
+// the given HTTP request.
+func StatsEnabled(req *http.Request) bool {
+	// It's fine to parse the form more than once, and it avoids
+	// bugs from not parsing it.
+	req.ParseForm()
+	return req.Form.Get("stats") != "0"
+}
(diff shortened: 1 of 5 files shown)


internal/v4/stats.go
```diff
@@ -110,3 +110,12 @@ func (h *Handler) serveStatsCounter(_ http.Header, r *http.Request) (interface{}
 
 	return items, nil
 }
+
+// StatsEnabled reports whether statistics should be gathered for
+// the given HTTP request.
+func StatsEnabled(req *http.Request) bool {
+	// It's fine to parse the form more than once, and it avoids
```
mhilton, line 117: In our case won't this always have already been called by the router?

internal/v4/stats_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/juju/testing/httptesting"
 	gc "gopkg.in/check.v1"
 
+	"github.com/juju/charmstore/internal/v4"
```
mhilton, line 16: These should be in alphabetical order

mhilton:

frankban:
:+1: thank you!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/778/
Test PASSed.

internal/v4/stats_test.go
rogpeppe, line 16: done

internal/v4/stats.go
rogpeppe, line 117: Not necessarily in the legacy case. I don't mind leaving it there - it's only a few ns.

rogpeppe:
:shipit:

jujugui:

PR #310, opened 2015-03-10
Update config values for idm defaults

bac:
In the default config file, update the port and public key for the identity manager.


bac:
This is a very trivial change, part of an effort to make all of our services have reasonable, non-conflicting defaults.  The goal is to be able to run the equivalent of 'make server' in each to get cooperating services up and running with a minimal of editing of config files.

Most of the changes will be made to blues browser.

The default ports are
8080 - charmstore
8081 - idm

Omnibus has agreed to move their ports to a range starting at 8580 so we won't bump into them.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/783/
Test PASSed.

mhilton:
:+1: LGTM I do wonder if this should be the details for api.jujucharms.com/identity though, that might make it easier to work on the charmstore without any other infrastructure.

bac:
@mhilton I was trying to unify all of the default configs for running services locally, which is how this one was originally set up. If you think it makes more sense to do api.jujucharms.com/identity then we can do that and keep the other setting in the file but commented out, for ease of use when needed.  Do you have the public key for production identity?

mhilton:
@bac hmHaPgCC1UfuhYHUSX5+aihSAZesqpVdjRv0mgfIwjo= just so you know you can get it from https://api.jujucharms.com/identity/v1/discharger/publickey

rogpeppe:
LGTM, thanks.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/797/
Test PASSed.

bac:
:shipit:

jujugui:

PR #317, opened 2015-03-12
API endpoint to set promulgated flag.

mhilton:
Added a new $id/meta/promulgated endpoint to query the promulgated status
of an entity. Also added a $id/restricted/promulgate endpoint that
privileged users can use to promulgate a base entity.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -505,10 +505,11 @@ func (s *Store) UpdateBaseEntity(url *charm.Reference, update interface{}) error
 	return nil
 }
 
-// Promulgate sets the base entity of url to be promulgated, and unsets
-// promulgated on any other base entity for charms with the same name. It
-// also calculates the next promulgated URL for the charms owned by the
-// new owner and sets those charms.
+// SetPromulgated sets whether the base entity of url is promulgated, If
+// promulgated is true it also unsets promulgated on any other base
+// entity for entities with the same name. It also calculates the next
+// promulgated URL for the entities owned by the new owner and sets those
+// entities appropriately.
 //
 // Note: This code is known to have some unfortunate (but not dangerous)
 // race conditions. It is possible that if one or more promulgations
@@ -522,12 +523,22 @@ func (s *Store) UpdateBaseEntity(url *charm.Reference, update interface{}) error
 // This will be remedied when a new charm is uploaded by the promulgated
 // user. As promulgation is a rare operation, it is considered that the
 // chances this will happen are slim.
-func (s *Store) Promulgate(url *charm.Reference) error {
-	entity, err := s.FindEntity(url)
+func (s *Store) SetPromulgated(url *charm.Reference, promulgate bool) error {
+	entity, err := s.FindBestEntity(url)
 	if err != nil {
 		return errgo.Mask(err, errgo.Any)
 	}
 	bURL := baseURL(entity.URL)
+	if !promulgate {
+		err := s.DB.BaseEntities().UpdateId(
+			bURL,
+			bson.D{{"$set", bson.D{{"promulgated", mongodoc.IntBool(false)}}}},
+		)
+		if err != nil {
+			return errgo.Mask(err)
+		}
+		return nil
+	}
 
 	// Clear the promulgated flag on any Base entities with the same name,
 	// but aren't the one that is being set.
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -144,9 +144,10 @@ type Router struct {
 // The Cause of the resolveURL error will be left unchanged,
 // as for the handlers.
 //
-// The authorize function will be called to authorize the request.
-// The Cause of the authorize error will be left unchanged,
-// as for the handlers.
+// The authorize function will be called to authorize the request
+// to any BulkIncludeHandlers. All other handlers are expected
+// to handle their own authorization. The Cause of the authorize
+// error will be left unchanged, as for the handlers.
 //
 // The exists function may be called to test whether an entity
 // exists when an API endpoint needs to know that
@@ -241,9 +242,6 @@ func (r *Router) serveIds(w http.ResponseWriter, req *http.Request) error {
 	}
 	if handler != nil {
 		req.URL.Path = path
-		if err := r.authorize(url, req); err != nil {
-			return errgo.Mask(err, errgo.Any)
-		}
 		err := handler(url, fullySpecified, w, req)
 		// Note: preserve error cause from handlers.
 		return errgo.Mask(err, errgo.Any)
--- a/internal/storetesting/entities.go
+++ b/internal/storetesting/entities.go
@@ -1,0 +1,131 @@
+// Copyright 2014 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package storetesting
+
+import (
+	jc "github.com/juju/testing/checkers"
+	gc "gopkg.in/check.v1"
+	"gopkg.in/juju/charm.v5-unstable"
+	"gopkg.in/mgo.v2"
+
+	"gopkg.in/juju/charmstore.v4/internal/mongodoc"
+)
+
+// EntityBuilder provides a convenient way to describe a mongodoc.Entity
+// for tests that is correctly formed and contains the desired
+// information.
+type EntityBuilder struct {
+	entity *mongodoc.Entity
+}
+
+// NewEntity creates a new EntityBuilder for the provided URL.
+func NewEntity(url string) EntityBuilder {
+	URL := charm.MustParseReference(url)
+	return EntityBuilder{
+		entity: &mongodoc.Entity{
+			URL:                 URL,
+			Name:                URL.Name,
+			Series:              URL.Series,
+			Revision:            URL.Revision,
+			User:                URL.User,
+			BaseURL:             baseURL(URL),
+			PromulgatedRevision: -1,
+		},
+	}
+}
+
+func copyURL(id *charm.Reference) *charm.Reference {
+	if id == nil {
+		return nil
+	}
+	id1 := *id
+	return &id1
+}
+
+func (b EntityBuilder) copy() EntityBuilder {
+	e := *b.entity
+	e.PromulgatedURL = copyURL(e.PromulgatedURL)
+	e.URL = copyURL(e.URL)
+	e.BaseURL = copyURL(e.BaseURL)
+	return EntityBuilder{&e}
+}
+
+// WithPromulgatedURL sets the PromulgatedURL and PromulgatedRevision of the
+// entity being built.
+func (b EntityBuilder) WithPromulgatedURL(url string) EntityBuilder {
+	b = b.copy()
+	if url == "" {
+		b.entity.PromulgatedURL = nil
+		b.entity.PromulgatedRevision = -1
+	} else {
+		b.entity.PromulgatedURL = charm.MustParseReference(url)
+		b.entity.PromulgatedRevision = b.entity.PromulgatedURL.Revision
+	}
+	return b
+}
+
+// Build creates a mongodoc.Entity from the EntityBuilder.
+func (b EntityBuilder) Build() *mongodoc.Entity {
+	return b.copy().entity
+}
+
+// AssertEntity checks that db contains an entity that matches expect.
+func AssertEntity(c *gc.C, db *mgo.Collection, expect *mongodoc.Entity) {
+	var entity mongodoc.Entity
+	err := db.FindId(expect.URL).One(&entity)
+	c.Assert(err, gc.IsNil)
+	c.Assert(&entity, jc.DeepEquals, expect)
+}
+
+// BaseEntityBuilder provides a convenient way to describe a
+// mongodoc.BaseEntity for tests that is correctly formed and contains the
+// desired information.
+type BaseEntityBuilder struct {
+	baseEntity *mongodoc.BaseEntity
+}
+
+// NewBaseEntity creates a new BaseEntityBuilder for the provided URL.
+func NewBaseEntity(url string) BaseEntityBuilder {
+	URL := charm.MustParseReference(url)
+	return BaseEntityBuilder{
+		baseEntity: &mongodoc.BaseEntity{
+			URL:  URL,
+			Name: URL.Name,
+			User: URL.User,
+		},
+	}
+}
+
+func (b BaseEntityBuilder) copy() BaseEntityBuilder {
+	e := *b.baseEntity
+	e.URL = copyURL(e.URL)
+	return BaseEntityBuilder{&e}
+}
+
+// WithPromulgated sets the promulgated flag on the BaseEntity.
+func (b BaseEntityBuilder) WithPromulgated(promulgated bool) BaseEntityBuilder {
+	b = b.copy()
+	b.baseEntity.Promulgated = mongodoc.IntBool(promulgated)
+	return b
+}
+
+// Build creates a mongodoc.BaseEntity from the BaseEntityBuilder.
+func (b BaseEntityBuilder) Build() *mongodoc.BaseEntity {
+	return b.copy().baseEntity
+}
+
+// AssertBaseEntity checks that db contains a base entity that matches expect.
+func AssertBaseEntity(c *gc.C, db *mgo.Collection, expect *mongodoc.BaseEntity) {
+	var baseEntity mongodoc.BaseEntity
+	err := db.FindId(expect.URL).One(&baseEntity)
+	c.Assert(err, gc.IsNil)
+	c.Assert(&baseEntity, jc.DeepEquals, expect)
+}
+
+func baseURL(url *charm.Reference) *charm.Reference {
+	baseURL := *url
+	baseURL.Series = ""
+	baseURL.Revision = -1
+	return &baseURL
+}
--- a/internal/v4/api.go
+++ b/internal/v4/api.go
@@ -6,6 +6,7 @@ package v4
 import (
 	"archive/zip"
 	"encoding/json"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -58,13 +59,14 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
 			"macaroon":           router.HandleJSON(h.serveMacaroon),
 		},
 		Id: map[string]router.IdHandler{
-			"archive":     h.serveArchive,
-			"archive/":    h.serveArchiveFile,
-			"diagram.svg": h.serveDiagram,
-			"expand-id":   h.serveExpandId,
-			"icon.svg":    h.serveIcon,
-			"readme":      h.serveReadMe,
-			"resources":   h.serveResources,
+			"archive":     h.authIdHandler(h.serveArchive),
+			"archive/":    h.authIdHandler(h.serveArchiveFile),
+			"diagram.svg": h.authIdHandler(h.serveDiagram),
+			"expand-id":   h.authIdHandler(h.serveExpandId),
+			"icon.svg":    h.authIdHandler(h.serveIcon),
+			"readme":      h.authIdHandler(h.serveReadMe),
+			"resources":   h.authIdHandler(h.serveResources),
+			"promulgate":  h.serveAdminPromulgate,
 		},
 		Meta: map[string]router.BulkIncludeHandler{
 			"archive-size":         h.entityHandler(h.metaArchiveSize, "size"),
@@ -97,6 +99,7 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
 			"manifest":      h.entityHandler(h.metaManifest, "blobname"),
 			"perm":          h.puttableBaseEntityHandler(h.metaPerm, h.putMetaPerm, "acls"),
 			"perm/":         h.puttableBaseEntityHandler(h.metaPermWithKey, h.putMetaPermWithKey, "acls"),
+			"promulgated":   h.baseEntityHandler(h.metaPromulgated, "promulgated"),
 			"revision-info": router.SingleIncludeHandler(h.metaRevisionInfo),
 			"stats":         h.entityHandler(h.metaStats),
 			"tags":          h.entityHandler(h.metaTags, "charmmeta", "bundledata"),
@@ -707,6 +710,14 @@ func (h *Handler) putMetaPerm(id *charm.Reference, path string, val *json.RawMes
 	return nil
 }
 
+// GET id/meta/promulgated
+// See https://github.com/juju/charmstore/blob/v4/docs/API.md#get-idmetapromulgated
+func (h *Handler) metaPromulgated(entity *mongodoc.BaseEntity, id *charm.Reference, path string, flags url.Values, req *http.Request) (interface{}, error) {
+	return params.PromulgatedResponse{
+		Promulgated: bool(entity.Promulgated),
+	}, nil
+}
+
 // GET id/meta/perm/key
 // https://github.com/juju/charmstore/blob/v4/docs/API.md#get-idmetapermkey
 func (h *Handler) metaPermWithKey(entity *mongodoc.BaseEntity, id *charm.Reference, path string, flags url.Values, req *http.Request) (interface{}, error) {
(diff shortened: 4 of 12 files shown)


mhilton:
Note: I'm not convinced by the naming of the /restricted subtree, I welcome suggestions.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/805/
Test PASSed.

mhilton:
## QA Instructions 

```
http :8080/v4/cinder-ceph/meta/any?include=promulgated
{
    Id: "cs:trusty/cinder-ceph-4"
    Meta: {
        promulgated: {
            Promulgated: true
        }
    }
}

http :8080/v4/cinder-ceph/meta/promulgated
{
    Promulgated: true
}

http PUT http://admin:password@:8080/v4/~openstack-charmers/cinder-ceph/promulgate --json --stdin <<EOF
{"Promulgate": false}
EOF

http :8080/v4/~openstack-charmers/cinder-ceph/meta/promulgated
{
    Promulgated: false
}

http PUT http://admin:password@:8080/v4/~openstack-charmers/cinder-ceph/promulgate --json --stdin <<EOF
{"Promulgate": true}
EOF

http :8080/v4/~openstack-charmers/cinder-ceph/meta/promulgated
{
    Promulgated: true
}
```

internal/router/router.go
```diff
@@ -151,17 +167,23 @@ type Router struct {
...
+//
+// The authorizeAdmin function will be called to authorize admin
+// requests. The Cause of the authorizeAdmin error will be left
+// unchanged, as for the handlers.
 func New(
 	handlers *Handlers,
 	resolveURL func(id *charm.Reference) error,
 	authorize func(id *charm.Reference, req *http.Request) error,
 	exists func(id *charm.Reference, req *http.Request) (bool, error),
+	authorizeWithPerms func(req *http.Request, read, write []string) error,
```
rogpeppe: I wonder if rather than having two functions with similar but overlapping behaviour, maybe we could have two functions that don't overlap so much, say:

```
permsForEntity func(id *charm.Reference) ([]string, error)
authorize func(req *http.Request, read, write []string) error
```

The first one would be called any time we'd previously call authorize.
The second one would be called for any authorization request.

This opens the door for permissions on other kinds of entity with a
reasonably uniform mechanism.

internal/router/router.go
```diff
@@ -230,7 +252,7 @@ func (r *Router) serveIds(w http.ResponseWriter, req *http.Request) error {
 	fullySpecified := url.Series != "" && url.Revision != -1
 	handler := r.handlers.Id[key]
 	if handler == nil || idHandlerNeedsResolveURL(req) {
-		// If it's not an id handler, it's a meta endpoint, so
+		// If it's not an id handler, it's a meta or admin endpoint, so
```
rogpeppe: s/admin/restricted/

internal/router/router.go
```diff
@@ -694,6 +720,24 @@ func (r *Router) PutMetadata(id *charm.Reference, data map[string]*json.RawMessa
 	return nil
 }
 
+func (r *Router) serveRestricted(id *charm.Reference, fullySpecified bool, w http.ResponseWriter, req *http.Request) error {
+	key, path := handlerKey(req.URL.Path)
+	if key == "" {
+		return errgo.WithCausef(nil, params.ErrNotFound, "")
+	}
+	handler := r.handlers.Restricted[key]
+	if handler.Handler != nil {
```
rogpeppe:
```
 if handler.Handler == nil {
      return params.ErrNotFound
 }
```

?

internal/router/router.go
```diff
@@ -694,6 +720,24 @@ func (r *Router) PutMetadata(id *charm.Reference, data map[string]*json.RawMessa
 	return nil
 }
 
+func (r *Router) serveRestricted(id *charm.Reference, fullySpecified bool, w http.ResponseWriter, req *http.Request) error {
+	key, path := handlerKey(req.URL.Path)
+	if key == "" {
+		return errgo.WithCausef(nil, params.ErrNotFound, "")
```
rogpeppe: or just
     return params.ErrNotFound
?

internal/router/router.go
urosj: permsForEntity should probably return a map, no? But I agree that what you suggest leads to a more flexible approach.

internal/router/router.go
```diff
@@ -694,6 +720,24 @@ func (r *Router) PutMetadata(id *charm.Reference, data map[string]*json.RawMessa
 	return nil
 }
 
+func (r *Router) serveRestricted(id *charm.Reference, fullySpecified bool, w http.ResponseWriter, req *http.Request) error {
```
urosj: Could you document this function, please, how and why it does that. Thank you.

internal/v4/api.go
```diff
@@ -104,7 +106,14 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
 			// endpoints not yet implemented:
 			// "color": router.SingleIncludeHandler(h.metaColor),
 		},
-	}, h.resolveURL, h.authorizeEntity, h.entityExists)
+		Restricted: map[string]router.RestrictedIdHandler{
```
urosj: I see what you mean, but kind of don't like the "/restricted" endpoint, as it is no more special as any other restricted endpoint, imho. You just need to be a member of a certain group to have access to it.

urosj:
Looked - need clarification why we need to have special /restricted path, as I don't see this "promulgate" endpoint no more special than any other restricted endpoint.

internal/storetesting/entities.go
```diff
@@ -0,0 +1,102 @@
...
+	jc "github.com/juju/testing/checkers"
+	gc "gopkg.in/check.v1"
+	"gopkg.in/juju/charm.v5-unstable"
+	"gopkg.in/mgo.v2"
+
+	"gopkg.in/juju/charmstore.v4/internal/mongodoc"
+)
+
+// EntityBuilder provides a convenient way to describe a mongodoc.Entity
+// for tests that is correctly formed and contain the desired
```
rogpeppe: s/contain/contains/

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/814/
Test PASSed.

internal/storetesting/entities.go
```diff
@@ -0,0 +1,102 @@
...
+			BaseURL:             baseURL(URL),
+			PromulgatedRevision: -1,
+		},
+	}
+}
+
+// PromulgatedURL sets the PromulgatedURL and PromulgatedRevision of the
+// entity being built.
+func (b EntityBuilder) PromulgatedURL(url string) EntityBuilder {
+	b.entity.PromulgatedURL = charm.MustParseReference(url)
```
rogpeppe: I wonder if this should always make a copy, just to avoid awkward aliasing mistakes
in test code. Also, "PromulgatedURL" doesn't sound like it changes anything
(I was a bit confused by that when I saw some code using this earlier)

Something like this maybe?

```
func (b EntityBuilder) WithPromulgatedURL(url string) EntityBuilder {
   b = b.copy()
   b.entity.PromulgatedURL = charm.MustParseReference(url)
   b.entity.PromulgatedRevision = b.entity.PromulgatedURL.Revision
   return b
}

func (b EntityBuilder) copy(EntityBuilder) EntityBuilder {
    e := *b.entity
    e.PromulgatedURL = copyURL(e.PromulgatedURL)
    e.URL = copyURL(e.URL)
    e.BaseURL = copyURL(e.BaseURL)
    return EntityBuilder{&e}
}

func (b EntityBuilder) Build() *Entity {
     return b.copy().entity
}
```

?

internal/storetesting/entities.go
```diff
@@ -0,0 +1,102 @@
...
+		baseEntity: &mongodoc.BaseEntity{
+			URL:  URL,
+			Name: URL.Name,
+			User: URL.User,
+		},
+	}
+}
+
+// Promulgated sets the promulgated flag on the BaseEntity.
+func (b BaseEntityBuilder) Promulgate() BaseEntityBuilder {
```
rogpeppe: s/Promulgate/Promulgated/

?

internal/storetesting/entities.go
```diff
@@ -0,0 +1,102 @@
...
+	var entity mongodoc.Entity
+	err := db.FindId(expect.URL).One(&entity)
+	c.Assert(err, gc.IsNil)
+	c.Assert(&entity, jc.DeepEquals, expect)
+}
+
+// BaseEntityBuilder provides a convenient way to describe a
+// mongodoc.BaseEntity for tests that is correctly formed and contain the
+// desired information.
+type BaseEntityBuilder struct {
```
rogpeppe, line 84: same thoughts apply here

internal/v4/api.go
```diff
@@ -58,13 +59,14 @@ func New(store *charmstore.Store, config charmstore.ServerParams) *Handler {
...
 		},
 		Id: map[string]router.IdHandler{
-			"archive":     h.serveArchive,
-			"archive/":    h.serveArchiveFile,
-			"diagram.svg": h.serveDiagram,
-			"expand-id":   h.serveExpandId,
-			"icon.svg":    h.serveIcon,
-			"readme":      h.serveReadMe,
-			"resources":   h.serveResources,
+			"archive":     h.idHandler(h.serveArchive),
```
rogpeppe: Perhaps it might be better to name "idHandler" after the
reason we're using it, so:

s/idHandler/authIdHandler/

?

internal/v4/api.go
```diff
@@ -815,3 +826,44 @@ func (h *Handler) serveChangesPublished(_ http.Header, r *http.Request) (interfa
...
+	var promulgate params.PromulgateRequest
+	if err := json.Unmarshal(data, &promulgate); err != nil {
+		return errgo.WithCausef(err, params.ErrBadRequest, "")
+	}
+	if promulgate.Promulgate {
+		if err := h.store.Promulgate(id); err != nil {
+			return errgo.Mask(err, errgo.Any)
+		}
+	} else {
+		if err := h.store.UpdateBaseEntity(id, bson.D{{"$set", bson.D{{"promulgated", mongodoc.IntBool(false)}}}}); err != nil {
```
rogpeppe: It does seem slightly odd that this isn't in charmstore alongside
Promulgate.
I wonder if we should have Store.SetPromulgated(id, isPromulgated).
The isPromulgated=true case is trivial, admittedly, but at least
that keeps the base entity promulgation code all in one place.

internal/v4/api_test.go
```diff
@@ -40,6 +41,15 @@ import (
 	"gopkg.in/juju/charmstore.v4/params"
 )
 
+var publicKey = bakery.PublicKey{
```
rogpeppe: s/publicKey/testPublicKey/
?

params/params.go
```diff
@@ -190,6 +190,18 @@ type PermRequest struct {
 	Write []string
 }
 
+// PromulgatedResponse holds the result of an id/meta/promulgated GET request.
+// See TODO(mhilton)
```
rogpeppe: Where's that?

params/params.go
```diff
@@ -190,6 +190,18 @@ type PermRequest struct {
...
+type PromulgatedResponse struct {
+	Promulgated bool
+}
+
+// PromulgateRequest holds the request of an id/restricted/promulgate PUT request.
+// See TODO(mhilton)
```
rogpeppe: ditto

internal/storetesting/entities.go
rogpeppe: Scratch that - Promulgate is probably better here

internal/charmstore/store.go
```diff
@@ -523,7 +523,7 @@ func (s *Store) UpdateBaseEntity(url *charm.Reference, update interface{}) error
 // user. As promulgation is a rare operation, it is considered that the
 // chances this will happen are slim.
 func (s *Store) Promulgate(url *charm.Reference) error {
-	entity, err := s.FindEntity(url)
+	entity, err := s.FindBestEntity(url)
 	if err != nil {
 		return errgo.Mask(err, errgo.Any)
 	}
```
frankban, line 530: Since we now grab a base entity, do we still need to call baseUrl here below?

internal/storetesting/entities.go
```diff
@@ -0,0 +1,102 @@
...
+			User:                URL.User,
+			BaseURL:             baseURL(URL),
+			PromulgatedRevision: -1,
+		},
+	}
+}
+
+// PromulgatedURL sets the PromulgatedURL and PromulgatedRevision of the
+// entity being built.
+func (b EntityBuilder) PromulgatedURL(url string) EntityBuilder {
```
frankban: This seems more a getter than a setter. Maybe just Promulgate(url string)? [shrug]

frankban: Oh, it returns an entity and does not modify it in place.
So maybe WithPromulgatedURL and fix the docstring?

internal/storetesting/entities.go
```diff
@@ -0,0 +1,102 @@
...
+
+// EntityBuilder provides a convenient way to describe a mongodoc.Entity
+// for tests that is correctly formed and contain the desired
+// information.
+type EntityBuilder struct {
+	entity *mongodoc.Entity
+}
+
+// NewEntity creates a new EntityBuilder for the provided URL.
+func NewEntity(url string) EntityBuilder {
```
frankban, line 23: NewEntityBuilder?

internal/storetesting/entities.go
```diff
@@ -0,0 +1,102 @@
...
+
+// BaseEntityBuilder provides a convenient way to describe a
+// mongodoc.BaseEntity for tests that is correctly formed and contain the
+// desired information.
+type BaseEntityBuilder struct {
+	baseEntity *mongodoc.BaseEntity
+}
+
+// NewBaseEntity creates a new BaseEntityBuilder for the provided URL.
+func NewBaseEntity(url string) BaseEntityBuilder {
```
frankban, line 89: NewBaseEntityBuilder?

internal/v4/api.go
```diff
@@ -815,3 +826,44 @@ func (h *Handler) serveChangesPublished(_ http.Header, r *http.Request) (interfa
 func (h *Handler) serveMacaroon(_ http.Header, _ *http.Request) (interface{}, error) {
 	return h.newMacaroon()
 }
+
+// GET id/promulgate
+// TODO (mhilton): document admin/promulgate endpoint
```
frankban: admin?

internal/v4/api.go
```diff
@@ -815,3 +826,44 @@ func (h *Handler) serveChangesPublished(_ http.Header, r *http.Request) (interfa
...
+func (h *Handler) serveAdminPromulgate(id *charm.Reference, _ bool, w http.ResponseWriter, req *http.Request) error {
+	if err := h.authorize(req, []string{"promulgators"}); err != nil {
```
frankban: I'd put "promulgators" as a constant somewhere.

internal/v4/api.go
frankban: Should we have a store.Unpromulgate helper?

params/params.go
```diff
@@ -190,6 +190,18 @@ type PermRequest struct {
 	Write []string
 }
 
+// PromulgatedResponse holds the result of an id/meta/promulgated GET request.
+// See TODO(mhilton)
+type PromulgatedResponse struct {
+	Promulgated bool
+}
+
+// PromulgateRequest holds the request of an id/restricted/promulgate PUT request.
```
frankban: This is no longer restricted, right?

internal/v4/api_test.go
```diff
@@ -2048,3 +2073,339 @@ func (s *APISuite) TestMacaroon(c *gc.C) {
...
+		for _, e := range test.expectBaseEntities {
+			storetesting.AssertBaseEntity(c, store.DB.BaseEntities(), e)
+		}
+	}
+}
+
+// dischargeRequiredBody returns a httptesting.BodyAsserter that checks
+// that the response body contains a discharge required error holding a macaroon
+// with a third-party caveat addressed to expectedEntityLocation.
+func dischargeRequiredBody(expectIdentityLocation string) httptesting.BodyAsserter {
```
frankban, line 2396: Very nice!

internal/v4/api_test.go
```diff
@@ -2048,3 +2073,339 @@ func (s *APISuite) TestMacaroon(c *gc.C) {
...
+		if test.method == "" {
+			test.method = "PUT"
+		}
+		header := http.Header{"Content-Type": {"application/json"}}
+		if len(test.caveats) > 0 {
+			m, err := store.Bakery.NewMacaroon("", nil, test.caveats)
+			c.Assert(err, gc.IsNil)
+			cookie, err := httpbakery.NewCookie(macaroon.Slice{m})
+			c.Assert(err, gc.IsNil)
+			header.Set("Cookie", (&http.Cookie{
```
frankban, line 2362: Why not just header.Add("Set-Cookie", cookie.String())?

internal/v4/api_test.go
```diff
@@ -2048,3 +2073,339 @@ func (s *APISuite) TestMacaroon(c *gc.C) {
 		},
 	})
 }
+
+var promulgateTests = []struct {
```
frankban, line 2077: Very nice tests.

frankban:
:+1: with minor changes.
Thank you!

internal/storetesting/entities.go
mhilton: Done.

internal/storetesting/entities.go
mhilton: I changed this to WithPromulgated in the end.

params/params.go
mhilton: done.

params/params.go
mhilton: done.

internal/storetesting/entities.go
mhilton, line 84: Done.

internal/v4/api.go
mhilton: Done.

internal/v4/api_test.go
rogpeppe, line 2362: Because "Set-Cookie" is a header returned by a server to a client, not the other way around.
The reason for making a new cookie is that httpbakery.NewCookie returns a cookie that's
appropriate for server use - when it is converted to a string, it may include information
that's not appropriate for a client call. See the docs for http.Cookie.String.

internal/v4/api.go
mhilton: Done, based on @rogpeppe's suggestion.

internal/v4/api_test.go
mhilton: Done.

internal/charmstore/store.go
mhilton, line 530: We don't we find the best match to the requested entity.

internal/storetesting/entities.go
mhilton: @rogpeppe had the same idea, done.

internal/storetesting/entities.go
mhilton, line 23: I think I'm, going to leave these as they are I quite like that you can read it as NewEntity(...).WithPromulgatedURL(...)

internal/storetesting/entities.go
mhilton, line 89: See above.

internal/v4/api.go
mhilton: That was an old location for it, more importantly it needs documenting, which is now done.

jrwren:
I did an ingestion before QA to be sure I had reasonable test data.

```
$ http :8080/v4/cinder-ceph/meta/any?include=promulgated
{
    Message: "no matching charm or bundle for cs:trusty/cinder-ceph-5"
    Code: "not found"
}
```

This suggests to me that something is wrong with how I did ingestion.

So I move to the set promulgation tests and I get unauthorized, even though i'm using http auth.

```
$ http PUT http://admin:password@:8080/v4/~openstack-charmers/cinder-ceph/promulgate --json --stdin <<EOF
> {"Promulgate": false}
> {
    Message: "invalid user name or password"
    Code: "unauthorized"
}
```

The get works:

```
$ http :8080/v4/~openstack-charmers/cinder-ceph/meta/promulgated
{
    Promulgated: false
}
```

Maybe I am missing some QA instructions?

params/params.go
mhilton: correct, updated.

internal/v4/api_test.go
mhilton, line 2362: It's Cookie rather than Set-Cookie because it's in the request. http.Cookie.String() will include fields we don't want in the request if they're specified. This provides some future proofing.

mhilton:
@jrwren are you sure admin:password is correct for your system?

jrwren:
@mhilton 

I used admin:example-password, which is correct for my system.

```
$ http PUT http://admin:example-password@:8080/v4/~openstack-charmers/cinder-ceph/promulgate --json --stdin <<EOF
> {"Promulgate": false}
> {
    Message: "invalid user name or password"
    Code: "unauthorized"
}
```

jrwren:
:+1: 

```
$ http :8080/v4/cinder-ceph/meta/any?include=promulgated
{
    Id: "cs:trusty/cinder-ceph-5"
    Meta: {
        promulgated: {
            Promulgated: true
        }
    }
}
$ http :8080/v4/cinder-ceph/meta/promulgated
{
    Promulgated: true
}
$ http PUT http://admin:example-passwd@:8080/v4/~openstack-charmers/cinder-ceph/promulgate --json --stdin <<EOF
> {"Promulgate": false}
> $ http :8080/v4/~openstack-charmers/cinder-ceph/meta/promulgated
{
    Promulgated: false
}
$ http PUT http://admin:example-passwd@:8080/v4/~openstack-charmers/cinder-ceph/promulgate --json --stdin <<EOF
> {"Promulgate": true}
> $ http :8080/v4/~openstack-charmers/cinder-ceph/meta/promulgated
{
    Promulgated: true
}
```

internal/v4/api.go
mhilton: done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/815/
Test PASSed.

internal/v4/api.go
```diff
@@ -26,6 +27,8 @@ import (
 	"gopkg.in/juju/charmstore.v4/params"
 )
 
+const promulgatorsGroup = "promulgators"
```
rogpeppe: I think I'd expect to find this in auth.go

internal/v4/api_test.go
```diff
@@ -2048,3 +2073,339 @@ func (s *APISuite) TestMacaroon(c *gc.C) {
...
+		httptesting.AssertJSONCall(c, httptesting.JSONCallParams{
+			Handler:      srv,
+			URL:          storeURL(test.id + "/promulgate"),
+			Method:       test.method,
+			Body:         test.body,
+			Header:       header,
+			Username:     test.username,
+			Password:     test.password,
+			ExpectStatus: test.expectStatus,
+			ExpectBody:   test.expectBody,
```
rogpeppe, line 2376: This uses enough fields from it that it seems like it might be worth just using JSONCallParams in the test. Something to consider for a future PR.

params/params.go
```diff
@@ -190,6 +190,18 @@ type PermRequest struct {
...
+// PromulgatedResponse holds the result of an id/meta/promulgated GET request.
+// See https://github.com/juju/charmstore/blob/v4/docs/API.md#get-idmetapromulgated
+type PromulgatedResponse struct {
+	Promulgated bool
+}
+
+// PromulgateRequest holds the request of an id/promulgate PUT request.
+// See https://github.com/juju/charmstore/blob/v4/docs/API.md#put-idpromulgate
+type PromulgateRequest struct {
+	Promulgate bool
```
rogpeppe: I think this should probably be Promulgated, as that's what we're putting.

rogpeppe:
LGTM modulo a couple of minor suggestions.

internal/v4/api.go
mhilton: done

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/817/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/818/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/819/
Test PASSed.

mhilton:
:shipit:

jujugui:

PR #333, opened 2015-03-31
Make sure only the latest promulgated version is listed in search result...

mhilton:
...s.

### DIFF
--- a/internal/charmstore/search.go
+++ b/internal/charmstore/search.go
@@ -78,12 +78,6 @@ func (s *Store) UpdateSearch(r *router.ResolvedURL) error {
 	if deprecatedSeries[r.URL.Series] {
 		return nil
 	}
-	// TODO This is not necessarily good enough, as we
-	// can get more than one entity in the search results
-	// even when there is only one promulgated entity.
-	// This can happen when a promulgated entity changes owners.
-	// Perhaps key the search entities by preferred URL, not
-	// canonical URL.
 
 	var query *mgo.Query
 	query = s.DB.Entities().Find(bson.D{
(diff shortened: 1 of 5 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/848/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/849/
Test PASSed.

internal/charmstore/search.go
```diff
@@ -98,11 +92,59 @@ func (s *Store) UpdateSearch(r *router.ResolvedURL) error {
...
 	}
-	doc, err := s.searchDocFromEntity(&entity, baseEntity)
+	iter := s.DB.Entities().Pipe([]bson.D{
+		{{"$match", bson.D{{"baseurl", baseURL}}}},
+		{{"$sort", bson.D{{"revision", 1}}}},
+		{{"$group", bson.D{
+			{"_id", "$series"},
+			{"entity", bson.D{{"$last", "$$CURRENT"}}},
+		}}},
+	}).Iter()
```
urosj, line 132: Should this be part of /charmstore/store.go as in FindLatestEntitesForBaseEntity? It would be nice if we have all direct access to DB in one place instead of all over the service. But I'm aware that we don't follow this rule already.

internal/charmstore/search.go
```diff
@@ -98,11 +92,59 @@ func (s *Store) UpdateSearch(r *router.ResolvedURL) error {
...
+	// database round trips. Unfortunately the version of mongoDB we support
+	// (2.4) would require every field to be enumerated in this query, which
+	// would make it too fragile.
+	iter := s.DB.Entities().Pipe([]bson.D{
+		{{"$match", bson.D{{"baseurl", baseURL}}}},
+		{{"$sort", bson.D{{"revision", 1}}}},
+		{{"$group", bson.D{
+			{"_id", "$series"},
+			{"url", bson.D{{"$last", "$_id"}}},
+		}}},
```
urosj, line 131: Should this be part of /charmstore/store.go as in FindLatestEntitesForBaseEntity? It would be nice if we have all direct access to DB in one place instead of all over the service. But I'm aware that we don't follow this rule already.

urosj:
:+1: with suggestion. Haven't done QA yet.

urosj:
QA ok.
- changed promulgation to another user, do search, ES returned proper results
- returned promulgation to the original charm, do search, ES returned proper results
- the new promulgation ID was taken but the "latest" was actually "last-1", which is ok

internal/charmstore/search.go
```diff
@@ -98,11 +92,59 @@ func (s *Store) UpdateSearch(r *router.ResolvedURL) error {
...
+		return errgo.Notef(err, "cannot get %s", entity.BaseURL)
 	}
-	doc, err := s.searchDocFromEntity(&entity, baseEntity)
+	if err := s.updateSearchEntity(&entity, baseEntity); err != nil {
+		return errgo.Notef(err, "cannot update search record for %q", entity.URL)
+	}
+	return nil
+}
+
+// UpdateSearchBaseURL updates the search record for all entities with the specified base URL.
```
rogpeppe: // It must be called whenever the entry for the given URL in
// the BaseEntitites collection has changed.
?

internal/charmstore/search.go
```diff
@@ -98,11 +92,59 @@ func (s *Store) UpdateSearch(r *router.ResolvedURL) error {
...
+		return errgo.New("base url cannot contain series")
+	}
+	if baseURL.Revision != -1 {
+		return errgo.New("base url cannot contain revision")
+	}
+	// Note: It it possible to return the complete entity here and save some
+	// database round trips. Unfortunately the version of mongoDB we support
+	// (2.4) would require every field to be enumerated in this query, which
+	// would make it too fragile.
+	iter := s.DB.Entities().Pipe([]bson.D{
```
rogpeppe, line 125: // For all entities that have the given base URL, find
// the ones with the latest revision.

?

internal/charmstore/search.go
```diff
@@ -98,11 +92,59 @@ func (s *Store) UpdateSearch(r *router.ResolvedURL) error {
...
+		{{"$match", bson.D{{"baseurl", baseURL}}}},
+		{{"$sort", bson.D{{"revision", 1}}}},
+		{{"$group", bson.D{
+			{"_id", "$series"},
+			{"url", bson.D{{"$last", "$_id"}}},
+		}}},
+	}).Iter()
```
rogpeppe, line 132: Do we want to do a Select on the iterator so that we're not retrieving
the entire document for each entity?

internal/charmstore/search.go
mhilton, line 131: I think this case is sufficiently specialized to not warrant it's own function, as it was originally intended to return the complete entity documents, I don't think that is the sort of thing we should return from a function call.

internal/charmstore/search.go
mhilton, line 132: As the entire document is the series and the URL I don't think it matters much. I'm not sure if we can filter out the "_id" without upsetting mongodb.

internal/charmstore/search.go
```diff
@@ -134,11 +176,16 @@ func (s *Store) UpdateSearchFields(r *router.ResolvedURL, fields map[string]inte
...
 
-// searchDocFromEntity performs the processing required to convert a mongodoc.Entity
-// to an esDoc for indexing.
+// searchDocFromEntity performs the processing required to convert a
+// mongodoc.Entity and the corresponding mongodoc.BaseEntity to an esDoc
+// for indexing.
 func (s *Store) searchDocFromEntity(e *mongodoc.Entity, be *mongodoc.BaseEntity) (*SearchDoc, error) {
 	doc := SearchDoc{Entity: e}
 	doc.ReadACLs = be.ACLs.Read
+	if !be.Promulgated {
```
rogpeppe, line 190: Perhaps a comment here, justifying the decision to exclude entities with promulgated URLs that don't have the promulgated bit set in their base entities?

internal/charmstore/store_test.go
```diff
@@ -2030,6 +2030,74 @@ func (s *StoreSuite) TestPromulgate(c *gc.C) {
 	}
 }
 
+func (s *StoreSuite) TestSetPromulgatedUpdateSearch(c *gc.C) {
```
rogpeppe, line 2033: Some commentary indicating what's being tested in each section of this function would be useful, I think.

rogpeppe:
LGTM modulo a few suggestions for additional comments above.

internal/charmstore/search.go
mhilton: Done

internal/charmstore/search.go
mhilton, line 125: Done, although I worded it as:
// From the entities with the specified base URL find the latest revision in
// each of the available series.

internal/charmstore/search.go
mhilton, line 190: Done.

internal/charmstore/store_test.go
mhilton, line 2033: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/856/
Test PASSed.

mhilton:
:shipit:

jujugui:

PR #347, opened 2015-04-08
Testing server improvements

frankban:


testing/server.go
```diff
@@ -4,6 +4,7 @@
 package testing
```
rogpeppe: I have to say I'm not so keen on yet another "testing" package - there are dozens
in juju-core and it can be confusing. I'd be inclined to name it "charmstoretesting"
because most importers will want to rename it to something like that anyway.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/883/
Test PASSed.

testing/server.go
```diff
@@ -28,12 +29,25 @@ import (
 // Callers are responsible of closing the server by calling Close().
 func OpenServer(c *gc.C, session *mgo.Session, params charmstore.ServerParams) *Server {
 	db := session.DB("charmstore-testing")
+	if params.AuthUsername == "" {
+		params.AuthUsername = "charmstore-testing-user"
```
rogpeppe: We should probably declare these as constants and document that
they're used like this.

urosj: Could you move this out of function as a constant, please?

testing/server.go
```diff
@@ -106,12 +121,7 @@ func (s *Server) UploadCharm(c *gc.C, ch charm.Charm, id *charm.Reference, promu
 	}
 
 	// Upload the charm.
```
rogpeppe, line 134: The docs for this function should probably be clearer about how the parameters are used.
For example, it should mention that the revision must be specified and whether the
promulgated flag causes the base entity to be promulgated. (sorry, I can't comment
on the function declaration itself)

testing/server.go
```diff
@@ -28,12 +29,25 @@ import (
...
 	handler, err := charmstore.NewServer(db, nil, "", params, charmstore.V4)
 	c.Assert(err, jc.ErrorIsNil)
+
+	srv := httptest.NewServer(handler)
 	return &Server{
-		srv:     httptest.NewServer(handler),
+		srv:     srv,
 		handler: handler,
 		params:  params,
+		client: csclient.New(csclient.Params{
```
rogpeppe: Doesn't this mean that the server keeps client cookies around for the lifetime of the server? I wonder if that's actually a good idea, as I can imagine that tests might
create the server as part of test suite setup and then you'd run the danger of test cross-contamination. I think I'd be inclined to keep server strictly separate from client (but provide an easy way of creating a client from the server).

testing/server.go
frankban: Done.

testing/server.go
frankban: Done.

testing/server.go
```diff
@@ -126,3 +136,22 @@ func (s *Server) UploadCharm(c *gc.C, ch charm.Charm, id *charm.Reference, promu
 	}
 	return result.Id
 }
+
+// ChangeReadPerm changes read permissions of the given entity.
+func (s *Server) ChangeReadPerm(c *gc.C, id *charm.Reference, perms ...string) {
```
rogpeppe: I'd prefer to add this method (or something like it) to the cs client itself - it's actually useful functionality, and not just for testing.

rogpeppe: Perhaps a PUT equivalent to the Get method would be a good starting point.

testing/server.go
frankban: Done.

testing/server.go
frankban, line 134: Done.

testing/server.go
frankban: Removed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/884/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/885/
Test PASSed.

charmstoretesting/server.go
```diff
@@ -24,12 +24,29 @@ import (
...
+	// AuthUsername is the default authentication user name used by the testing
+	// charm store server if none is provided.
+	AuthUsername = "charmstore-testing-user"
+
+	// AuthPassword is the default authentication password used by the testing
+	// charm store server if none is provided.
+	AuthPassword = "charmstore-testing-password"
+)
+
 // OpenServer instantiates a new charm store server instance.
```
rogpeppe, line 34: // If params.AuthUsername or params.AuthPassword are empty,
// AuthUsername and AuthPassword will be used.

?

frankban, line 34: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/886/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/887/
Test PASSed.

charmstoretesting/server.go
```diff
@@ -59,7 +73,17 @@ func (s *Server) Close() {
 	s.srv.Close()
 }
 
+func (s *Server) client() *csclient.Client {
```
rogpeppe: I wonder if it might be better to have:

```
// Client returns a new client that  will talk to the Server using basic
// (non-macaroon) authentication.
func (s *Server) NewClient() *csclient.Client
```

so that it's easy for tests to do this too.

I think you probably want:

```
func (s *Server) URL() string
```

so that tests can also create their own csclient.Client
easily with whatever params they want.

rogpeppe:
LGTM, thanks, modulo one last comment above.

charmstoretesting/server.go
frankban: Done.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/888/
Test PASSed.

PR #361, opened 2015-04-22
internal/charmstore: add /debug/check endpoint

mhilton:
Add a /debug/check endpoint to quickly assess health.

### DIFF
--- a/internal/charmstore/debug.go
+++ b/internal/charmstore/debug.go
@@ -4,7 +4,14 @@
 package charmstore
 
 import (
+	"fmt"
 	"net/http"
+	"sort"
+	"strings"
+	"time"
+
+	"gopkg.in/errgo.v1"
+	"gopkg.in/mgo.v2"
 
 	"gopkg.in/juju/charmstore.v4/internal/router"
 	appver "gopkg.in/juju/charmstore.v4/version"
@@ -15,8 +22,75 @@ func serveDebugInfo(http.Header, *http.Request) (interface{}, error) {
 	return appver.VersionInfo, nil
 }
 
-func newServiceDebugHandler() http.Handler {
+// GET /debug/check.
+func debugCheck(checks map[string]func() error) http.Handler {
+	return router.HandleJSON(func(http.Header, *http.Request) (interface{}, error) {
+		n := len(checks)
+		type result struct {
+			name string
+			err  error
+		}
+		c := make(chan result)
+		for name, check := range checks {
+			name, check := name, check
+			go func() {
+				c <- result{name: name, err: check()}
+			}()
+		}
+		results := make(map[string]string, n)
+		var failed bool
+		for ; n > 0; n-- {
+			res := <-c
+			if res.err == nil {
+				results[res.name] = "OK"
+			} else {
+				failed = true
+				results[res.name] = res.err.Error()
+			}
+		}
+		if failed {
+			keys := make([]string, 0, len(results))
+			for k := range results {
+				keys = append(keys, k)
+			}
+			sort.Strings(keys)
+			msgs := make([]string, len(results))
+			for i, k := range keys {
+				msgs[i] = fmt.Sprintf("[%s: %s]", k, results[k])
+			}
+			return nil, errgo.Newf("check failure: %s", strings.Join(msgs, " "))
+		}
+		return results, nil
+	})
+}
+
+func checkDB(db *mgo.Database) func() error {
+	return func() error {
+		s := db.Session.Copy()
+		s.SetSyncTimeout(500 * time.Millisecond)
+		defer s.Close()
+		return s.Ping()
+	}
+}
+
+func checkES(si *SearchIndex) func() error {
+	if si == nil || si.Database == nil {
+		return func() error {
+			return nil
+		}
+	}
+	return func() error {
+		_, err := si.Health()
+		return err
+	}
+}
+
+func newServiceDebugHandler(db *mgo.Database, si *SearchIndex) http.Handler {
 	mux := router.NewServeMux()
 	mux.Handle("/info", router.HandleJSON(serveDebugInfo))
+	mux.Handle("/check", debugCheck(map[string]func() error{
+		"mongodb":       checkDB(db),
+		"elasticsearch": checkES(si),
+	}))
 	return mux
 }
--- a/internal/charmstore/debug_test.go
+++ b/internal/charmstore/debug_test.go
@@ -1,0 +1,100 @@
+// Copyright 2015 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package charmstore
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/juju/testing/httptesting"
+	gc "gopkg.in/check.v1"
+
+	"gopkg.in/juju/charmstore.v4/internal/router"
+	"gopkg.in/juju/charmstore.v4/params"
+	appver "gopkg.in/juju/charmstore.v4/version"
+)
+
+type debugSuite struct{}
+
+var _ = gc.Suite(&debugSuite{})
+
+var debugCheckTests = []struct {
+	about        string
+	checks       map[string]func() error
+	expectStatus int
+	expectBody   interface{}
+}{{
+	about:        "no checks",
+	expectStatus: http.StatusOK,
+	expectBody:   map[string]string{},
+}, {
+	about: "passing check",
+	checks: map[string]func() error{
+		"pass": func() error { return nil },
+	},
+	expectStatus: http.StatusOK,
+	expectBody: map[string]string{
+		"pass": "OK",
+	},
+}, {
+	about: "failing check",
+	checks: map[string]func() error{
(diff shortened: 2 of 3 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/906/
Test PASSed.

internal/charmstore/debug.go
```diff
@@ -15,8 +16,46 @@ func serveDebugInfo(http.Header, *http.Request) (interface{}, error) {
 	return appver.VersionInfo, nil
 }
 
-func newServiceDebugHandler() http.Handler {
+// GET /debug/check.
+func debugCheck(checks ...func() error) http.Handler {
+	return router.HandleErrors(func(http.ResponseWriter, *http.Request) error {
+		for _, check := range checks {
+			err := check()
```
rogpeppe: This isn't going to give us a very useful error on failure.
Also, if one is down, we don't know if both are, and we probably should.

I wonder if the checks should be map[string]func() error
and we should go through them all and if any fail
produce an error message that includes the ones that
succeeded vs failed.

Also, it would be nice to have _some_ response from the
handler on success.

How about, on success:

   ["mongo ping", "es health"]

and on failure:

```
some checks failed: [mongo ping: ok] [es health: connection timeout]
```

what do you think?

mhilton:
## QA Instructions

```
% http -h :8080/debug/check
HTTP/1.1 200 OK
Content-Length: 38
Content-Type: application/json
Date: Thu, 23 Apr 2015 08:35:25 GMT

{
    elasticsearch: "OK"
    mongodb: "OK"
}
% sudo service mongodb stop
mongodb stop/waiting
% http -h :8080/debug/check
HTTP/1.1 500 Internal Server Error
Content-Length: 56
Content-Type: application/json
Date: Thu, 23 Apr 2015 08:35:41 GMT

{
    elasticsearch: "OK"
    mongodb: "no reachable servers"
}
% sudo service mongodb start
mongodb start/running, process 5927
% http -h :8080/debug/check
HTTP/1.1 200 OK
Content-Length: 38
Content-Type: application/json
Date: Thu, 23 Apr 2015 08:36:02 GMT

{
    elasticsearch: "OK"
    mongodb: "OK"
}
% sudo service elasticsearch stop
 * Stopping Elasticsearch Server
   ...done.
% http -h :8080/debug/check
HTTP/1.1 500 Internal Server Error
Content-Length: 122
Content-Type: application/json
Date: Thu, 23 Apr 2015 08:36:19 GMT

{
    elasticsearch: "Get http://localhost:9200/_cluster/health: dial tcp 127.0.0.1:9200: connection refused"
    mongodb: "OK"
}
% sudo service mongodb stop
mongodb stop/waiting
% http -h :8080/debug/check
HTTP/1.1 500 Internal Server Error
Content-Length: 140
Content-Type: application/json
Date: Thu, 23 Apr 2015 08:36:43 GMT

{
    elasticsearch: "Get http://localhost:9200/_cluster/health: dial tcp 127.0.0.1:9200: connection refused"
    mongodb: "no reachable servers"
}
% sudo service mongodb start
mongodb start/running, process 5997
% sudo service elasticsearch start
 * Starting Elasticsearch Server
   ...done.
% http -h :8080/debug/check
HTTP/1.1 200 OK
Content-Length: 38
Content-Type: application/json
Date: Thu, 23 Apr 2015 08:37:34 GMT

{
    elasticsearch: "OK"
    mongodb: "OK"
}
```

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/907/
Test PASSed.

internal/charmstore/debug.go
```diff
@@ -15,8 +18,75 @@ func serveDebugInfo(http.Header, *http.Request) (interface{}, error) {
...
+func debugCheck(checks map[string]func() error) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		n := len(checks)
+		type result struct {
+			name string
+			err  error
+		}
+		c := make(chan result)
+		for name, check := range checks {
+			go func(n string, check func() error) {
```
rogpeppe: I'd tend to prefer:

```
 for name, check := range checks {
      name, check := name, check
      go func() {
           c <- result{name: name, err: check()}
      }()
  }
```

as there's then no way you can get it wrong inside
the closure because the original names are hidden.
(and more concise and arguably easier to read,
once you're familiar with the idiom)

internal/charmstore/debug.go
```diff
@@ -15,8 +18,75 @@ func serveDebugInfo(http.Header, *http.Request) (interface{}, error) {
...
+				results[res.name] = res.err.Error()
+			}
+			n--
+		}
+		w.Header().Set("Content-Type", "application/json")
+		if failed {
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+		enc := json.NewEncoder(w)
+		if err := enc.Encode(results); err != nil {
```
rogpeppe: Although I like the idea of producing the same body regardless of
error or success, I do feel this somewhat breaks our otherwise universal
contract that if we produce an error, it looks like a params.Error.

That's why I suggested the error message form originally.

Another possibility would be to add a field to the Error return type
for this one call.

```
type error struct { 
   params.Error
   Info struct {
        Results map[string] string
   }
}
```

internal/charmstore/debug.go
```diff
@@ -15,8 +18,75 @@ func serveDebugInfo(http.Header, *http.Request) (interface{}, error) {
...
+		return s.Ping()
+	}
+}
+
+func checkES(si *SearchIndex) func() error {
+	if si == nil || si.Database == nil {
+		return func() error {
+			return nil
+		}
+	} else {
```
rogpeppe: lose the else to keep indentation smaller ?

internal/charmstore/debug.go
```diff
@@ -15,8 +18,75 @@ func serveDebugInfo(http.Header, *http.Request) (interface{}, error) {
...
+		return func() error {
+			_, err := si.Health()
+			if err != nil {
```
rogpeppe: return err ?

rogpeppe:
Looks great (with a couple of suggestions about). I think we should have at least one or two tests for this functionality though.

internal/charmstore/debug.go
mhilton: This is pretty academic as it's really only useful for things that are really only going to look at the status code anyway. But if we are producing a different output for the errors then I think we should produce no output on success. Humans should use /v1/debug/status.

internal/charmstore/debug.go
mhilton: Done: I didn't know that worked.

internal/charmstore/debug.go
mhilton: done

internal/charmstore/debug.go
mhilton: done

internal/charmstore/debug.go
```diff
@@ -15,8 +22,76 @@ func serveDebugInfo(http.Header, *http.Request) (interface{}, error) {
...
+			keys := make([]string, 0, len(results))
+			for k := range results {
+				keys = append(keys, k)
+			}
+			sort.Strings(keys)
+			msgs := make([]string, len(results))
+			for i, k := range keys {
+				msgs[i] = fmt.Sprintf("[%s: %s]", k, results[k])
+			}
+			return nil, errgo.New(strings.Join(msgs, " "))
```
rogpeppe: errno.Newf("check failure: %s", strings.Join(msgs, " ")) ?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/908/
Test PASSed.

internal/charmstore/debug.go
```diff
@@ -15,8 +22,76 @@ func serveDebugInfo(http.Header, *http.Request) (interface{}, error) {
...
+		c := make(chan result)
+		for name, check := range checks {
+			name, check := name, check
+			go func() {
+				c <- result{name: name, err: check()}
+			}()
+		}
+		results := make(map[string]string, n)
+		var failed bool
+		for n > 0 {
```
rogpeppe:
```
for ; n > 0; n-- {
}
```

?

internal/charmstore/debug.go
mhilton: done

rogpeppe:
LGTM modulo two minor comments. Ta!

internal/charmstore/debug.go
mhilton: done

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/909/
Test PASSed.

urosj:
:+1:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/910/
Test PASSed.

mhilton:
:shipit:

jujugui:
Status: merge request accepted. Url: http://localhost:8080/job/charmstore-merge

PR #397, opened 2015-06-11
internal/charmstore: limit asynchronous goroutine count

rogpeppe:
We make sure that the number of goroutines can't exceed
a reasonable set limit and make sure they've all completed
when the store is closed.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/juju/loggo"
+	"github.com/juju/utils/parallel"
 	"gopkg.in/errgo.v1"
 	"gopkg.in/juju/charm.v6-unstable"
 	"gopkg.in/juju/charmrepo.v0/csclient/params"
@@ -43,6 +44,7 @@ type Pool struct {
 	es           *SearchIndex
 	bakeryParams *bakery.NewServiceParams
 	stats        stats
+	run          *parallel.Run
 
 	// statsCache holds a cache of AggregatedCounts
 	// values, keyed by entity id. When the id has no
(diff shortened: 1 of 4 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1025/
Test PASSed.

internal/charmstore/store.go
```diff
@@ -215,7 +222,10 @@ func (p *Pool) requestStoreNB(always bool) (*Store, error) {
 		bp.Store = macStore
 		bsvc, err := bakery.NewService(bp)
 		if err != nil {
-			return nil, errgo.Notef(err, "cannot make bakery service")
+			// This should never happen because the only reason bakery.NewService
```
mhilton, line 225: This in not guaranteed for all future versions, but it is reasonable to deal with that at a later time.

mhilton:
:+1: LGTM

internal/charmstore/store.go
rogpeppe, line 225: One way of fixing that (backwardly incompatibly, unfortunately) is by changing bakery.NewService so it doesn't return an error, and provide a function to fill out bakery parameter defaults. There's really no reason why it should return an error, even in the future, as it doesn't interact with anything external. I may have to eat my words at some point in the future though :)

internal/charmstore/store.go
```diff
@@ -71,6 +73,10 @@ type Pool struct {
 // limit specified by config.MaxMgoSessions.
 const reqStoreCacheSize = 50
 
+// maxAsyncGoroutines holds the maximum number
+// of goroutines that will be started by Store.Go.
+const maxAsyncGoroutines = 50
```
urosj: If this is not part of the configuration, the max-mgo-sessions attribute is not fully functional.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1039/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1040/
Test PASSed.

urosj:
:+1: Awesome!

rogpeppe:
:shipit:

jujugui:

PR #412, opened 2015-07-17
Invalidate cache entry for meta stats on demand.

fabricematrat:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1090/
Test PASSed.

internal/cache/cache.go
```diff
@@ -100,7 +129,6 @@ func (c *Cache) getAtTime(key string, fetch func() (interface{}, error), now tim
 		value:  val,
 		expire: now.Add(c.maxAge - time.Duration(rand.Int63n(int64(c.maxAge/2)))),
 	}
-	return val, nil
```
urosj: Why remove this?

internal/cache/cache.go
```diff
@@ -73,6 +73,14 @@ func (c *Cache) Get(key string, fetch func() (interface{}, error)) (interface{},
 	return c.getAtTime(key, fetch, time.Now())
 }
 
+// Refresh returns the value for the given key, using fetch to fetch
+// the value and store the value in the cache refreshing any previous value.
+// If fetch returns an error, the returned error from Refresh will have
+// the same cause.
+func (c *Cache) Refresh(key string, fetch func() (interface{}, error)) (interface{}, error) {
```
rogpeppe: Rather than duplicating the whole signature of Get, how about just adding
a more specific Evict method (since we've already got EvictAll):

```
// Evict removes the entry with the given key from the cache.
func (c *Cache) Evict(key string)
```

This would make for simpler tests and code, I think.

internal/v4/api.go
```diff
@@ -614,8 +614,10 @@ func (h *ReqHandler) metaTags(entity *mongodoc.Entity, id *router.ResolvedURL, p
 // GET id/meta/stats/
 // https://github.com/juju/charmstore/blob/v4/docs/API.md#get-idmetastats
 func (h *ReqHandler) metaStats(entity *mongodoc.Entity, id *router.ResolvedURL, path string, flags url.Values, req *http.Request) (interface{}, error) {
+
 	// Retrieve the aggregated downloads count for the specific revision.
-	counts, countsAllRevisions, err := h.Store.ArchiveDownloadCounts(id.PreferredURL())
+	_, forceRefreshPresent := flags["force_refresh"]
```
rogpeppe: I think we should treat this as a boolean parameter like any other.
Also, how about just "refresh" rather than "force_refresh" ?
For example:

```
            refresh, err = router.ParseBool(flags.Get("refresh))
        if err != nil {
            return charmstore.SearchParams{}, badRequestf(err, "invalid refresh parameter")
        }
```

rogpeppe:
This looks OK, thanks, but I have a couple of suggestions to make things a little simpler and more consistent. Also, we should update the API docs (in docs/API.md) to reflect this new addition.

internal/cache/cache.go
fabricematrat: Done.

internal/v4/api.go
fabricematrat: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1091/
Test PASSed.

internal/cache/cache.go
```diff
@@ -57,6 +57,14 @@ func (c *Cache) Len() int {
 	return len(c.old) + len(c.new)
 }
 
+// Evict removes an entry from the cache if found based on
```
rogpeppe: // Evict removes the entry with the given key from the cache if present.

?

rogpeppe:
LGTM with one minor suggestion, thanks!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1092/
Test PASSed.

docs/API.md
```diff
@@ -1313,6 +1317,8 @@ type StatsCount struct {
 }
 ```
 
+Refresh boolean will allow to fetch the latest stats.
```
rogpeppe: If the refresh boolean parameter is non-zero, the latest stats will be returned without caching.

?

fabricematrat: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1093/
Test PASSed.

urosj:
:+1: Although i am thinking that with opening the "refresh" to all we are loosing the control over cache. We'll need to protect the endpoint I think.

fabricematrat:
:shipit: thanks @urosj && @rogpeppe

jujugui:

PR #417, opened 2015-08-04
denormalize ACL data to into entities.

mhilton:
ACL data is copied from base entities into entitites to prevent
excessive lookups being required for certain operations.


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1100/
Test PASSed.

mhilton:
## QA Instructions

You need to use mongo to notice any difference with this code as it is all to do with data representation.
1. Use http to set permissions on a charm. (meta/perm)
2. Use mongo to check that the juju.entities collection document corresponding to your charm has the correct ACL (as does everything else with the same base URL).

internal/charmstore/store.go
```diff
@@ -1127,9 +1157,29 @@ func (s *Store) findZipFile(blob io.ReadSeeker, size int64, isFile func(f *zip.F
...
+	if err := s.DB.BaseEntities().UpdateId(bid, bson.D{{"$set",
 		bson.D{{"acls." + which, acl}},
-	}})
+	}}); err != nil {
+		return errgo.NoteMask(err, "cannot update base entity", errgo.Is(mgo.ErrNotFound))
+	}
+	return s.CopyACLs(bid, nil)
+}
+
+// CopyACLs does a serverside copy of the ACLs from be into e (if
```
urosj: ... from BaseEntities (be) to Entities (e) ...

internal/router/fieldinclude.go
```diff
@@ -38,6 +39,11 @@ func (u *FieldUpdater) UpdateSearch() {
 	u.search = true
 }
 
+// CopyACLs requests that the ACL records of entities are updated.
+func (u *FieldUpdater) CopyACLs() {
```
rogpeppe: UpdateACLs ?

After all, the copying of ACLs is really just an implementation artifact - the information we're conveying here is that the base ACLs have changed, I think.

internal/router/fieldinclude.go
```diff
@@ -58,6 +64,10 @@ type FieldGetFunc func(doc interface{}, id *ResolvedURL, path string, flags url.
 // after the initial prefix has been removed.
 type FieldPutFunc func(id *ResolvedURL, path string, val *json.RawMessage, updater *FieldUpdater, req *http.Request) error
 
+// CopyACLsFunc copies the ACL from then base entity id into all entities
+// with that base entity.
+type CopyACLsFunc func(id *ResolvedURL) error
```
rogpeppe: similarly:

// UpdateACLsFunc updates the ACL information for the given URL.
type UpdateACLsFunc func(*ResolvedURL) error

?

internal/mongodoc/doc.go
```diff
@@ -108,6 +108,12 @@ type Entity struct {
 	// PromulgatedRevision holds the revision number from the promulgated URL.
 	// If the entity is not promulgated this should be set to -1.
 	PromulgatedRevision int `bson:"promulgated-revision"`
+
+	// ACLs holds permission information relevant to the base entity.
+	// It is identitcal to the same field in the relevent base
```
rogpeppe: s/identitcal/identical/

internal/charmstore/store.go
```diff
@@ -83,6 +84,22 @@ const reqStoreCacheSize = 50
 // of goroutines that will be started by Store.Go.
 const maxAsyncGoroutines = 50
 
+// jsACLCopyFunc is the text of the function run on the mongodb server to
+// copy ACLs from the base entity to an entity (or all entities) after an
+// update.
+const jsACLCopyFunc = `
+	function(base_entity, entity) {
+		b_col = db.getCollection(base_entities_collection_name)
```
rogpeppe: var b_col = 

I don't think we want to be setting global variables here.

internal/charmstore/store.go
```diff
@@ -83,6 +84,22 @@ const reqStoreCacheSize = 50
...
+		e_col = db.getCollection(entities_collection_name)
```
rogpeppe: ditto

internal/charmstore/store.go
```diff
@@ -83,6 +84,22 @@ const reqStoreCacheSize = 50
...
+		var be = b_col.findOne({_id: base_entity}, {acls: 1});
+		if (entity != null) {
```
rogpeppe: !==

mhilton: Good point.

internal/charmstore/migrations.go
```diff
@@ -222,3 +226,33 @@ func populateWriteACL(db StoreDatabase) error {
 	logger.Infof("%d base entities updated", counter)
 	return nil
 }
+
+// denormalizeEntityACLs updates all entities with an ACL copied from the
+// base entity. This function is not supposed to be called directly.
+func denormalizeEntityACLs(db StoreDatabase) error {
+	copyACLs := bson.JavaScript{
+		Code: jsACLCopyFunc,
+		Scope: bson.D{{
```
rogpeppe, line 235: Is it actually worth passing the scope here? Couldn't we just instantiate some JS that holds the collection names inline, given that the collection names never change?

rogpeppe:
LGTM I guess, but it makes me uneasy, and I'd like to see a concurrency test
where several entities are updated concurrently.

If we find ourselves doing much more of this, we should consider moving
to using an explicit lock collection like the original charmstore did when
creating entities.

internal/charmstore/migrations.go
mhilton, line 235: I don't know that doing that would change anything much, but it could be done. I think I'll leave it for now.

internal/charmstore/store.go

internal/charmstore/store.go

internal/charmstore/store.go

internal/charmstore/store.go

internal/mongodoc/doc.go
mhilton: done

internal/router/fieldinclude.go
mhilton: done

internal/router/fieldinclude.go
mhilton: done

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1101/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1102/
Test PASSed.

PR #423, opened 2015-08-27
List of update for stats/update endpoint instead of single one.

fabricematrat:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1122/
Test PASSed.

mhilton:
LGTM

internal/v4/stats.go
```diff
@@ -132,17 +132,35 @@ func (h *ReqHandler) serveStatsUpdate(_ http.Header, r *http.Request) (interface
...
 
-	rid, err := h.resolveURL(req.CharmReference)
-	if err != nil {
-		return nil, errgo.Mask(err, errgo.Is(params.ErrNotFound))
-	}
+	errors := make([]error, 0)
+	for i := range req {
+		rid, err := h.resolveURL(req[i].CharmReference)
+		if err != nil {
+			errors = append(errors, errgo.Notef(err, "cannot find entity for url: %s", req[i].CharmReference))
```
rogpeppe: Perhaps:
    errors = append(errors, errgo.Notef(err, "cannot find entity for %s", req[i].CharmReference))

?
(we usually use the : for underlying errors)

internal/v4/stats.go
```diff
@@ -132,17 +132,35 @@ func (h *ReqHandler) serveStatsUpdate(_ http.Header, r *http.Request) (interface
 		return nil, errgo.Notef(err, "cannot unmarshal body")
 	}
 
-	rid, err := h.resolveURL(req.CharmReference)
-	if err != nil {
-		return nil, errgo.Mask(err, errgo.Is(params.ErrNotFound))
-	}
+	errors := make([]error, 0)
```
rogpeppe, line 135: var errors []error

internal/v4/stats.go
```diff
@@ -132,17 +132,35 @@ func (h *ReqHandler) serveStatsUpdate(_ http.Header, r *http.Request) (interface
...
+			continue
+		}
 	}
 
+	if len(errors) != 0 {
+		messages := make([]string, len(errors))
+		for i, err := range errors {
+			messages[i] = err.Error()
+		}
+		encodedMessages, err := json.Marshal(messages)
```
rogpeppe: Rather than encoding the errors as JSON (which won't look very nice if we see the error), I guess we could use the params.Error.Info field to store an error for each entry and return that. Perhaps using the string form of the index number as the key. Alternatively, just keep the first error and the count and return a message based on that, e.g.

```
 cannot increment stats for trusty/wordpress: cannot resolve URL: charm not found (and 2134 more errors)
```

That probably fits best with what we actually want to see. We can log the other errors.

internal/v4/stats.go
```diff
@@ -132,17 +132,35 @@ func (h *ReqHandler) serveStatsUpdate(_ http.Header, r *http.Request) (interface
 		return nil, errgo.Notef(err, "cannot unmarshal body")
 	}
 
-	rid, err := h.resolveURL(req.CharmReference)
-	if err != nil {
-		return nil, errgo.Mask(err, errgo.Is(params.ErrNotFound))
-	}
+	errors := make([]error, 0)
+	for i := range req {
+		rid, err := h.resolveURL(req[i].CharmReference)
```
rogpeppe: Note that I don't think resolveURL does anything if the URL is already full specified, which means that it would be possible to game the system by adding stats for non-existent charms and revisions. Perhaps we should actually try to retrieve the charm?

internal/v4/stats.go
```diff
@@ -122,7 +122,7 @@ func (h *ReqHandler) serveStatsUpdate(_ http.Header, r *http.Request) (interface
 		return nil, errgo.WithCausef(nil, params.ErrMethodNotAllowed, "%s not allowed", r.Method)
 	}
 
-	var req params.StatsUpdateRequest
+	var req []params.StatsUpdateRequest
```
rogpeppe: Rather than have a slice here, we've been quite careful elsewhere to use
only a struct as an API parameter because that is more future-proof (it's easy
to add fields and maintain backward compatiblity). I suggest we change StatsUpdateRequest to be a struct holding a slice member.

That means that if we want to make things more efficient in the future (for example by passing a map[url][]timestamp) then that possibility remains open.

rogpeppe:
Good direction, thanks, with a few comments and suggestions above.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1123/
Test PASSed.

internal/v4/stats.go
fabricematrat: Done.

internal/v4/stats.go
fabricematrat: I will leave that one as we are already checking it in the Increment method.

internal/v4/stats.go
fabricematrat: Done.

internal/v4/stats.go
fabricematrat, line 135: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1124/
Test PASSed.

internal/v4/stats.go
fabricematrat: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1125/
Test PASSed.

internal/v4/stats_test.go
```diff
@@ -90,44 +90,109 @@ func (s *StatsSuite) TestServerStatsStatus(c *gc.C) {
...
+	c.Assert(err, gc.IsNil)
+
+	var countsBefore, countsAfter charmstore.AggregatedCounts
+	for i, test := range tests {
+		c.Logf("test %d. %s", i, test.path)
+
+		_, countsBefore, err = s.store.ArchiveDownloadCounts(ref, true)
+		c.Assert(err, gc.IsNil)
+
+		rec := httptesting.DoRequest(c, httptesting.DoRequestParams{
```
rogpeppe, line 141: You can still use AssertJSONCall for this. (that should probably be better documented in the AssertJSONCall doc comment)

rogpeppe:
LGTM with the thought that perhaps you can merge those two test functions given that you can use AssertJSONCall for both.

fabricematrat:
:shipit: thanks @mhilton && @rogpeppe.

jujugui:

PR #439, opened 2015-11-05
internal/charmstore: remove redundant migrations

rogpeppe:
We also refactor the insertEntity code so that it's more future-proof,
and record the expected entity fields at each version.


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1174/
Test FAILed.

PR #444, opened 2015-11-06
internal/charmstore: implement multi-series find-id logic

rogpeppe:
This implements the behaviour described in https://docs.google.com/document/d/1hiQ1aka40Jx9QdnXHXgtGzAZopIuAkDIO-pV-cZhASc .

The EntitiesQuery logic could be made more efficient but I've left it as is because I think it more clear this way, and the actual performance hit from the redundant condition checks is likely to be negligible.

Note that this also introduces a behaviour change into the API. A charm id with a revision but no specified series will not match old charms. If anyone was using that functionality previously, they were almost certainly doing it wrong.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -542,7 +542,7 @@ func (s *Store) AddCharm(c charm.Charm, p AddParams) (err error) {
 		CharmRequiredInterfaces: interfacesForRelations(c.Meta().Requires),
 		Contents:                p.Contents,
 	}
-	denormalizeEntity(entity)
+	DenormalizeEntity(entity)
 
 	// Check that we're not going to create a charm that duplicates
 	// the name of a bundle. This is racy, but it's the best we can do.
@@ -561,14 +561,17 @@ func (s *Store) AddCharm(c charm.Charm, p AddParams) (err error) {
 	return nil
 }
 
-// denormalizeEntity sets all denormalized fields in e
+// DenormalizeEntity sets all denormalized fields in e
 // from their associated canonical fields.
 //
 // It is the responsibility of the caller to set e.SupportedSeries
 // if the entity URL does not contain a series. If the entity
 // URL *does* contain a series, e.SupportedSeries will
 // be overwritten.
-func denormalizeEntity(e *mongodoc.Entity) {
+//
+// This is exported for the purposes of tests that
+// need to create directly into the database.
+func DenormalizeEntity(e *mongodoc.Entity) {
 	e.BaseURL = baseURL(e.URL)
 	e.Name = e.URL.Name
 	e.User = e.URL.User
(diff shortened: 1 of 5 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1191/
Test PASSed.

internal/charmstore/store.go
```diff
@@ -561,14 +561,17 @@ func (s *Store) AddCharm(c charm.Charm, p AddParams) (err error) {
...
 //
 // It is the responsibility of the caller to set e.SupportedSeries
 // if the entity URL does not contain a series. If the entity
 // URL *does* contain a series, e.SupportedSeries will
 // be overwritten.
-func denormalizeEntity(e *mongodoc.Entity) {
+//
+// This is exported for the purposes of tests that
+// need to create directly into the database.
+func DenormalizeEntity(e *mongodoc.Entity) {
```
urosj, line 574: Hm, do we really want to expose this?

internal/charmstore/store.go
```diff
@@ -719,40 +723,92 @@ var seriesScore = map[string]int{
...
+		return entities.Find(bson.D{{
+			"name", url.Name,
+		}, {
+			"user", url.User,
+		}, {
+			"supportedseries", url.Series,
+		}})
+	case url.User != "" && url.Revision != -1:
+		return entities.FindId(url)
+	case url.User != "":
```
urosj: I guess it's safe to not check that name is not empty as that's invalid entity.

internal/charmstore/store_test.go
```diff
@@ -1765,6 +1766,187 @@ func (s *StoreSuite) TestFindBestEntity(c *gc.C) {
...
+		SupportedSeries: []string{"precise"},
+	}, {
+		URL:             charm.MustParseReference("~charmers/wordpress-13"),
+		SupportedSeries: []string{"trusty"},
+	}, {
+		URL:             charm.MustParseReference("~bob/wordpress-14"),
+		SupportedSeries: []string{"precise"},
+	}},
+	url:       "~charmers/precise/wordpress",
+	expectURL: "~charmers/wordpress-12",
```
urosj: Shouldn't this be ~charmers/wordpress-13 ?

internal/charmstore/store.go
rogpeppe, line 574: As long as there are tests outside of internal/charmstore that rely on writing entities to the database directly, I think it's reasonable to do so. We're not exposing it to the outside world after all.

internal/charmstore/store.go
rogpeppe: Yup.

urosj:
:+1:

internal/charmstore/store.go
mhilton, line 574: I think this should probably only be exported in an export_test.go or similar.

internal/charmstore/store.go
```diff
@@ -673,6 +676,7 @@ func (s *Store) FindEntities(url *charm.Reference, fields ...string) ([]*mongodo
 // populated in the returned entities. If the given URL has no user then
 // only promulgated entities will be queried.
 func (s *Store) FindBestEntity(url *charm.Reference, fields ...string) (*mongodoc.Entity, error) {
+	logger.Infof("FindBestEntity %s", url)
```
mhilton: hmm, I'm not sure that this adds much value.

internal/charmstore/store.go
```diff
@@ -719,40 +723,92 @@ var seriesScore = map[string]int{
 	"utopic":  4,
 	"vivid":   5,
 	"wily":    6,
+	// When we find a multi-series charm (no series) we
+	// will always choose it in preference to a series-specific
+	// charm
+	"": 5000,
```
mhilton, line 728: MaxInt or something?

internal/charmstore/store.go
```diff
@@ -719,40 +723,92 @@ var seriesScore = map[string]int{
...
 // the produced query will only match promulgated entities.
 func (s *Store) EntitiesQuery(url *charm.Reference) *mgo.Query {
-	if url.User != "" && url.Series != "" && url.Revision != -1 {
-		// Find a specific owned entity, for instance ~who/utopic/django-42.
-		return s.DB.Entities().FindId(url)
-	}
-	if url.Series != "" && url.Revision != -1 {
-		// Find a specific promulgated entity, for instance utopic/django-42.
+	entities := s.DB.Entities()
+	// TODO this could be phrased more efficiently.
```
mhilton: This seems unnecessarily hard to follow. Which I expect is on the basis of some kind of theoretical optimizations. How about: 

``` go
query := bson.D{{
    "name", url.Name,
}}
if url.User == "" {
    if url.Revision > -1 {
        query = append(query, bson.DocElem{"promulgated-revision", url.Revision})
    } else {
        query = append(query, bson.DocElem{"promulgated-revision", bson.D{{"$gt", -1}}})
    }
} else {
    query = append(query, bson.DocElem{"user", url.User})
    if url.Revision > -1 {
        query = append(query, bson.DocElem{"revision", url.Revision})
    }
}
if url.Series == "" {
    query = append(query, bson.DocElem{"series", ""})
} else {
    query = append(query, bson.DocElem{"supportedseries", url.Series})
}
return entities.Find(query)
```

internal/v4/api_test.go
```diff
@@ -1345,8 +1345,12 @@ var serveExpandIdTests = []struct {
 		{Id: "cs:precise/haproxy-1"},
 	},
 }, {
-	about: "single result",
+	about: "revision with series doesn't match anything",
```
mhilton: without series? or with empty series.

internal/v4/archive_test.go
```diff
@@ -1448,52 +1437,37 @@ func (s *ArchiveSuite) TestGetNewPromulgatedRevision(c *gc.C) {
...
-			Series:              "trusty",
-			Revision:            i,
-			PromulgatedURL:      pid,
-			PromulgatedRevision: i,
-		})
+		entity := &mongodoc.Entity{
+			URL:            id,
+			PromulgatedURL: pid,
+		}
+		charmstore.DenormalizeEntity(entity)
```
mhilton, line 1469: Hmm, I think this is an unreasonable usecase to export Denormalise entity for. This test should use other store methods to add the charms it needs.

mhilton:
I don't think I'm quite +1 on this yet, The EntitiesQuery function is so much more convoluted than seems to be needed I'd like to have a chat about it before it lands.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1193/
Test PASSed.

internal/v4/archive_test.go
rogpeppe, line 1469: If it's OK, can we leave off fixing this test for a separate PR?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1194/
Test PASSed.

mhilton:
LGTM with DenormalizeEntity no longer exported, There is certainly a better way to make that test work.

internal/charmstore/store.go
rogpeppe, line 728: Given that we've got a known set of integers enumerated here, maxint seems like overkill to me.

rogpeppe:
:shipit:

rogpeppe:
:shipit:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1195/
Test PASSed.

PR #449, opened 2015-11-11
V5 terms

alesstimec:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1209/
Test PASSed.

elasticsearch/elasticsearch.go
```diff
@@ -334,7 +334,7 @@ func (db *Database) Search(index, type_ string, q QueryDSL) (SearchResult, error
 }
 
 // do performs a request on the elasticsearch server. If body is not nil it will be
-// marsheled as a json object and sent with the request. If v is non nil the response
+// marshaled as a json object and sent with the request. If v is non nil the response
```
mhilton, line 337: good spot, thanks

mhilton:
:+1:

internal/v4/auth.go
```diff
@@ -75,6 +76,46 @@ func (h *ReqHandler) authorize(req *http.Request, acl []string, alwaysAuth bool,
...
+
+		_, verr := httpbakery.CheckRequest(h.Store.Bakery, req, nil, checkers.OperationChecker("get-archive"))
+		if verr == nil {
+			return nil
+		}
+		if _, ok := errgo.Cause(verr).(*bakery.VerificationError); !ok {
+			return errgo.Mask(verr)
+		}
+		bk := h.Store.Bakery
+		if bk == nil || h.handler.config.TermsLocation == "" {
```
fabricematrat: h.handler.config.TermsLocation can't be "" as it is checked line 92 ?

internal/v4/auth.go
```diff
@@ -75,6 +76,46 @@ func (h *ReqHandler) authorize(req *http.Request, acl []string, alwaysAuth bool,
...
+			return errgo.WithCausef(verr, params.ErrUnauthorized, "authentication failed")
+		}
+		m, err := h.Store.Bakery.NewMacaroon("", nil, []checkers.Caveat{
+			checkers.Caveat{h.handler.config.TermsLocation, fmt.Sprintf("has-agreed %s", strings.Join(entity.CharmMeta.Terms, ","))},
+			checkers.AllowCaveat("get-archive"),
+		})
+		if err != nil {
+			return errgo.Mask(err)
+		}
+		return httpbakery.NewDischargeRequiredError(m, "/", verr)
```
rogpeppe: This should use NewDischargeRequiredErrorForRequest, otherwise the protocol is not friendly to browsers.

internal/v4/auth.go
fabricematrat: if bk was nill it would have been an issue for line 96  I think
I would remove the whole if test maybe ?

internal/v4/auth.go
fabricematrat: use NewDischargeRequiredErrorForRequest like line 76

fabricematrat:
LGTM  with a few suggestions

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1216/
Test PASSed.

internal/v4/auth.go
```diff
@@ -75,6 +76,43 @@ func (h *ReqHandler) authorize(req *http.Request, acl []string, alwaysAuth bool,
...
+		}
+
+		_, verr := httpbakery.CheckRequest(bk, req, nil, checkers.OperationChecker("get-archive"))
+		if verr == nil {
+			return nil
+		}
+		if _, ok := errgo.Cause(verr).(*bakery.VerificationError); !ok {
+			return errgo.Mask(verr)
+		}
+		m, err := bk.NewMacaroon("", nil, []checkers.Caveat{
```
rogpeppe: It seems a bit of a pity that we'll be generating two macaroons that both guard access to the same entity when theoretically one would do. Apart from the fact that it implies more round trips, this increases the likelihood of running out of client cookiejar space.

I'm also not that keen that we're calling CheckRequest twice and substantially duplicating the logic in checkRequest. 

I think that perhaps we should add support for authorising specified operations into AuthorizeEntity, so there's only one place that generates macaroons. This might require a bit of thought.

rogpeppe:
This is a good direction, thanks. I like the idea of terms-and-conditions as a third party service, but I think we need to consider this implementation a little more. Also, AFAICS this won't work when the environment downloads the charm, as is the default case - we'll need support in delegatable-macaroon for that.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1314/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1319/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1322/
Test PASSed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1358/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1361/
Test PASSed.

cmd/charmd/main.go
```diff
@@ -94,7 +97,37 @@ func serve(confPath string) error {
 	}
 
 	ring := bakery.NewPublicKeyRing()
-	ring.AddPublicKeyForLocation(cfg.IdentityLocation, false, conf.IdentityPublicKey)
+	if conf.IdentityPublicKey != nil {
```
rogpeppe: How about factoring out the duplicated code below into a separate function, say:

```
func addPublicKey(ring bakery.PublicKeyRing, loc string, key *bakery.PublicKey) error
```

?

or alternatively, you could use httpbakery.PublicKeyRing.

e.g. 

```
cache := bakery.NewPublicKeyRing()
if cfg.IdentityPublicKey != nil {
    cache.AddPublicKeyForLocation(conf.IdentityLocation, false, conf.IdentityPublicKey)
}
cfg.PublicKeyLocator = httpbakery.NewPublicKeyRing(http.DefaultClient, cache)
```

docs/API.md
```diff
@@ -2057,21 +2057,16 @@ Request body:
...
 prior authorization is required.
 
 #### GET /delegatable-macaroon
 
 This endpoint returns a macaroon in JSON format that can be passed to
 third parties to allow them to access the charm store on the user's
-behalf.  A first party "is-entity" caveat may be added to restrict those
-parties so that they can only access a given charmstore entity with a
-specified id.
+behalf. The entity id must be added as a url encoded parameter and is added as a first party "is-entity" caveat to restrict the usability of the returned macaroon, so that third parties can only access a given charmstore entity with a the specified id.
```
rogpeppe: This doesn't say what the name of the URL-encoded parameter is.
Let's document the URL parameters similarly to other examples in this
doc (e.g. in GET meta/endpoint).

Also, I'm not sure we need to discuss the exact caveats that are added here - that's
really an implementation detail (they are _first_-party caveats after all).

Perhaps simply something like:

"If the id parameter is specified, the returned macaroon will be restricted for
use only with the entity with the given id."

?

cmd/charmd/main.go
```diff
@@ -73,10 +74,12 @@ func serve(confPath string) error {
...
-		AuthUsername:            conf.AuthUsername,
-		AuthPassword:            conf.AuthPassword,
-		IdentityLocation:        conf.IdentityLocation,
-		IdentityAPIURL:          conf.IdentityAPIURL,
+		AuthUsername:     conf.AuthUsername,
+		AuthPassword:     conf.AuthPassword,
+		IdentityLocation: conf.IdentityLocation,
+		IdentityAPIURL:   conf.IdentityAPIURL,
+		TermsLocation:    conf.TermsLocation,
+
```
mhilton: Why break here?

docs/API.md
```diff
@@ -2057,21 +2057,16 @@ Request body:
...
 
-A delegatable macaroon will only be returned to an authorized user (not
-including admin). It will carry the same privileges as the macaroon used
-to authorize the request.
+A delegatable macaroon will only be returned to an authorized user (not including admin). It will carry the same privileges as the macaroon used to authorize the request.
```
rogpeppe: ... but is suitable for use by third parties.

?

(otherwise why would you bother using it?)

cmd/charmd/main.go
mhilton: This and the part below look almost identical, perhaps they could be made into a function.

docs/API.md
mhilton: It would be nice if you could keep the length of these lines to around 80 chars.

docs/API.md
```diff
@@ -2057,21 +2057,16 @@ Request body:
 
 #### GET /macaroon
 
-This endpoint returns a macaroon in JSON format that, when its third
-party caveats are discharged, will allow access to the charm store. No
+This endpoint returns a macaroon in JSON format that, when its third party caveats are discharged, will allow access to the charm store. No
```
rogpeppe: Let's leave this this with reasonable lengths please.

docs/API.md
rogpeppe: Also, please leave the line lengths reasonable here too.

internal/v5/api.go
```diff
@@ -1107,21 +1107,66 @@ func (h *ReqHandler) serveMacaroon(_ http.Header, _ *http.Request) (interface{},
...
+			checkers.DeclaredCaveat(UsernameAttr, auth.Username),
+			checkers.TimeBeforeCaveat(time.Now().Add(DelegatableMacaroonExpiry)),
+			checkers.DenyCaveat(opAccessCharmWithTerms),
+		})
+		if err != nil {
+			return nil, errgo.Mask(err)
+		}
+		return m, nil
+	}
+	var resolvedURLs []*router.ResolvedURL
```
rogpeppe: resolvedURLs := make([]*router.ResolvedURL, len(entityIds))

to save a few allocations?

internal/v5/auth.go
```diff
@@ -19,8 +21,10 @@ import (
 )
 
 const (
-	basicRealm        = "CharmStore4"
-	PromulgatorsGroup = "charmers"
+	PromulgatorsGroup      = "charmers"
+	opAccessCharmWithTerms = "access-operation"
```
rogpeppe: Let's add some comments here saying what these operations signify, please.

Also, how about using shorter names, as these will end up in cookies, and space is
at a premium there? e.g. op-access, op-other ?

cmd/charmd/main.go
alesstimec: done.

docs/API.md
alesstimec: done

cmd/charmd/main.go
alesstimec: done

docs/API.md
alesstimec: done

internal/v5/api.go
alesstimec: done

internal/v5/auth.go
alesstimec: done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1368/
Test PASSed.

mhilton:
:+1: This is looking really good, thanks.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1372/
Test PASSed.

cmd/charmd/config.yaml
```diff
@@ -20,3 +20,5 @@ identity-api-url: http://localhost:8081
 #stats-cache-max-age: 1h
 #request-timeout: 500ms
 #search-cache-max-age: 0s
+# Uncomment to test with a terms service running locally
+#terms-location: localhost:8081
```
cmars: @alesstimec Let's change this to 8085

rogpeppe: +1

internal/v4/common_test.go
```diff
@@ -76,6 +77,10 @@ type commonSuite struct {
 	idM        *idM
 	idMServer  *httptest.Server
 
+	dischargeTerms  func(cav, arg string) ([]checkers.Caveat, error)
```
rogpeppe: I thought we weren't putting terms-specific stuff in the v4 tests?

internal/v5/archive_test.go
```diff
@@ -37,12 +40,36 @@ import (
 	"gopkg.in/juju/charmstore.v5-unstable/internal/v5"
 )
 
-type ArchiveSuite struct {
+type commonArchiveSuite struct {
```
rogpeppe, line 43: Do we actually need another ArchiveSuite (and hence commonArchiveSuite) ?

I'd suggest (if it works OK) just having a single ArchiveSuite - the main reason why this might not work ok is if any terms stuff needed to be set up before SetUpTest is called, but that doesn't appear to be the case AFAICS.

internal/v5/archive_test.go
```diff
@@ -1729,3 +1756,125 @@ func (s *ArchiveSearchSuite) TestGetSearchUpdate(c *gc.C) {
...
+
+	archiveUrl := storeURL("~charmers/precise/terms-0/archive")
+	rec := httptesting.DoRequest(c, httptesting.DoRequestParams{
+		Handler: s.srv,
+		URL:     archiveUrl,
+		Do:      bakeryDo(client),
+	})
+	c.Assert(rec.Code, gc.Equals, http.StatusOK)
+	c.Assert(rec.Body.Bytes(), gc.DeepEquals, archiveBytes)
+	c.Assert(rec.Header().Get(params.ContentHashHeader), gc.Equals, hashOfBytes(archiveBytes))
```
rogpeppe: I don't think we need to duplicate the whole of TestGet here - the download logic is the same in each case, so we actually only need to check that we can get the archive.
Given that, we can probably delete lines 1800 to 1816 here.

internal/v5/archive_test.go
```diff
@@ -1729,3 +1756,125 @@ func (s *ArchiveSearchSuite) TestGetSearchUpdate(c *gc.C) {
...
+	s.enableIdentity = true
+}
+
+func (s *ArchiveSuiteWithTerms) SetUpTest(c *gc.C) {
+	s.commonSuite.SetUpTest(c)
+	s.discharge = dischargeForUser("bob")
+}
+
+func (s *ArchiveSuiteWithTerms) TestGetUserHasAgreedToTermsAndConditions(c *gc.C) {
+	s.dischargeTerms = func(s1, s2 string) ([]checkers.Caveat, error) {
```
rogpeppe: Let's check that that the third party caveat being checked is the one we're expecting.

rogpeppe: And also let's leave something around to assure us that the terms discharge checker _is_ in fact being called in this case.

internal/v5/archive_test.go
```diff
@@ -1729,3 +1756,125 @@ func (s *ArchiveSearchSuite) TestGetSearchUpdate(c *gc.C) {
...
+	archiveUrl := storeURL("~charmers/precise/terms-0/archive")
+	rec := httptesting.DoRequest(c, httptesting.DoRequestParams{
+		Handler: s.srv,
+		URL:     archiveUrl,
+		Header:  basicAuthHeader(testUsername, testPassword),
+	})
+	c.Assert(rec.Code, gc.Equals, http.StatusOK)
+	c.Assert(rec.Body.Bytes(), gc.DeepEquals, archiveBytes)
+	c.Assert(rec.Header().Get(params.ContentHashHeader), gc.Equals, hashOfBytes(archiveBytes))
+	c.Assert(rec.Header().Get(params.EntityIdHeader), gc.Equals, "cs:~charmers/precise/terms-0")
```
rogpeppe: Again, no need to duplicate all of TestGet.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1381/
Test PASSed.

internal/v5/archive_test.go
```diff
@@ -1729,3 +1756,125 @@ func (s *ArchiveSearchSuite) TestGetSearchUpdate(c *gc.C) {
...
+	assertCacheControl(c, rec.Header(), true)
```
rogpeppe: And again.
Perhaps consider making a assertArchiveDownload function that can be called by all these tests. Then it would be reasonable for the range checking to be in that, as it wouldn't be duplicated.

internal/v5/auth.go
```diff
@@ -19,8 +20,13 @@ import (
 )
 
 const (
-	basicRealm        = "CharmStore4"
 	PromulgatorsGroup = "charmers"
+	// opAccessCharmWitTerms indicates an operation of accessing the archive of
+	// a charm that requires agreement to certain terms and conditions.
+	opAccessCharmWithTerms = "op-get-with-terms"
+	// opOther indicated all other operations.
```
rogpeppe: s/indicated/indicates/

cmd/charmd/main.go
```diff
@@ -105,6 +116,17 @@ func serve(confPath string) error {
 	return http.ListenAndServe(conf.APIAddr, debug.Handler("", server))
 }
 
+func addPublicKey(ring *bakery.PublicKeyRing, loc string, key *bakery.PublicKey) error {
+	if key != nil {
+		return ring.AddPublicKeyForLocation(loc, false, key)
+	}
+	pubKey, err := httpbakery.PublicKeyForLocation(http.DefaultClient, loc)
```
rogpeppe, line 124: comment still to be addressed here.

docs/API.md
```diff
@@ -2057,26 +2057,27 @@ Request body:
...
 
 #### GET /delegatable-macaroon
 
 This endpoint returns a macaroon in JSON format that can be passed to
 third parties to allow them to access the charm store on the user's
-behalf.  A first party "is-entity" caveat may be added to restrict those
-parties so that they can only access a given charmstore entity with a
-specified id.
+behalf. The entity id must be added as a url encoded parameter and is added as
+a first party "is-entity" caveat to restrict the usability of the returned
```
rogpeppe: comment still to be addressed here.

internal/v5/api.go
```diff
@@ -1120,21 +1120,65 @@ func (h *ReqHandler) serveMacaroon(_ http.Header, _ *http.Request) (interface{},
 // GET /delegatable-macaroon
 // See https://github.com/juju/charmstore/blob/v4/docs/API.md#get-delegatable-macaroon
 func (h *ReqHandler) serveDelegatableMacaroon(_ http.Header, req *http.Request) (interface{}, error) {
+	values, err := url.ParseQuery(req.URL.RawQuery)
+	if err != nil {
+		return nil, errgo.Mask(err)
+	}
+	entityIds := values["id"]
+	if len(entityIds) == 0 {
+		auth, err := h.authorize(req, []string{params.Everyone}, true, nil)
```
rogpeppe, line 1126: // No entity ids, so we provide a macaroon that's good for any entity that the
// user can access, as long as that entity doesn't have terms and conditions.

?

internal/v5/archive.go
```diff
@@ -108,6 +108,10 @@ func (h *ReqHandler) authorizeUpload(id *charm.URL, req *http.Request) error {
 }
 
 func (h *ReqHandler) serveGetArchive(id *router.ResolvedURL, w http.ResponseWriter, req *http.Request) error {
+	_, err := h.authorizeEntityAndTerms(req, []*router.ResolvedURL{id})
+	if err != nil {
+		return err
```
rogpeppe: errgo.Mask(err, errgo.Any)

internal/v5/archive_test.go
```diff
@@ -37,12 +40,36 @@ import (
...
 }
 
+type ArchiveSuite struct {
+	commonArchiveSuite
+}
+
 var _ = gc.Suite(&ArchiveSuite{})
 
+func (s *ArchiveSuite) TestGetCharmWithTerms(c *gc.C) {
+	s.dischargeTerms = func(s1, s2 string) ([]checkers.Caveat, error) {
```
rogpeppe: If the terms service isn't configured, why does this need to be set?

internal/v5/auth.go
```diff
@@ -64,10 +75,127 @@ func (h *ReqHandler) authorize(req *http.Request, acl []string, alwaysAuth bool,
...
+		if baseEntity == nil {
+			return authorization{}, errgo.WithCausef(nil, params.ErrNotFound, "could not find the base entity %v", entityId.URL.String())
+		}
+
+		ACLs[i] = baseEntity.ACLs.Read
+		if entityId.Development {
+			ACLs[i] = baseEntity.DevelopmentACLs.Read
+		}
+
+		if (entity.CharmMeta == nil) || len(entity.CharmMeta.Terms) == 0 {
```
rogpeppe: if entity.CharmMeta == nil || len(entity.CharmMeta.Terms) == 0 {

internal/v5/auth.go
```diff
@@ -64,10 +75,127 @@ func (h *ReqHandler) authorize(req *http.Request, acl []string, alwaysAuth bool,
...
+		}
+		if entity == nil {
+			return authorization{}, errgo.WithCausef(nil, params.ErrNotFound, "could not find entity %q", entityId.String())
+		}
+
+		baseEntity, err := h.Store.FindBaseEntity(&entityId.URL, "acls", "developmentacls")
+		if err != nil {
+			return authorization{}, errgo.Mask(err, errgo.Is(params.ErrNotFound))
+		}
+		if baseEntity == nil {
```
rogpeppe: This can never happen.

internal/v5/auth.go
```diff
@@ -64,10 +75,127 @@ func (h *ReqHandler) authorize(req *http.Request, acl []string, alwaysAuth bool,
...
+
+	ACLs := make([][]string, len(entityIds))
+	requiredTerms := make(map[string]struct{})
+	allowAuth := true
+	for i, entityId := range entityIds {
+		entity, err := h.Store.FindEntity(entityId)
+		if err != nil {
+			return authorization{}, errgo.Mask(err, errgo.Is(params.ErrNotFound))
+		}
+		if entity == nil {
```
rogpeppe: This can never happen.

docs/API.md
alesstimec: ?

cmd/charmd/main.go
alesstimec, line 124: ?

internal/v5/auth.go
alesstimec: done

internal/v5/api.go
alesstimec, line 1126: done

internal/v5/archive.go
alesstimec: done

internal/v5/archive_test.go
alesstimec: done

internal/v5/auth.go
alesstimec: done

internal/v5/auth.go
alesstimec: famous last word, but done

internal/v5/auth.go
alesstimec: done

internal/v5/archive_test.go
alesstimec: done

internal/v5/archive_test.go
alesstimec: done

cmd/charmd/main.go
cmars, line 124: @alesstimec I think @rogpeppe is referring to comments he made on #491 after it landed yesterday.

cmars:
@alesstimec The handler constructor func in [internal/v5/api.go](https://github.com/juju/charmstore/blob/297e453ca1ba27cc7a1571063f937dff789e285a/internal/v5/api.go#L91) needs to use the public key locator set up in main -- otherwise this won't use the configured terms service public key. 

Update: #496

cmars:
However, for our QA environment, we will still need to either:
- Land go-macaroon-bakery/macaroon-bakery#86 and use it here, to allow cached public key lookups to succeed for HTTP locations. I think this may be reasonable for public keys exchanged out-of-band as service configuration.
- Add a configuration option to the charmstore that sets httpbakery.PublicKeyRing.AllowInsecure().

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1387/
Test PASSed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1395/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1398/
Test PASSed.

cmd/charmd/main.go
rogpeppe, line 124: Yes - I didn't realise it had merged when I was reviewing.

internal/v4/common_test.go
```diff
@@ -142,6 +143,7 @@ func (s *commonSuite) startServer(c *gc.C) {
 		StatsCacheMaxAge: time.Nanosecond,
 		MaxMgoSessions:   s.maxMgoSessions,
 	}
+	keyring := bakery.NewPublicKeyRing()
```
rogpeppe, line 146: or just use httpbakery.NewPublicKeyRing - then you won't need to explicitly add the public key below.

internal/v5/archive_test.go
```diff
@@ -1729,3 +1746,125 @@ func (s *ArchiveSearchSuite) TestGetSearchUpdate(c *gc.C) {
...
+
+	client := httpbakery.NewHTTPClient()
+
+	id := newResolvedURL("cs:~charmers/precise/terms-0", -1)
+	s.assertUploadCharm(c, "POST", id, "terms")
+	err := s.store.SetPerms(&id.URL, "read", "bob")
+	c.Assert(err, gc.IsNil)
+
+	archiveUrl := storeURL("~charmers/precise/terms-0/archive")
+	//s.assertArchiveDownload(c, "~charmers/precise/terms-0/archive", id.URL.String(), nil)
```
rogpeppe: Why not use assertArchiveDownload here too?

internal/v5/archive_test.go
```diff
@@ -1729,3 +1746,125 @@ func (s *ArchiveSearchSuite) TestGetSearchUpdate(c *gc.C) {
...
+	rec := httptesting.DoRequest(c, httptesting.DoRequestParams{
+		Handler: s.srv,
+		URL:     archiveUrl,
+		Do:      bakeryDo(client),
+	})
+	c.Assert(rec.Code, gc.Equals, http.StatusOK)
+
+	c.Assert(termsDischargeAccessed, gc.Equals, true)
+
+	// Check that the HTTP range logic is plugged in OK. If this
```
rogpeppe: This range logic testing doesn't seem necessary here, as mentioned before.

internal/v5/archive_test.go
alesstimec: done.

internal/v5/archive_test.go
alesstimec: done.

internal/v5/archive_test.go
alesstimec: done.

internal/v5/auth.go
```diff
@@ -64,10 +76,120 @@ func (h *ReqHandler) authorize(req *http.Request, acl []string, alwaysAuth bool,
...
+		entityIds)
+
+	if len(entityIds) == 0 {
+		return authorization{}, errgo.WithCausef(nil, params.ErrUnauthorized, "entity id not specified")
+	}
+
+	ACLs := make([][]string, len(entityIds))
+	requiredTerms := make(map[string]struct{})
+	allowAuth := true
+	for i, entityId := range entityIds {
```
rogpeppe: This is quite a long function and this is perhaps a coherent section of it that might be removed into a separate function, making its purpose and side-effects more clear.

Something like:

```
// entityAuthInfo returns authorization on the entities with the given ids.
// If public is true, no authorization is required, otherwise acls holds
// an entry for each id with the corresponding ACL for each entity,
// and requiredTerms holds entries for all required terms.
func (h *ReqHandler) entityAuthInfo(entityIds []*router.ResolvedURL) (public bool, acls [][]string, requiredTerms map[string]bool, err error)
```

perhaps?

internal/v5/archive_test.go
```diff
@@ -124,6 +141,21 @@ func (s *ArchiveSuite) TestGetDevelopment(c *gc.C) {
 	})
 }
 
+func (s *commonSuite) assertArchiveDownload(c *gc.C, url, id string, archiveBytes []byte) {
+	rec := httptesting.DoRequest(c, httptesting.DoRequestParams{
+		Handler: s.srv,
+		URL:     storeURL(url),
+	})
+	c.Assert(rec.Code, gc.Equals, http.StatusOK)
+	if id != "" {
```
rogpeppe: When would this ever be empty? Don't we want to check this always?

internal/v5/archive_test.go
```diff
@@ -37,12 +40,33 @@ import (
...
+	archiveUrl := storeURL("~charmers/precise/terms-0/archive")
+	rec := httptesting.DoRequest(c, httptesting.DoRequestParams{
+		Handler: s.srv,
+		URL:     archiveUrl,
+		Do:      bakeryDo(client),
+	})
+	c.Assert(rec.Code, gc.Equals, http.StatusUnauthorized)
+}
+
 func (s *ArchiveSuite) TestGet(c *gc.C) {
```
rogpeppe, line 70: This test should probably also use assertArchiveDownload.

internal/v5/archive_test.go
alesstimec: it's a test function.. sometimes, i guess, we do not need to check this..  (it wasn't in the original cs code)

internal/v5/archive_test.go
```diff
@@ -124,6 +141,21 @@ func (s *ArchiveSuite) TestGetDevelopment(c *gc.C) {
 	})
 }
 
+func (s *commonSuite) assertArchiveDownload(c *gc.C, url, id string, archiveBytes []byte) {
```
rogpeppe: Given you have this function now, I think it should be easy to check
the archive contents always (and perhaps even do the range check).
How about:

```
 // assertArchiveDownload asserts that the given charm archive can be
 // downloaded from the path with the given id, and that the resulting
 // EntityIdHeader is equal to expectId. If expectId is empty, id will be
 // used instead.
 assertArchiveDownload(c, id string, archiveArchive *charm.CharmArchive, expectId string)
```

We can append the "/archive" suffix to id within the function.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1405/
Test PASSed.

internal/v4/common_test.go
alesstimec, line 146: done.

internal/v5/archive_test.go
alesstimec, line 70: done, i believe

cmars:
:+1:

cmd/charmd/config.yaml
```diff
@@ -5,12 +5,12 @@ auth-username: admin
...
-identity-location: http://localhost:8081/v1/discharger
-identity-api-url: http://localhost:8081
+#identity-public-key: CIdWcEUN+0OZnKW9KwruRQnQDY/qqzVdD30CijwiWCk=
+#identity-location: http://localhost:8081/v1/discharger
+#identity-api-url: http://localhost:8081
 # For production identity manager.
-#identity-public-key: hmHaPgCC1UfuhYHUSX5+aihSAZesqpVdjRv0mgfIwjo=
-#identity-location: https://api.jujucharms.com/identity/v1/discharger
+identity-public-key: hmHaPgCC1UfuhYHUSX5+aihSAZesqpVdjRv0mgfIwjo=
+identity-location: https://api.jujucharms.com/identity/v1/discharger
```
urosj, line 13: Leave defaults pointing to localhost, not the production one, please.

cmd/charmd/config.yaml
```diff
@@ -20,3 +20,5 @@ identity-api-url: http://localhost:8081
 #stats-cache-max-age: 1h
 #request-timeout: 500ms
 #search-cache-max-age: 0s
+# Uncomment to test with a terms service running locally
```
urosj, line 23: Move this up after identity, as this is external service specification. In the end we have charm store attributes only.

alesstimec:
$$merge$$

alesstimec:
:shipit:

jujugui:

PR #453, opened 2015-11-12
Support multi-series charms in elasticsearch.

mhilton:
Ensure mutli-series charms update elasticsearch correctly and that search works correctly.

### DIFF
--- a/internal/charmstore/search.go
+++ b/internal/charmstore/search.go
@@ -66,6 +66,7 @@ type SearchDoc struct {
 	*mongodoc.Entity
 	TotalDownloads int64
 	ReadACLs       []string
+	Series         []string
 }
 
 // UpdateSearchAsync will update the search record for the entity
@@ -210,6 +211,11 @@ func (s *Store) searchDocFromEntity(e *mongodoc.Entity, be *mongodoc.BaseEntity)
 		return nil, errgo.Mask(err)
 	}
 	doc.TotalDownloads = allRevisions.Total
+	if doc.Entity.Series == "bundle" {
+		doc.Series = []string{"bundle"}
+	} else {
+		doc.Series = doc.Entity.SupportedSeries
+	}
 	return &doc, nil
 }
 
@@ -229,6 +235,21 @@ func (si *SearchIndex) update(doc *SearchDoc) error {
 		doc)
 	if err != nil && err != elasticsearch.ErrConflict {
 		return errgo.Mask(err)
+	}
+	if doc.Entity.URL.Series != "" {
+		return nil
+	}
+	// This document represents a multi-series charm. It might be
+	// replacing existing documents for previous single-series
+	// charms. Remove any documents that this replaces.
+	for _, series := range doc.Entity.SupportedSeries {
+		u := *doc.Entity.URL
+		u.Series = series
+		err := si.DeleteDocument(si.Index, typeName, si.getID(&u))
+		if err != nil && errgo.Cause(err) != elasticsearch.ErrNotFound {
+			u.Revision = -1
+			logger.Errorf("cannot remove old search document for %q: %s", u, err)
+		}
 	}
 	return nil
 }
--- a/internal/charmstore/search_test.go
+++ b/internal/charmstore/search_test.go
@@ -11,8 +11,10 @@ import (
 
 	jc "github.com/juju/testing/checkers"
 	gc "gopkg.in/check.v1"
+	"gopkg.in/errgo.v1"
 	"gopkg.in/juju/charmrepo.v1/csclient/params"
 
+	"gopkg.in/juju/charmstore.v5-unstable/elasticsearch"
 	"gopkg.in/juju/charmstore.v5-unstable/internal/mongodoc"
 	"gopkg.in/juju/charmstore.v5-unstable/internal/router"
 	"gopkg.in/juju/charmstore.v5-unstable/internal/storetesting"
@@ -92,6 +94,7 @@ func (s *StoreSearchSuite) TestSuccessfulExport(c *gc.C) {
 			Entity:         entity,
 			TotalDownloads: int64(charmDownloadCounts[name]),
 			ReadACLs:       readACLs,
+			Series:         entity.SupportedSeries,
 		}
 		c.Assert(string(actual), jc.JSONEquals, doc)
 	}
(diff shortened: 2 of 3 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1220/
Test PASSed.

internal/charmstore/search.go
```diff
@@ -210,6 +211,11 @@ func (s *Store) searchDocFromEntity(e *mongodoc.Entity, be *mongodoc.BaseEntity)
 		return nil, errgo.Mask(err)
 	}
 	doc.TotalDownloads = allRevisions.Total
+	if len(doc.Entity.SupportedSeries) == 0 {
```
rogpeppe: This will only be true for bundles, right?

Wouldn't it be more obvious to make the check specific to "bundle" ?

internal/charmstore/search.go
```diff
@@ -230,6 +236,21 @@ func (si *SearchIndex) update(doc *SearchDoc) error {
 	if err != nil && err != elasticsearch.ErrConflict {
 		return errgo.Mask(err)
 	}
+	if doc.Entity.URL.Series != "" {
+		return nil
+	}
+	// This document represents a multi-series charm it might be
```
rogpeppe: s/charm it/charm. It/

internal/charmstore/search.go
```diff
@@ -230,6 +236,21 @@ func (si *SearchIndex) update(doc *SearchDoc) error {
...
+	// replacing existing documents for previous single-series
+	// charms. Remove any documents that this replaces.
+	for _, series := range doc.Entity.SupportedSeries {
```
rogpeppe, line 245: Is this actually sufficient? What if there's an old "precise" charm but the multi-series charm that's been uploaded doesn't support precise?

internal/charmstore/search_test.go
```diff
@@ -130,7 +133,37 @@ func (s *StoreSearchSuite) TestExportOnlyLatest(c *gc.C) {
...
+}
+
+func (s *StoreSearchSuite) TestExportMultiSeriesCharmsReplaceEarlierOnes(c *gc.C) {
+	charmArchive := storetesting.Charms.CharmDir("wordpress")
+	url := newResolvedURL("cs:~charmers/trusty/juju-gui-24", -1)
+	err := s.store.AddCharmWithArchive(url, charmArchive)
+	c.Assert(err, gc.IsNil)
+	charmArchive = storetesting.Charms.CharmDir("juju-gui")
+	url = newResolvedURL("cs:~charmers/juju-gui-25", -1)
+	err = s.store.AddCharmWithArchive(url, charmArchive)
```
rogpeppe, line 151: Shoudn't this fail because the uploaded charm doesn't have any supported series in its metadata? I'd like to see a test where there are multiple previous entities with different series and the latest added charm doesn't support all of them.

internal/charmstore/search_test.go
```diff
@@ -804,3 +837,75 @@ func (s *StoreSearchSuite) TestUpdateConflict(c *gc.C) {
...
+	bURL := baseURL(&url.URL)
+	baseEntity, err := s.store.FindBaseEntity(bURL)
+	c.Assert(err, gc.IsNil)
+	baseEntity.ACLs.Read = append(baseEntity.ACLs.Read, params.Everyone)
+	err = s.store.DB.BaseEntities().UpdateId(baseEntity.URL, baseEntity)
+	c.Assert(err, gc.IsNil)
+	err = s.store.UpdateSearch(url)
+	c.Assert(err, gc.IsNil)
+	s.store.ES.Database.RefreshIndex(s.TestIndex)
+	res, err := s.store.Search(SearchParams{
```
rogpeppe: table?

internal/charmstore/search_test.go
```diff
@@ -804,3 +837,75 @@ func (s *StoreSearchSuite) TestUpdateConflict(c *gc.C) {
...
+func (s *StoreSearchSuite) TestMultiSeriesCharmSortsSeriesCorrectly(c *gc.C) {
+	charmArchive := storetesting.Charms.CharmDir("juju-gui")
+	url := newResolvedURL("cs:~charmers/juju-gui-25", -1)
+	err := s.store.AddCharmWithArchive(url, charmArchive)
+	c.Assert(err, gc.IsNil)
+	bURL := baseURL(&url.URL)
+	baseEntity, err := s.store.FindBaseEntity(bURL)
+	c.Assert(err, gc.IsNil)
+	baseEntity.ACLs.Read = append(baseEntity.ACLs.Read, params.Everyone)
+	err = s.store.DB.BaseEntities().UpdateId(baseEntity.URL, baseEntity)
```
rogpeppe: Why do we need to do this? And given that we're doing it, why not do it at the same level as other tests and use s.store.SetPerms ?

internal/charmstore/store_test.go
```diff
@@ -1167,7 +1167,7 @@ func (s *StoreSuite) TestAddCharmWithBundleSeries(c *gc.C) {
 	c.Assert(err, gc.ErrorMatches, `charm added with invalid id cs:~charmers/bundle/wordpress-2`)
 }
 
-func (s *StoreSuite) TestAddCharmWithMultipleSeries(c *gc.C) {
+func (s *StoreSuite) TestAddCharmWithMultiSeries(c *gc.C) {
 	store := s.newStore(c, false)
 	defer store.Close()
 	ch := storetesting.Charms.CharmArchive(c.MkDir(), "juju-gui")
```
rogpeppe: Perhaps a comment saying that the juju-gui charm supports many series?
Or even name the charm such that it's obvious that it supports multiple series?

rogpeppe:
Looks good in general, thanks, but some of the logic seems potentially suspect.

internal/charmstore/search.go
mhilton, line 245: I think in that case you should still see the precise search result as that is what you would install for precise. There is also the notion of deprecated series which means that eventually those charms would not be added to search in the first place.

internal/charmstore/search_test.go
mhilton, line 151: To the question - no because the juju-gui charm does have supported series in the metadata, the wordpress one does not, but that has a series in the URL.
To the second point, I'll add a test.

internal/charmstore/search.go
mhilton: good idea, done

internal/charmstore/search.go

internal/charmstore/search_test.go
mhilton: We need to make the charm readable by "everybody" otherwise it doesn't show up in search results. but it should be done using SetPerms you're right.

internal/charmstore/store_test.go
mhilton: done as discussed in other PRs

internal/charmstore/search_test.go
mhilton, line 151: done

internal/charmstore/search_test.go
mhilton: done

internal/charmstore/search_test.go
mhilton: done

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1249/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1250/
Test PASSed.

internal/charmstore/search.go
```diff
@@ -230,6 +236,21 @@ func (si *SearchIndex) update(doc *SearchDoc) error {
 	if err != nil && err != elasticsearch.ErrConflict {
 		return errgo.Mask(err)
 	}
+	if doc.Entity.URL.Series != "" {
+		return nil
```
urosj: Why?

internal/charmstore/search.go
```diff
@@ -230,6 +236,21 @@ func (si *SearchIndex) update(doc *SearchDoc) error {
...
+	// This document represents a multi-series charm. It might be
+	// replacing existing documents for previous single-series
+	// charms. Remove any documents that this replaces.
+	for _, series := range doc.Entity.SupportedSeries {
+		u := *doc.Entity.URL
+		u.Series = series
+		err := si.DeleteDocument(si.Index, typeName, si.getID(&u))
+		if err != nil && errgo.Cause(err) != elasticsearch.ErrNotFound {
+			u.Revision = -1
+			logger.Warningf("cannot remove old search document for %q: %s", u.String(), err)
```
urosj: I'd say this is an error and not a warning. It puts ES into inconsistent state.

urosj:
:+1: with upgraded warning/error notice

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1251/
Test PASSed.

internal/charmstore/search.go
```diff
@@ -230,6 +236,21 @@ func (si *SearchIndex) update(doc *SearchDoc) error {
...
+	// This document represents a multi-series charm. It might be
+	// replacing existing documents for previous single-series
+	// charms. Remove any documents that this replaces.
+	for _, series := range doc.Entity.SupportedSeries {
+		u := *doc.Entity.URL
+		u.Series = series
+		err := si.DeleteDocument(si.Index, typeName, si.getID(&u))
+		if err != nil && errgo.Cause(err) != elasticsearch.ErrNotFound {
+			u.Revision = -1
+			logger.Errorf("cannot remove old search document for %q: %s", u.String(), err)
```
rogpeppe: s/.String()//

internal/charmstore/search_test.go
```diff
@@ -804,3 +873,60 @@ func (s *StoreSearchSuite) TestUpdateConflict(c *gc.C) {
...
+	c.Assert(err, gc.IsNil)
+	err = s.store.UpdateSearch(url)
+	c.Assert(err, gc.IsNil)
+	s.store.ES.Database.RefreshIndex(s.TestIndex)
+	var sp SearchParams
+	sp.ParseSortFields("-series")
+	res, err := s.store.Search(sp)
+	c.Assert(err, gc.IsNil)
+	c.Assert(len(res.Results) > 1, gc.Equals, true, gc.Commentf("expecting more than one result %#v", res.Results))
+	c.Assert(res.Results[0].URL.String(), gc.Equals, url.String())
```
rogpeppe: Shouldn't this really test that all results are actually sorted by series?

rogpeppe:
LGTM modulo a couple of trivials

internal/charmstore/search_test.go
mhilton: The sorting is tested elsewhere. This is more to test that if you have more than one series in a document it gets sorted into the correct place.

internal/charmstore/search.go

internal/charmstore/search_test.go
mhilton: done

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1253/
Test PASSed.

mhilton:
:shipit: thanks for the reviews

jujugui:

PR #451, opened 2015-11-12
insert charms with multiple series

mhilton:

PR #454, opened 2015-11-12
add multi-series charm check

mhilton:
When a charm with a series is added if will fail if there is a
multi-series charm with the same name.

@urosj was correct, this should have been in #451 but that wasn't so obvious to me at the time.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -549,7 +549,9 @@ func (s *Store) AddCharm(c charm.Charm, p AddParams) (err error) {
 	denormalizeEntity(entity)
 
 	// Check that we're not going to create a charm that duplicates
-	// the name of a bundle. This is racy, but it's the best we can do.
+	// the name of a bundle. This is racy, but it's the best we can
+	// do. Also check that there isn't an existing multi-series charm
+	// that would be replaced by this one.
 	entities, err := s.FindEntities(entity.BaseURL)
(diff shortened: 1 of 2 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1221/
Test PASSed.

frankban:
:+1: thank you!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1246/
Test PASSed.

internal/charmstore/store.go
```diff
@@ -549,7 +549,9 @@ func (s *Store) AddCharm(c charm.Charm, p AddParams) (err error) {
 	denormalizeEntity(entity)
 
 	// Check that we're not going to create a charm that duplicates
-	// the name of a bundle. This is racy, but it's the best we can do.
+	// the name of a bundle. This is racy, but it's the best we can
```
urosj, line 552: s/This is racy/There is a possibility of a race condition maybe?

urosj:
:+1:

internal/charmstore/store.go
mhilton, line 552: I'm going to leave this as it was the comment that was already there, just reformatted.

mhilton:
:shipit:

jujugui:

PR #456, opened 2015-11-12
mongodoc: introduce development revisions and development ACLs.

frankban:
Also implement the corresponding migration.

### DIFF
--- a/internal/charmstore/migrations.go
+++ b/internal/charmstore/migrations.go
@@ -12,7 +12,8 @@ import (
 )
 
 const (
-	migrationAddSupportedSeries mongodoc.MigrationName = "add supported series"
+	migrationAddSupportedSeries        mongodoc.MigrationName = "add supported series"
+	migrationHandleDevelopmentEntities mongodoc.MigrationName = "handle development entities"
 )
 
 // migrations holds all the migration functions that are executed in the order
@@ -39,6 +40,9 @@ var migrations = []migration{{
 }, {
 	name:    migrationAddSupportedSeries,
 	migrate: addSupportedSeries,
+}, {
+	name:    migrationHandleDevelopmentEntities,
+	migrate: handleDevelopmentEntities,
 }}
 
 // migration holds a migration function with its corresponding name.
@@ -134,6 +138,36 @@ func addSupportedSeries(db StoreDatabase) error {
 	}
 	if err := iter.Close(); err != nil {
 		return errgo.Notef(err, "cannot iterate entities")
+	}
+	return nil
+}
+
+// handleDevelopmentEntities adds the Development field to all entities and
+// sets up ACLs for development revisions.
+func handleDevelopmentEntities(db StoreDatabase) error {
+	logger.Infof("adding development field to all entities")
+	if _, err := db.Entities().UpdateAll(nil, bson.D{{
+		"$set", bson.D{{"development", false}},
+	}}); err != nil {
+		return errgo.Notef(err, "cannot add development field to all entities")
+	}
+
+	logger.Infof("adding development ACLs to all base entities")
+	baseEntities := db.BaseEntities()
+	var baseEntity mongodoc.BaseEntity
+	iter := baseEntities.Find(bson.D{{
+		"developmentacls", bson.D{{"$exists", false}},
+	}}).Select(bson.D{{"_id", 1}, {"acls", 1}}).Iter()
+	defer iter.Close()
+	for iter.Next(&baseEntity) {
+		if err := baseEntities.UpdateId(baseEntity.URL, bson.D{{
+			"$set", bson.D{{"developmentacls", baseEntity.ACLs}},
+		}}); err != nil {
+			return errgo.Notef(err, "cannot add development ACLs to base entity id %s", baseEntity.URL)
+		}
+	}
+	if err := iter.Close(); err != nil {
+		return errgo.Notef(err, "cannot iterate base entities")
 	}
 	return nil
 }
--- a/internal/charmstore/migrations.go
+++ b/internal/charmstore/migrations.go
@@ -12,8 +12,9 @@ import (
 )
 
 const (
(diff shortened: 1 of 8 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1232/
Test FAILed.

internal/charmstore/migrations.go
```diff
@@ -138,6 +142,36 @@ func addSupportedSeries(db StoreDatabase) error {
 	return nil
 }
 
+// handleDevelopmentEntities adds the Development field to all entities and
```
mhilton: handle doesn't seem to be the correct verb here perhaps it should be two migrations, addDevelopment & addDevelopmentACLs

internal/charmstore/migrations.go
```diff
@@ -138,6 +142,36 @@ func addSupportedSeries(db StoreDatabase) error {
...
+// sets up ACLs for development revisions.
+func handleDevelopmentEntities(db StoreDatabase) error {
+	logger.Infof("adding development field to all entities")
+	if _, err := db.Entities().UpdateAll(nil, bson.D{{
```
mhilton: This should probably only update those that don't have a development field already. That way it's idempotent and its fairly easy to do.

internal/charmstore/migrations.go
```diff
@@ -138,6 +142,36 @@ func addSupportedSeries(db StoreDatabase) error {
...
+		"$set", bson.D{{"development", false}},
+	}}); err != nil {
+		return errgo.Notef(err, "cannot add development field to all entities")
+	}
+
+	logger.Infof("adding development ACLs to all base entities")
+	baseEntities := db.BaseEntities()
+	var baseEntity mongodoc.BaseEntity
+	iter := baseEntities.Find(bson.D{{
+		"developmentacls", bson.D{{"$exists", false}},
```
mhilton, line 169: It should be possible to do this without retrieving every document locally, but it isn't so... moan

internal/charmstore/store_test.go
```diff
@@ -1119,6 +1126,15 @@ func (s *StoreSuite) TestAddUserOwnedCharmArchive(c *gc.C) {
 	s.checkAddCharm(c, charmArchive, false, newResolvedURL("~charmers/precise/wordpress-1", -1))
 }
 
+func (s *StoreSuite) TestAddDevelopmentCharmArchive(c *gc.C) {
+	charmArchive := storetesting.Charms.CharmArchive(c.MkDir(), "wordpress")
+	url := newResolvedURL("~charmers/precise/wordpress-1", 1)
```
mhilton, line 1131: perhaps add a boolean to newResolvedURL for development as that function is only used for tests, although it's used in lots of them.

internal/mongodoc/doc.go
```diff
@@ -146,11 +151,13 @@ type BaseEntity struct {
...
 	Public bool
 
-	// ACLs holds permission information relevant to
-	// the base entity. The permissions apply to all
-	// revisions.
+	// ACLs holds permission information relevant to the base entity.
+	// The permissions apply to all revisions.
 	ACLs ACL
 
+	// DevelopmentACLs is similar to ACLs but applies to development revisions.
```
mhilton: all development revisions?

mhilton:
LGTM with a few thoughts.

bac:
Hi Francesco, the feature branch problem with CI is that it was using 'make build' but the Makefile has:

PROJECT := gopkg.in/juju/charmstore.v5-unstable
PROJECT_DIR := $(shell go list -e -f '{{.Dir}}' $(PROJECT))

Since the default branch 'v5-unstable' is hard-coded in the Makefile using it for feature branches does not work.

An easy fix would be to have 'go list' compare to the project parent:

PROJECT_DIR := $(shell go list -e -f '{{.Dir}}' $(PROJECT)/..)

That symbol is only used to test to ensure the cwd is on the GOPATH so it would not affect anything else.  Make this change as a drive-by if you want.  Otherwise the Makefile is useless for any work outside of the hard-code branch.

bac:
Another problem.  v5-unstable is hardcoded in imports like:

```
"gopkg.in/juju/charmstore.v5-unstable/elasticsearch"
```

internal/charmstore/migrations.go
frankban: Done.

internal/charmstore/migrations.go
frankban: Good idea, done.

internal/charmstore/migrations.go
frankban, line 169: I am not aware of a way to do that.

internal/charmstore/store_test.go
frankban, line 1131: I'd prefer to wait a bit for the proper fix and not change the function signature and therefore the function calls everywhere, just to change them back later.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1241/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1243/
Test PASSed.

internal/charmstore/migrations.go
mhilton, line 169: it's not obviously possible with mongodb

internal/charmstore/store_test.go
mhilton, line 1131: sounds good

urosj:
:+1: 

QA: migration passed, also, result of this was empty:

db.base_entities.aggregate([{
    $project : {
        equal : {
            $eq : ["$acls", "$developmentacls"] 
        },
        doc : "$$ROOT"  
    }
}, {
    $match : {
        equal : false  
    }
}]);

frankban:
:shipit: 
Thanks Martin and Uros for the reviews!

jujugui:

PR #476, opened 2015-11-18
Implement the publish API.

frankban:
Use the proper permission checks when uploading charms or bundle.
Allow uploading and publishing at the same time.

This branch should complete the publishing work.
Later we will add promulgation policy as requested by eco.
In essence, on promulgating we can set the published write ACLs for
the entity to promulgators only, so that we don't allow
user publishing of promugated charms by default.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -263,6 +263,56 @@ Request body:
 }
 ```
 
+### Charm and bundle publishing
+
+#### PUT *id*/publish
+
+A PUT to ~*user*/*anyseries*/*name*-*anyrevision* sets whether the
+corresponding charm or bundle is published and can be accessed through a URL
+with no channel. If the revision number is not specified, the id is resolved to
+the charm or bundle with the latest development revision number. The id must
+not include the development channel.
+
+```go
+type PublishRequest struct {
+    Published bool
+}
+```
+
+If Published is true, the charm or bundle is made available at the
+non-development URL with the same revision number. If Published is false, the
+id is unpublished.
+
+The response include the id and promulgated id of the entity after the action
+is performed:
+
+```go
+type PublishResponse struct {
+    Id            *charm.URL
+    PromulgatedId *charm.URL `json:",omitempty"`
+}
+```
+
+If the charm or bundle have been unpublished, the identifiers in the response
+will represent a development charm or bundle.
+
+Example: `PUT ~charmers/trusty/django-42/publish`
+
+Request body:
+```json
+{
+    "Publish" : true,
+}
+```
+
+Response body:
+```json
+{
+    "Id" : "cs:~charmers/trusty/django-42",
+    "PromulgatedId": "cs:trusty/django-10",
+}
+```
+
 ### Stats
 
 #### GET stats/counter/...
@@ -1555,7 +1605,7 @@ Example: `GET ~bob/trusty/wordpress-42/meta/id-series`
 
 The meta/common-info path reports any common metadata recorded for the base
 entity. This contains only information stored by clients - the API server
-itself does not populate any fields. The resulting object holds an entry for 
+itself does not populate any fields. The resulting object holds an entry for
 each piece of metadata recorded with a PUT to `meta/common-info`.
 
 ```go
--- a/docs/API.md
+++ b/docs/API.md
@@ -270,8 +270,9 @@ Request body:
 A PUT to ~*user*/*anyseries*/*name*-*anyrevision* sets whether the
 corresponding charm or bundle is published and can be accessed through a URL
 with no channel. If the revision number is not specified, the id is resolved to
-the charm or bundle with the latest development revision number. The id must
-not include the development channel.
+the charm or bundle with the latest development revision number when
+publishing, and to the charm or bundle with the latest non-development revision
+number when unpublishing. The id must not include the development channel.
 
 ```go
 type PublishRequest struct {
@@ -283,7 +284,7 @@ If Published is true, the charm or bundle is made available at the
 non-development URL with the same revision number. If Published is false, the
 id is unpublished.
 
-The response include the id and promulgated id of the entity after the action
+The response includes the id and promulgated id of the entity after the action
 is performed:
 
 ```go
@@ -294,14 +295,14 @@ type PublishResponse struct {
 ```
 
 If the charm or bundle have been unpublished, the identifiers in the response
-will represent a development charm or bundle.
+will represent the corresponding development charm or bundle.
 
 Example: `PUT ~charmers/trusty/django-42/publish`
 
 Request body:
 ```json
 {
-    "Publish" : true,
+    "Published" : true,
 }
 ```
 
--- a/internal/v4/api.go
+++ b/internal/v4/api.go
@@ -1177,8 +1177,8 @@ func (h *ReqHandler) servePublish(id *charm.URL, w http.ResponseWriter, req *htt
 	if req.Method != "PUT" {
 		return errgo.WithCausef(nil, params.ErrMethodNotAllowed, "%s not allowed", req.Method)
 	}
-	if id.Channel == charm.DevelopmentChannel {
-		return errgo.WithCausef(nil, params.ErrForbidden, "cannot publish or unpublish development charm or bundle %q", id)
+	if id.Channel != "" {
+		return errgo.WithCausef(nil, params.ErrForbidden, "can only set publish on published URL, %q provided", id)
 	}
 	jsonContentType := "application/json"
 	if ct := req.Header.Get("Content-Type"); ct != jsonContentType {
--- a/internal/v4/api.go
+++ b/internal/v4/api.go
@@ -1180,15 +1180,12 @@ func (h *ReqHandler) servePublish(id *charm.URL, w http.ResponseWriter, req *htt
 	if id.Channel != "" {
 		return errgo.WithCausef(nil, params.ErrForbidden, "can only set publish on published URL, %q provided", id)
 	}
-	jsonContentType := "application/json"
-	if ct := req.Header.Get("Content-Type"); ct != jsonContentType {
-		return errgo.WithCausef(nil, params.ErrBadRequest, "unexpected Content-Type %q; expected %q", ct, jsonContentType)
-	}
 
 	// Retrieve the requested action from the request body.
-	var publish params.PublishRequest
-	decoder := json.NewDecoder(req.Body)
-	if err := decoder.Decode(&publish); err != nil {
+	var publish struct {
+		params.PublishRequest `httprequest:",body"`
+	}
+	if err := httprequest.Unmarshal(httprequest.Params{Request: req}, &publish); err != nil {
 		return errgo.WithCausef(err, params.ErrBadRequest, "cannot unmarshal publish request body")
 	}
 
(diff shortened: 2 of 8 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1302/
Test PASSed.

internal/v4/archive.go
```diff
@@ -70,32 +70,38 @@ func (h *ReqHandler) serveArchive(id *charm.URL, w http.ResponseWriter, req *htt
...
+		}
+		return nil
 	}
 	if errgo.Cause(err) != params.ErrNotFound {
 		return errgo.Notef(err, "cannot retrieve entity %q for authorization", id)
 	}
 	// The base entity does not currently exist, so we default to
 	// assuming write permissions for the entity user.
-	return h.authorizeWithPerms(req, nil, []string{id.User}, nil)
+	return errgo.Mask(h.authorizeWithPerms(req, nil, []string{id.User}, nil), errgo.Any)
```
mhilton: won't this mask nil in confusing ways?

mhilton:
I have a feeling, although I haven't tried it, that search will need some finagling in order to work with the idea of unpublishing properly. The edge cases here are sufficiently hard that I would advocate that unpublishing shouldn't be allowed. Apart from anything else what if someone is using the published charm? Can you add tests for the search index for all of your scenarios to make sure it behaves sensibly?

internal/v4/archive.go
frankban: No, it will just return nil in that case.

frankban:
Unpublishing of something breaks all existing scripts relying on that thing. It should be a very exceptional operation only allowed to charm store admins. This is similar to DELETE id/archive. Search can be confused by those two, and this is a corner case indeed, but I think we should fix search later rather than completely disallow a possible emergency operation.

urosj:
Getting to broken charm is equally bad as providing results and then 404 when you actually want to see or use that charm. We do need to fix this in new PRs.

mhilton:
:+1:

docs/API.md
```diff
@@ -263,6 +263,56 @@ Request body:
...
+
+If the charm or bundle have been unpublished, the identifiers in the response
+will represent a development charm or bundle.
+
+Example: `PUT ~charmers/trusty/django-42/publish`
+
+Request body:
+```json
+{
+    "Publish" : true,
```
urosj: "Published" : true

docs/API.md
```diff
@@ -263,6 +263,56 @@ Request body:
...
+type PublishRequest struct {
+    Published bool
+}
+```
+
+If Published is true, the charm or bundle is made available at the
+non-development URL with the same revision number. If Published is false, the
+id is unpublished.
+
+The response include the id and promulgated id of the entity after the action
```
rogpeppe: s/include/includes/

docs/API.md
```diff
@@ -263,6 +263,56 @@ Request body:
...
+
+```go
+type PublishResponse struct {
+    Id            *charm.URL
+    PromulgatedId *charm.URL `json:",omitempty"`
+}
+```
+
+If the charm or bundle have been unpublished, the identifiers in the response
+will represent a development charm or bundle.
```
rogpeppe: s/a/the corresponding/ ?

internal/v4/api.go
```diff
@@ -144,9 +144,10 @@ func newReqHandler() *ReqHandler {
 			"diagram.svg": h.resolveId(h.authId(h.serveDiagram)),
 			"expand-id":   h.resolveId(h.authId(h.serveExpandId)),
 			"icon.svg":    h.resolveId(h.authId(h.serveIcon)),
+			"promulgate":  h.resolveId(h.serveAdminPromulgate),
```
rogpeppe, line 147: thanks for the reorder.

internal/v4/api.go
```diff
@@ -1169,6 +1170,61 @@ func (h *ReqHandler) serveAdminPromulgate(id *router.ResolvedURL, w http.Respons
...
 }
 
+// PUT id/publish
+// See https://github.com/juju/charmstore/blob/v4/docs/API.md#put-idpublish
+func (h *ReqHandler) servePublish(id *charm.URL, w http.ResponseWriter, req *http.Request) error {
+	// Perform basic validation of the request.
+	if req.Method != "PUT" {
+		return errgo.WithCausef(nil, params.ErrMethodNotAllowed, "%s not allowed", req.Method)
+	}
+	if id.Channel == charm.DevelopmentChannel {
```
rogpeppe: if id.Channel != "" {

might be a little more future-proof. (the error message might be "can only set publish on published URL" or something like that)

internal/v4/api.go
```diff
@@ -1169,6 +1170,61 @@ func (h *ReqHandler) serveAdminPromulgate(id *router.ResolvedURL, w http.Respons
...
+	}
+	jsonContentType := "application/json"
+	if ct := req.Header.Get("Content-Type"); ct != jsonContentType {
+		return errgo.WithCausef(nil, params.ErrBadRequest, "unexpected Content-Type %q; expected %q", ct, jsonContentType)
+	}
+
+	// Retrieve the requested action from the request body.
+	var publish params.PublishRequest
+	decoder := json.NewDecoder(req.Body)
+	if err := decoder.Decode(&publish); err != nil {
```
rogpeppe: We should have a separate function to do this, or just use httprequest.Unmarshal.
For example, the content type is not parsed properly here and probably should be.

rogpeppe: e.g.

```
var publish struct {
    params.PublishRequest `httprequest:",body"`
}
err := httprequest.Unmarshal(httprequest.Params{Request: req}, &publish)
```

would probably do the job.

docs/API.md
```diff
@@ -263,6 +263,56 @@ Request body:
...
 ```
 
+### Charm and bundle publishing
+
+#### PUT *id*/publish
+
+A PUT to ~*user*/*anyseries*/*name*-*anyrevision* sets whether the
+corresponding charm or bundle is published and can be accessed through a URL
+with no channel. If the revision number is not specified, the id is resolved to
+the charm or bundle with the latest development revision number. The id must
```
rogpeppe: Looking at the code, this doesn't appear to be true when unpublishing - in that case the latest published revision number is chosen. If we want that to be the case (seems kinda reasonable, I guess) then we should document that.

internal/v4/api_test.go
```diff
@@ -2935,6 +2935,438 @@ func (s *APISuite) TestEndpointRequiringBaseEntityWithPromulgatedId(c *gc.C) {
...
+		newResolvedURL("~who/development/trusty/django-42", 10),
+		newResolvedURL("~who/trusty/django-47", 11),
+		newResolvedURL("~who/development/trusty/django-48", 12),
+		newResolvedURL("~who/development/trusty/django-49", 13),
+	},
+	expectBody: params.PublishResponse{
+		Id:            charm.MustParseURL("~who/development/trusty/django-48"),
+		PromulgatedId: charm.MustParseURL("development/trusty/django-12"),
+	},
+}}
```
rogpeppe, line 3221: Nice tests

internal/v4/archive.go
```diff
@@ -70,32 +70,38 @@ func (h *ReqHandler) serveArchive(id *charm.URL, w http.ResponseWriter, req *htt
...
 	if id.User == "" {
 		return badRequestf(nil, "user not specified in entity upload URL %q", id)
 	}
+	baseEntity, err := h.Store.FindBaseEntity(id, "acls", "developmentacls")
 	// Note that we pass a nil entity URL to authorizeWithPerms, because
 	// we haven't got a resolved URL at this point. At some
 	// point in the future, we may want to be able to allow
 	// is-entity first-party caveats to be allowed when uploading
 	// at which point we will need to rethink this a little.
-	baseURL := *id
```
rogpeppe, line 86: good catch

internal/v4/archive.go
```diff
@@ -70,32 +70,38 @@ func (h *ReqHandler) serveArchive(id *charm.URL, w http.ResponseWriter, req *htt
...
-	baseURL.Series = ""
-	baseEntity, err := h.Store.FindBaseEntity(id, "acls")
 	if err == nil {
-		return h.authorizeWithPerms(req, baseEntity.ACLs.Read, baseEntity.ACLs.Write, nil)
+		if err := h.authorizeWithPerms(req, baseEntity.DevelopmentACLs.Read, baseEntity.DevelopmentACLs.Write, nil); err != nil {
+			return errgo.Mask(err, errgo.Any)
+		}
+		// If uploading a published entity, also check that the user has
+		// publishing permissions.
+		if id.Channel != charm.DevelopmentChannel {
```
rogpeppe: id.Channel == "" ?

internal/v4/archive.go
rogpeppe: That's true (and by design) but these days I generally consider it cleaner to explicitly mask if not nil. _shrug_

internal/v4/archive.go
```diff
@@ -170,7 +176,17 @@ func (h *ReqHandler) servePostArchive(id *charm.URL, w http.ResponseWriter, req
 	}
 	if oldHash == hash {
 		// The hash matches the hash of the latest revision, so
-		// no need to upload anything.
+		// no need to upload anything. When uploading a published URL and
+		// the latest revision is a development entity, then we need to
+		// actually publish the existing entity. Note that at this point the
+		// user is already known to have required permissions.
```
rogpeppe: s/have/have the/ ?

internal/v4/archive.go
```diff
@@ -170,7 +176,17 @@ func (h *ReqHandler) servePostArchive(id *charm.URL, w http.ResponseWriter, req
...
+		underDevelopment := id.Channel == charm.DevelopmentChannel
+		if oldUrl.Development && !underDevelopment {
```
rogpeppe: we should probably spell oldUrl conventionally as oldURL.

rogpeppe:
LGTM modulo a few minor issues and with successful QA - this is a significant change. Thanks!

internal/v4/archive.go
frankban: Ok changed.

docs/API.md
frankban: Good catch, and sorry that this typo made QA difficult initially.

docs/API.md
frankban: Done.

docs/API.md
frankban: Done.

docs/API.md
frankban: Really good catch, fixed the documentation.

internal/v4/api.go
frankban: Done.

internal/v4/archive.go

internal/v4/archive.go

internal/v4/archive.go

internal/v4/api.go
frankban: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1304/
Test PASSed.

urosj:
QA :ok:

frankban:
Thanks for the reviews and intensive QA!
:shipit:

jujugui:

PR #485, opened 2015-11-26
List end point.

fabricematrat:


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1328/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1329/
Test PASSed.

docs/API.md
```diff
@@ -1808,6 +1808,89 @@ The Meta field is populated according to the include flag  - see the `meta`
 path for more info on how to use this.
 The `limit` flag is the same as for the "search" path.
 
+### List
+
+#### GET list
+
+The `list` path lists within the latest version of charms and bundles
```
mhilton: why just the latest version? I think it should list all of them (at least with an option)

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,36 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
 	return result, nil
 }
 
+// List searches the store for the given ListParams.
```
mhilton: avoid using the word searches here, really it's more filtering.

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,36 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+// It returns a SearchResult containing the results of the search.
```
mhilton: Similarly we should return a ListResult

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,36 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+	} else {
+		query = query.Sort(sp.sort...)
+	}
+	var entities []*mongodoc.Entity
+	err := query.All(&entities)
+	if err != nil {
+		return SearchResult{}, errgo.Mask(err)
+	}
+	r := SearchResult{
+		SearchTime: 0 * time.Millisecond,
```
mhilton: hmm if you can't have a good value, don't have it at all.

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,36 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+		query = query.Sort("_id")
+	} else {
+		query = query.Sort(sp.sort...)
+	}
+	var entities []*mongodoc.Entity
+	err := query.All(&entities)
+	if err != nil {
+		return SearchResult{}, errgo.Mask(err)
+	}
+	r := SearchResult{
```
mhilton: I wouldn't be so keen to fit the results into the search result structure. Find something that makes sense for the data you have.

internal/v4/search.go
```diff
@@ -38,9 +38,20 @@ func (h *ReqHandler) serveSearch(_ http.Header, req *http.Request) (interface{},
 		}
 		sp.Groups = append(sp.Groups, groups...)
 	}
+
 	return h.doSearch(sp, req)
 }
 
+// GET list[?filter=value…][&include=meta][&sort=field[+dir]]
+// https://github.com/juju/charmstore/blob/v4/docs/API.md#get-list
+func (h *ReqHandler) serveList(_ http.Header, req *http.Request) (interface{}, error) {
```
mhilton: This should be in its own file list.go or similar.

internal/v4/search.go
```diff
@@ -165,3 +200,64 @@ func parseSearchParams(req *http.Request) (charmstore.SearchParams, error) {
 	}
 	return sp, nil
 }
+
+// parseListParms extracts the list parameters from the request
+func parseListParams(req *http.Request) (charmstore.ListParams, error) {
+	sp := charmstore.ListParams{}
```
mhilton: s/sp/lp/ ?

mhilton:
A good start, but I think there is still a little work to do. There is little point in trying to make it look like search, the two operations have quite different semantics.

docs/API.md
fabricematrat: done.

internal/charmstore/store.go
fabricematrat: done.

internal/charmstore/store.go
fabricematrat: done.

internal/charmstore/store.go
fabricematrat: done.

internal/charmstore/store.go
fabricematrat: done.

internal/v4/search.go
fabricematrat: done.

internal/v4/search.go
fabricematrat: done.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1330/
Test FAILed.

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,35 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+// List lists the store for the given ListParams.
+// It returns a ListResult containing the results of the list.
+func (store *Store) List(sp ListParams) (ListResult, error) {
+	query := store.DB.Entities().Find(sp.Filters)
+	if sp.sort == nil {
+		query = query.Sort("_id")
+	} else {
+		query = query.Sort(sp.sort...)
+	}
+	var entities []*mongodoc.Entity
```
mhilton: If someone lists everything this is going to get big. perhaps use an Iter instead?

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,35 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
 	return result, nil
 }
 
+// List lists the store for the given ListParams.
+// It returns a ListResult containing the results of the list.
+func (store *Store) List(sp ListParams) (ListResult, error) {
+	query := store.DB.Entities().Find(sp.Filters)
+	if sp.sort == nil {
+		query = query.Sort("_id")
```
mhilton: by default I think that promulgated ones should come first. not sure if you can do that easily though.

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,35 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+	} else {
+		query = query.Sort(sp.sort...)
+	}
+	var entities []*mongodoc.Entity
+	err := query.All(&entities)
```
mhilton: As you only need the URL and promulgated URL add  Select to just get those.

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,35 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+	if err != nil {
+		return ListResult{}, errgo.Mask(err)
+	}
+	r := ListResult{
+		Total:      len(entities),
+		Results:    make([]*router.ResolvedURL, 0, len(entities)),
+	}
+	for _, entity := range entities {
+		id := &router.ResolvedURL{
+			URL: *entity.PreferredURL(true),
```
mhilton: In a ResolvedURL the URL field has to be the full name of the URL. not the preferred one. Also you'll need to check the development flag here too.

internal/v4/list.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// doList performs a listing specified by ListParams. If lp
+// specifies that additional metadata needs to be added to the results,
+// then it is added. It also filters the result based on ACLs.
+func (h *ReqHandler) doList(lp charmstore.ListParams, req *http.Request) (interface{}, error) {
+	// perform query
+	results, err := h.Store.List(lp)
+	if err != nil {
+		return nil, errgo.Notef(err, "error performing search")
```
mhilton: "error listing charms and bundles"?

internal/v4/search.go
```diff
@@ -164,4 +164,4 @@ func parseSearchParams(req *http.Request) (charmstore.SearchParams, error) {
 		}
 	}
 	return sp, nil
-}
+}
```
mhilton: newline please I think go fmt adds those.

mhilton:
:+1: with some fixes.

internal/charmstore/store.go
fabricematrat: Done

internal/charmstore/store.go
fabricematrat: It means using $project most likely so will leave that for later

internal/charmstore/store.go
fabricematrat: Done.

internal/charmstore/store.go
fabricematrat: Done.

internal/v4/list.go
fabricematrat: Done.

internal/v4/search.go
fabricematrat: Done.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1331/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1332/
Test FAILed.

internal/charmstore/search.go
```diff
@@ -483,6 +493,23 @@ func (sp *SearchParams) ParseSortFields(f ...string) error {
 	return nil
 }
 
+func (sp *ListParams) ParseSortFieldsList(f ...string) error {
+	for _, s := range f {
+		for _, s := range strings.Split(s, ",") {
+			sort := sortListFields[s]
```
rogpeppe: This won't work when s has a "-" prefix.

rogpeppe: You probably want to do something like this:

```
field := sortListFields[strings.TrimPrefix(s, "-")]
if field == "" {
     return errgo.Newf("unknown field %q", field)
}
if s[0] == '-' {
    field = "-" + field
}
sp.sort = append(sp.sort, field)
```

internal/charmstore/search.go
```diff
@@ -483,6 +493,23 @@ func (sp *SearchParams) ParseSortFields(f ...string) error {
 	return nil
 }
 
+func (sp *ListParams) ParseSortFieldsList(f ...string) error {
```
rogpeppe: Amusingly this a lot of work for an identity function :)

internal/v4/list.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, result := range results.Results {
+		if err = h.AuthorizeEntity(result, req); err == nil {
+			filteredACLResults = append(filteredACLResults, result)
+		}
+	}
+	results.Results = filteredACLResults
+	return h.addMetaData(results, lp.Include, req)
+}
+
+//addMetada adds the requested meta data with the include list.
```
rogpeppe: Most of this code is copy and pasted from ReqHandler.doSearch. Let's please use the same code for both as they're both doing almost exactly the same thing. (I think that will be easier when both are using EntityResult).

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1333/
Test FAILed.

internal/v4/list.go
```diff
@@ -0,0 +1,151 @@
...
+			response.Results[j] = result
+			j++
+		}
+	}
+	response.Results = response.Results[0:j]
+	return response, nil
+}
+
+// parseListParms extracts the list parameters from the request
+func parseListParams(req *http.Request) (charmstore.ListParams, error) {
```
rogpeppe: This function is almost identical to parseSearchParams. ISTM that there's so much in common between the sort params and the list params that we should probably be using the same code to parse them both and only after parsing check for valid fields and map to the underlying entity names. The specific thing that tips me over the balance here is that I think it would be wrong if the two endpoints had a search field that they held in common but that had different semantics in each. That is, to the end user, the interface should be identical where possible. By having two separate parsers, we're inviting the implementations to diverge.

How about having a single parseSearchParams function (or other name that isn't either "search" or "list" :]) that can parse everything acceptable to both search and list, but accepts a set of allowed parameters. The field mapping that currently happens in ParseSortFields could be done in createSort, so the result of parseSearchParams would be independent of whether it would be used for search or list.

mhilton: Maybe break the parsing up into the distinct parts and create a function for each of them. Search and List can then use the ones that make sense in their context. Include parameters will be identical between both. filters and sort have slightly different sets.

rogpeppe:
The semantics look fine, but I think we could avoid a bunch of the code duplication
here. The front-end form parsing should be doable with the same code in both cases, I think.

internal/charmstore/search.go
mhilton: It does work look at the next line.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1334/
Test FAILed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1335/
Test FAILed.

docs/API.md
```diff
@@ -1808,6 +1808,88 @@ The Meta field is populated according to the include flag  - see the `meta`
...
+Any number of filters may be specified, limiting the list to items with attributes that
+match the specified filter value. Items matching any of the selected values for
+a filter are selected, so `name=1&name=2` would match items whose name was
+either 1 or 2. However, if multiple filters are specified, the charm must match
+all of them, so `name=1&series=2` will only match charms whose name is 1 and
+whose series is 2. Available filters are:
+
+* name - the charm's name.
+* owner - the charm's owner (the ~user element of the charm id)
+* promulgated - the charm has been promulgated.
```
rogpeppe: Is this actually allowed currently?

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,102 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+		return ListResult{}, errgo.Mask(err)
+	}
+	query := store.DB.Entities().Find(filters)
+	if len(sort) == 0 {
+		query = query.Sort("_id")
+	} else {
+		query = query.Sort(sort...)
+	}
+
+	//Only select needed field
```
rogpeppe: // Only select the fields that are needed.

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,102 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+	query = query.Select(bson.D{{"_id", 1}, {"url", 1}, {"development", 1}, {"promulgated-url", 1}})
+
+	r := ListResult{
+		Results: make([]*router.ResolvedURL, 0),
```
rogpeppe, line 1524: the ", 0" is redundant here.

internal/v4/search.go
```diff
@@ -80,19 +85,19 @@ func (h *ReqHandler) doSearch(sp charmstore.SearchParams, req *http.Request) (in
...
+	for _, result := range entities {
 		if result.Id != nil {
-			response.Results[j] = result
+			entities[j] = result
 			j++
 		}
 	}
-	response.Results = response.Results[0:j]
-	return response, nil
+	entities = entities[0:j]
```
rogpeppe: return entities[0:j]

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,102 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
 
+var listFilters = map[string]string{
+	"name":        "name",
+	"owner":       "user",
+	"series":      "serties",
+	"type":        "type",
+	"promulgated": "promulgated-revision",
+}
+
+func prepareList(sp SearchParams) (filters map[string]interface{}, sort []string, err error) {
```
rogpeppe, line 1436: This also needs to check that some of the search-only parameters aren't set.
For example, if any of the Text, AutoComplete, Limit or Skip fields are set
in sp, we should return an error.

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,102 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+
+	//Only select needed field
+	query = query.Select(bson.D{{"_id", 1}, {"url", 1}, {"development", 1}, {"promulgated-url", 1}})
+
+	r := ListResult{
+		Results: make([]*router.ResolvedURL, 0),
+	}
+	var entity mongodoc.Entity
+	iter := query.Iter()
+	for iter.Next(&entity) {
```
rogpeppe, line 1528: ISTM that we're ignoring the Admin, Username and Groups fields in SearchParams here, so we'll be able to list all charms regardless of ACL. We should fix that.

internal/v4/list.go
```diff
@@ -0,0 +1,36 @@
...
+	"gopkg.in/errgo.v1"
+	"gopkg.in/juju/charmrepo.v2-unstable/csclient/params"
+
+	"gopkg.in/juju/charmstore.v5-unstable/internal/router"
+)
+
+// GET list[?filter=value…][&include=meta][&sort=field[+dir]]
+// https://github.com/juju/charmstore/blob/v4/docs/API.md#get-list
+func (h *ReqHandler) serveList(_ http.Header, req *http.Request) (interface{}, error) {
+	sp, err := parseSearchParams(req)
```
rogpeppe, line 18: We also need to add username and groups to the search params. Perhaps factor out that stuff from ReqHander.serveSearch.

rogpeppe:
This is loads nicer now, thanks. I have a few comments and suggestions above. The ACL changes in particular are important. Perhaps start by writing some tests which will fail with the current logic, then implement the checking.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1336/
Test FAILed.

internal/charmstore/store.go
```diff
@@ -1425,6 +1425,115 @@ func (store *Store) Search(sp SearchParams) (SearchResult, error) {
...
+			filters["series"] = v[0]
+		case "type":
+			if v[0] == "bundle" {
+				filters["series"] = "bundle"
+			} else {
+				filters["series"] = map[string]interface{}{"$ne": "bundle"}
+			}
+		case "promulgated":
+			if v[0] != "0" {
+				filters["promulgated-revision"] = map[string]interface{}{"$gt": 0}
```
mhilton: shouldn't this be $ge ? if promulgated-revision = 0 then it's still promulgated.

internal/v4/list.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	// perform query
+	results, err := h.Store.List(sp)
+	if err != nil {
+		return nil, errgo.Notef(err, "error listing charms and bundles")
+	}
+
+	// TODO 30th Nov 2015 Fabrice:
+	// we should follow the same pattern as search, and put the user, admin and groups
+    // into the SearchParams and leave the charmstore package to be responsible for filtering
```
mhilton: there's something wrong with the indent here.

internal/charmstore/store.go
mhilton, line 1528: That is being dealt with in the v1 layer.

docs/API.md
```diff
@@ -1808,6 +1808,88 @@ The Meta field is populated according to the include flag  - see the `meta`
...
+the sort field is not specified the order will be a server side logical order. 
+It is possible to specify more than one sort field to get
+multi-level sorting, e.g. sort=name,-series will get charms in order of the
+charm name and then in reverse order of series.
+
+The Meta field is populated according to the include flag  - see the `meta`
+path for more info on how to use this.
+
+```go
+[]ListResult
```
mhilton: I think this should now be EntityResult

mhilton:
:+1:

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1338/
Test FAILed.

internal/charmstore/store.go
fabricematrat: Done.

internal/v4/list.go
fabricematrat: Done.

docs/API.md
fabricematrat: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1340/
Test PASSed.

fabricematrat:
:shipit:

jujugui:

PR #490, opened 2015-12-09
internal/v4: implement custom ResolveURL

rogpeppe:
This ensures that (most) ids returned by the v4 API
will have a series, even for multiple-series charms.

Also pave the way for v4 to be able to override specific
methods in the v5 router (for example, we want search-like
methods to be expand multi-series charms into multiple
result entities).

This required a bunch of shuffling around in internal/v5;
we took this opportunity to factor the various function arguments
to router.New into an interface, and removed the now-redundant
EntityExists function.

### DIFF
--- a/internal/v4/api.go
+++ b/internal/v4/api.go
@@ -8,8 +8,10 @@ import (
 
 	"gopkg.in/errgo.v1"
 	"gopkg.in/juju/charm.v6-unstable"
+	"gopkg.in/juju/charmrepo.v2-unstable/csclient/params"
 
 	"gopkg.in/juju/charmstore.v5-unstable/internal/charmstore"
+	"gopkg.in/juju/charmstore.v5-unstable/internal/mempool"
 	"gopkg.in/juju/charmstore.v5-unstable/internal/router"
 	"gopkg.in/juju/charmstore.v5-unstable/internal/v5"
 )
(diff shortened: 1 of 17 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1363/
Test PASSed.

internal/v4/api.go
```diff
@@ -31,25 +42,105 @@ type ReqHandler struct {
...
+	}
+	defer rh.Close()
+	rh.ServeHTTP(w, req)
 }
 
 func NewAPIHandler(pool *charmstore.Pool, config charmstore.ServerParams) charmstore.HTTPCloseHandler {
 	return New(pool, config)
 }
 
+// NewReqHandler fetchs a new instance of ReqHandler
```
mhilton, line 70: It doesn't really. It gets a store from h.Pool and then plugs it into a router from h.reqHandlerPool

mhilton:
:+1:

internal/v4/api.go
```diff
@@ -31,25 +42,105 @@ type ReqHandler struct {
...
+// ResolveURL implements router.Context.ResolveURL,
+// ensuring that any resulting ResolvedURL always
+// has a non-empty PreferredSeries field.
+func (h ReqHandler) ResolveURL(url *charm.URL) (*router.ResolvedURL, error) {
+	return resolveURL(h.Store, url)
+}
+
+// resolveURL implements URL resolving for the ReqHandler.
+// It's defined as a separate function so it can be more
+// easily unit-tested.
```
urosj, line 107: Should there be a comment that this is required by clients depending to have series resolved in the URL?

urosj:
:+1:  QA :ok: 

One note, but take it or leave it.

internal/v4/api.go
rogpeppe, line 107: I don't think it's needed, as ReqHandler.ResolveURL has a similar comment and resolveURL is not exported (and used only by ReqHandler.ResolveURL)

rogpeppe:
:shipit:

jujugui:

PR #504, opened 2016-01-05
v4, v5: add cache (step 1)

rogpeppe:
This is a work-in-progress.
We've added the cache (and made the required interface changes)
but we're not properly using it yet, except for ResolveURL(s).

One somewhat invasive change is to pass the map[string]int
to FindEntity and friends rather than the slice of string.
This is to make the interface with the entity cache somewhat
smoother. One alternative might be to represent
all field selections as a bitmask which would be considerably
more efficient, but also more limiting (only 64 fields max,
of which we're currently using ~32).

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -237,10 +237,23 @@ type Context interface {
 	// cause.
 	ResolveURL(id *charm.URL) (*ResolvedURL, error)
 
+	// ResolveURLs is like ResolveURL but resolves multiple URLs
+	// at the same time. The length of the returned slice should
+	// be len(ids); any entities that are not found should be represented
+	// by nil elements.
+	ResolveURLs(ids []*charm.URL) ([]*ResolvedURL, error)
+
 	// The AuthorizeEntity function will be called to authorize requests
 	// to any BulkIncludeHandlers. All other handlers are expected
 	// to handle their own authorization.
 	AuthorizeEntity(id *ResolvedURL, req *http.Request) error
+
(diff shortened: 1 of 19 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1421/
Test PASSed.

internal/router/router.go
```diff
@@ -434,6 +434,27 @@ func (r *Router) serveMetaGet(id *ResolvedURL, req *http.Request) (interface{},
 	return nil, errgo.WithCausef(nil, params.ErrNotFound, "unknown metadata %q", strings.TrimPrefix(req.URL.Path, "/"))
 }
 
+// GET id/meta/any?[include=meta[&include=meta...]]
+// https://github.com/juju/charmstore/blob/v4/docs/API.md#get-idmetaany
+func (r *Router) serveMetaGetAny(id *ResolvedURL, req *http.Request) (interface{}, error) {
+	includes := req.Form["include"]
+	// If there are no includes, we have no handlers to generate
```
mhilton: This comment seems out of place now. As we have resolved the URL that will have worked out that the entity exists.

mhilton:
:+1: Nice work

jrwren:
:+1:

rogpeppe:
:shipit:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1422/
Test PASSed.

PR #507, opened 2016-01-07
internal/entitycache: better tests

rogpeppe:
We make the entitycache test mocks more faithful to the actual
mgo primitives by making sure we always select only the fields
that have actually been selected.

We also add a test that promulgated URLs do not issue
a concurrent base-entity fetch (because they can't)
and ensure that the fields needed for the entitycache
logic are always fetched.

### DIFF
--- a/internal/entitycache/cache.go
+++ b/internal/entitycache/cache.go
@@ -61,12 +61,23 @@ type Cache struct {
 	baseEntities stash
 }
 
+var requiredEntityFields = map[string]int{
+	"_id":             1,
+	"promulgated-url": 1,
+	"development":     1,
+	"baseurl":         1,
+}
+
+var requiredBaseEntityFields = map[string]int{
+	"_id": 1,
+}
+
 // New returns a new cache that uses the given store
 // for fetching entities.
 func New(store Store) *Cache {
 	var c Cache
-	c.entities.init(c.getEntity, &c.wg)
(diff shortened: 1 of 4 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1438/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1439/
Test PASSed.

internal/entitycache/cache.go
```diff
@@ -61,12 +61,23 @@ type Cache struct {
...
 
+var requiredEntityFields = map[string]int{
+	"_id":             1,
+	"promulgated-url": 1,
+	"development":     1,
+	"baseurl":         1,
+}
+
+var requiredBaseEntityFields = map[string]int{
+	"_id": 1,
```
mhilton, line 72: I believe you would always get _id anyway unless  you explicitely set it to 0

mhilton:
:+1:

internal/entitycache/cache.go
rogpeppe, line 72: You would, but from the point of view of this package, I think it's better to be explicit.

fabricematrat:
:+1:

rogpeppe:
:shipit:

jujugui:

PR #514, opened 2016-01-18
internal/v5: fix panic on public write perms

rogpeppe:
When write perms on an entity allowed "everyone", the auth
check did not bother checking any authorization, as none
was required. Unfortunately this meant that the code would
panic later when it tried to add an audit log entry but found
no username. We fix the issue by always requiring authorization
for non-read operations.

### DIFF
--- a/internal/v5/api_test.go
+++ b/internal/v5/api_test.go
@@ -654,6 +654,29 @@ func (s *APISuite) TestMetaPermAudit(c *gc.C) {
 	}})
 }
 
+func (s *APISuite) TestMetaPermPublicWrite(c *gc.C) {
+	url := newResolvedURL("~bob/precise/wordpress-23", 23)
+	s.addPublicCharm(c, "wordpress", url)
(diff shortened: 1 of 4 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1460/
Test PASSed.

fabricematrat:
:+1:  
On side note, giving everyone permission to write is probably not the right thing also. We probably should not allow it.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1461/
Test PASSed.

rogpeppe:
> On side note, giving everyone permission to write is probably not the right thing also. We probably should not allow it

I'm not sure that it's worth being prescriptive about permissions. There are lots of ways of giving too many permissions to a charm (for example by adding a group that has a very large number of members), and it is really the responsibility of the charm maintainer to use the permissions in an appropriate way.

mhilton:
:+1:

rogpeppe:
:shipit:

jujugui:

PR #523, opened 2016-02-01
wip: new charm store model

frankban:
This branch proposes an implementation a the slightly changed
uploading and publishing model for the charm store, as
discussed in Cape Town.

In the new model we separate entities in two groups:
- unpublished entities;
- entities with at least one channel assigned.
  For the time being we support two channels: stable and development.
  When you first upload an entity, it is unpublished, and it can be
  referenced only by a restricted set of users (defined in BaseEntity.ACLs)
  using the specific revision. The publish endpoint can then be used to
  assign one or more channels to the entity, and contextually set the 
  corresponding entity as the current one for that channel, meaning
  it's the one being referenced when the revision is not specified.
  The whole model is better described in the model change google doc.

It's a long diff and I apologize, but lots of lines are related
to v5/v4 test changes, some are code removal, etc.
And yet there are many things that are missing and that are 
required here:, notably:
- modify expand-id (both v5 nd v4);
- fix expand-id tests;
- add tests for the included migrations;
- move publish request to charm package;
- extend publish and meta/perm responses;
- update documentation to reflect new endpoint semantics;
- add meta/channels API endpoint (doable in another branch).

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -37,6 +37,12 @@ var (
 	ErrTooManySessions = errgo.New("too many mongo sessions in use")
 )
 
+const (
+	// StableChannel is the channel used for stable charms or bundles.
+	// TODO (frankban) move this to the charm package.
+	StableChannel charm.Channel = ""
+)
+
 // Pool holds a connection to the underlying charm and blob
 // data stores. Calling its Store method returns a new Store
 // from the pool that can be used to process short-lived requests
@@ -537,7 +543,7 @@ func (s *Store) AddCharm(c charm.Charm, p AddParams) (err error) {
 	if id.Series == "bundle" || id.User == "" || id.Revision == -1 {
 		return errgo.Newf("charm added with invalid id %v", &id)
 	}
-	logger.Infof("add charm url %s; prev %d; dev %v", &id, p.URL.PromulgatedRevision, p.URL.Development)
+	logger.Infof("add charm url %s; promulgated rev %d", &id, p.URL.PromulgatedRevision)
 	entity := &mongodoc.Entity{
 		URL:                     &id,
 		PromulgatedURL:          p.URL.DocPromulgatedURL(),
@@ -553,7 +559,6 @@ func (s *Store) AddCharm(c charm.Charm, p AddParams) (err error) {
 		CharmRequiredInterfaces: interfacesForRelations(c.Meta().Requires),
 		Contents:                p.Contents,
 		SupportedSeries:         c.Meta().Series,
-		Development:             p.URL.Development,
 	}
 	denormalizeEntity(entity)
 
@@ -622,6 +627,7 @@ func (s *Store) insertEntity(entity *mongodoc.Entity) (err error) {
 		Public:          false,
 		ACLs:            acls,
 		DevelopmentACLs: acls,
+		StableACLs:      acls,
 		Promulgated:     entity.PromulgatedURL != nil,
 	}
 	err = s.DB.BaseEntities().Insert(baseEntity)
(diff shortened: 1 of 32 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1489/
Test FAILed.

internal/charmstore/store.go
```diff
@@ -711,35 +706,68 @@ func (s *Store) FindBestEntity(url *charm.URL, fields map[string]int) (*mongodoc
...
+			// A not found error is not contemplated here, as it would mean we
+			// have an internal inconsistency in which a base entity field
+			// refers to a non existing entity.
+			return nil, errgo.Notef(err, "cannot find charm or bundle for an internal error")
+		}
+		return entity, nil
+	}
+	if url.Revision == -1 {
+		// Find the current stable or development release.
+		baseEntity, err := s.FindBaseEntity(url, FieldSelector("developmentseries", "stableseries"))
```
rogpeppe, line 726: Getting the base entity here means we bypass the entitycache mechanism, which means the base entity won't be available in the cache and will need to be fetched again to check ACLs.

internal/v4/archive_test.go
```diff
@@ -406,22 +404,33 @@ func (s *ArchiveSuite) TestPostCharm(c *gc.C) {
...
+
+	// Retrieving the development version returns the last development charm.
+	err = s.store.SetPerms(charm.MustParseURL("~charmers/development/wordpress"), "read", params.Everyone)
+	c.Assert(err, gc.IsNil)
+	rec = httptesting.DoRequest(c, httptesting.DoRequestParams{
+		Handler: s.srv,
+		URL:     storeURL("~charmers/development/wordpress/archive"),
+	})
+	c.Assert(rec.Code, gc.Equals, http.StatusOK)
+	c.Assert(rec.Header().Get(params.EntityIdHeader), gc.Equals, "cs:~charmers/development/precise/wordpress-2")
```
mhilton, line 416: We can't do this in v4 old versions of juju cannot necessarily parse a URL containing a channel. Any URLs we return in v4 must not include channels.

internal/v5/auth.go
```diff
@@ -293,20 +297,44 @@ func areAllowedEntities(entityIds []*router.ResolvedURL, allowedEntities string)
...
+	acls, err := h.getACLs(baseEntity, id)
+	if err != nil {
+		return errgo.Mask(err, errgo.Is(params.ErrNotFound))
 	}
 	return h.authorizeWithPerms(req, acls.Read, acls.Write, id)
 }
 
+// getACLs returns the ACLs to use for the given base entity and resolved id.
+// What ACLs apply here? If a channel has been explicitly specified in the
+// URL, then, assuming the entity has been already resolved and fond for
```
mhilton, line 316: s/fond/found/

internal/v5/auth.go
```diff
@@ -293,20 +297,44 @@ func areAllowedEntities(entityIds []*router.ResolvedURL, allowedEntities string)
...
+// that channel, channel specific ACLs apply. Otherwise, the URL has no
+// channels, and so stable ACLs apply if the entity has been published as
+// stable, then development ACLs if the entity is in development, otherwise
+// the system falls back to unpublished ACLs.
+func (h *ReqHandler) getACLs(baseEntity *mongodoc.BaseEntity, id *router.ResolvedURL) (mongodoc.ACL, error) {
```
mhilton, line 321: As written this has an interesting property that people can lose access when a charm is published. Say for example ACLs.Read contains {"tom", "dick"} and DevelopmentACLs.Read contains {"tom", "harry"} then if dick tries to install from the charm with no channel, then before it is published at all he will be able to deploy it, after it is publised to development he will not. Perhaps if the channel isn't specified the ACL should be the union of all ACLs for channels to which the charm is published.

PR #540, opened 2016-02-18
internal/charmstore: implement store.Publish

mhilton:
Implement store.Publish. The only change prom the prototype here is
naming the stable channel "stable", this name is designed to be used
internally to distinguish it from charms that aren't published in any
channel.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -29,6 +29,17 @@ import (
 
 var logger = loggo.GetLogger("charmstore.internal.charmstore")
 
+// Channel is the name of a channel in which a charm may be published.
+type Channel string
+
+const (
+	// DevelopmentChannel is the channel used for charms or bundles under development.
+	DevelopmentChannel Channel = "development"
+
+	// StableChannel is the channel used for stable charms or bundles.
+	StableChannel Channel = "stable"
+)
+
 var (
 	errClosed          = errgo.New("charm store has been closed")
 	ErrTooManySessions = errgo.New("too many mongo sessions in use")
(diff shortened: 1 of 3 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1544/
Test PASSed.

internal/charmstore/store.go
```diff
@@ -29,6 +29,11 @@ import (
 
 var logger = loggo.GetLogger("charmstore.internal.charmstore")
 
+const (
+	// StableChannel is the channel used internally for stable charms or bundles.
+	StableChannel charm.Channel = "stable"
```
frankban: Should we define this in charm? Or should we remove the channel concept from charm and then go ahead and define the type with both stable and development channels here? We are no longer using channels in URLs, so I'd propose to just consider channels a charm store concept. So perhaps define channels here and eventually remove from charm?

internal/charmstore/store.go
```diff
@@ -633,6 +638,59 @@ func (s *Store) UpdateBaseEntity(url *router.ResolvedURL, update interface{}) er
 	return nil
 }
 
+// Publish assigns or removes channels to the entity corresponding to the given
```
frankban: It only assigns channels actually. Leftover from the prototype I guess.

internal/charmstore/store.go
```diff
@@ -633,6 +638,59 @@ func (s *Store) UpdateBaseEntity(url *router.ResolvedURL, update interface{}) er
...
+		for _, s := range series {
+			update = append(update, bson.DocElem{string(c) + "series." + s, entity.URL})
+		}
+	}
+	if err := s.UpdateBaseEntity(url, bson.D{{"$set", update}}); err != nil {
+		return errgo.Mask(err)
+	}
+
+	// Add entity to ElasticSearch.
+	if err := s.UpdateSearch(url); err != nil {
```
frankban, line 694: And this function will be update in a follow up, correct?

frankban:
:+1: with a couple of suggestions.
Thank you Martin!

internal/charmstore/store.go
mhilton, line 694: yes

internal/charmstore/store.go
mhilton: I've changed the comment

internal/charmstore/store.go
mhilton: I've chosen to define them here for charmstore, that way we can remove it from charm later.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1545/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1546/
Test PASSed.

mhilton:
:shipit:

jujugui:

PR #541, opened 2016-02-18
authenticate for read access using channel ACLs

mhilton:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1548/
Test PASSed.

internal/charmstore/store.go
```diff
@@ -844,13 +845,22 @@ func (s *Store) SetPromulgated(url *router.ResolvedURL, promulgate bool) error {
...
 }
 
-// SetPerms sets the permissions for the base entity with
-// the given id for "which" operations ("read" or "write")
-// to the given ACL. This is mostly provided for testing.
+// SetPerms sets the ACL specified by which for the base entity with the
+// given id. The which parameter is in the form "[channel].operation",
+// where channel, if specified, is one of "development" or "stable" and
+// operation is one of "read" or "write". If which does not specify a
+// channel then the unpublished ACL is updated. This is mostly provided
```
frankban: s/This is mostly/This is only/ so that the scope of this method is more clear?

internal/charmstore/store.go
```diff
@@ -844,13 +845,22 @@ func (s *Store) SetPromulgated(url *router.ResolvedURL, promulgate bool) error {
...
+// for testing.
 func (s *Store) SetPerms(id *charm.URL, which string, acl ...string) error {
 	field := "acls"
+	spec := strings.Split(which, ".")
```
frankban: Use SplitN? [shrug]

internal/v5/auth.go
```diff
@@ -191,21 +192,19 @@ func (h *ReqHandler) newDischargeRequiredError(m *macaroon.Macaroon, verr error,
 // an entry for each id with the corresponding ACL for each entity,
 // and requiredTerms holds entries for all required terms.
 func (h *ReqHandler) entityAuthInfo(entityIds []*router.ResolvedURL) (public bool, acls [][]string, requiredTerms map[string]bool, err error) {
-	// TODO use cache for entity lookup
 	acls = make([][]string, len(entityIds))
 	requiredTerms = make(map[string]bool)
 	public = true
 	for i, entityId := range entityIds {
-		entity, err := h.Store.FindEntity(entityId, nil)
+		entity, err := h.Cache.Entity(entityId.UserOwnedURL(), charmstore.FieldSelector("charmmeta"))
```
frankban, line 199: Thanks for using the cache here.

internal/v5/auth.go
```diff
@@ -296,15 +295,33 @@ func areAllowedEntities(entityIds []*router.ResolvedURL, allowedEntities string)
...
-		return errgo.Notef(err, "cannot retrieve entity %q for authorization", id)
+		return mongodoc.ACL{}, errgo.Notef(err, "cannot retrieve entity %q for authorization", id)
+	}
+	baseEntity, err := h.Cache.BaseEntity(id.UserOwnedURL(), charmstore.FieldSelector("acls", "developmentacls", "stableacls"))
+	if err != nil {
+		return mongodoc.ACL{}, errgo.Notef(err, "cannot retrieve base entity %q for authorization", id)
 	}
 	acls := baseEntity.ACLs
-	return h.authorizeWithPerms(req, acls.Read, acls.Write, id)
+	if entity.Development {
```
frankban: So why we are adding to the unpublished permissions instead of just switching the ACLs entirely?
For example, in the promulgation case, we might want to remove write permissions on the stable ACLs to the original owner but still allow him to upload new unpublished revisions. Given that we should start with the same permissions on the three ACLs, and that therefore initially the user has permissions on all channels, is this really required? If so, please let's add a comment about why we are doing this?

internal/v5/auth_test.go
```diff
@@ -295,6 +302,45 @@ var readAuthorizationTests = []struct {
...
+	groups:              []string{"group1", "group2", "group3"},
+	readPerm:            []string{"picard", "sisko", "group42", "group47"},
+	developmentReadPerm: []string{"group12"},
+	stableReadPerm:      []string{"group2"},
+	channels:            []charmstore.Channel{charmstore.DevelopmentChannel},
+	expectStatus:        http.StatusUnauthorized,
+	expectBody: params.Error{
+		Code:    params.ErrUnauthorized,
+		Message: `unauthorized: access denied for user "kirk"`,
+	},
```
frankban, line 343: If we keep propagating unpublished ACLs to devel and stable, we should test those cases as well.
If not, we should test that having unpublished access does not grant development or stable authorization.

frankban:
:+1: with a couple of minors and a more important observation.
Thank you!

internal/charmstore/store.go
```diff
@@ -844,13 +845,22 @@ func (s *Store) SetPromulgated(url *router.ResolvedURL, promulgate bool) error {
...
+	op := spec[0]
```
urosj: isn't spec[0] channel name, not op?

internal/v5/auth.go
urosj: +1 on what @frankban is saying. Initial case, owner has access to all three. In promulgated case, owner has write access to only unpublished and devel, but not stable.

internal/v5/auth_test.go
urosj, line 343: We should not propagate unpublished ACLs to devel and even less stable - that is true only in the initial case, when base entity is created.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1549/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1550/
Test PASSed.

mhilton:
:shipit:

jujugui:

PR #544, opened 2016-02-22
internal/router: remove development channel remnants

rogpeppe:
DocPromulgatedURL and UserOwnedURL are no longer necessary.
The router no longer needs to recognise channels, but it rejects them because
the charm package can still parse them.

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -148,21 +148,14 @@ type ResolvedURL struct {
 // This function panics if urlStr cannot be parsed as a charm.URL
 // or if it is not fully specified, including user and revision.
 func MustNewResolvedURL(urlStr string, promulgatedRev int) *ResolvedURL {
-	url := charm.MustParseURL(urlStr)
+	url := mustParseURL(urlStr)
 	if url.User == "" || url.Revision == -1 {
 		panic(fmt.Errorf("incomplete url %v", urlStr))
 	}
 	return &ResolvedURL{
-		URL:                 *url.WithChannel(""),
+		URL:                 *url,
 		PromulgatedRevision: promulgatedRev,
 	}
-}
-
-// UserOwnedURL returns the non-promulgated URL for the given id.
-// The returned *charm.URL may be modified freely.
-func (id *ResolvedURL) UserOwnedURL() *charm.URL {
-	u := id.URL
-	return &u
 }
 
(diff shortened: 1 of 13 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1553/
Test PASSed.

frankban:
:+1: Thanks for removing the development concept from router URLs!

internal/router/router.go
```diff
@@ -896,9 +873,28 @@ func splitId(path string) (url *charm.URL, rest string, err error) {
...
 	urlStr := strings.TrimSuffix(path[0:i], "/")
-	url, err = charm.ParseURL(urlStr)
+	url, err = parseURL(urlStr)
 	if err != nil {
 		return nil, "", errgo.Mask(err)
 	}
 	return url, path[i:], nil
 }
+
+func mustParseURL(s string) *charm.URL {
```
mhilton, line 883: Surely we don't need this. It would be better to take channels out of charm, until we do that why not just ignore any channel in a URL?

internal/router/router.go
```diff
@@ -148,23 +148,16 @@ type ResolvedURL struct {
...
 	return &ResolvedURL{
-		URL:                 *url.WithChannel(""),
+		URL:                 *url,
 		PromulgatedRevision: promulgatedRev,
 	}
 }
 
-// UserOwnedURL returns the non-promulgated URL for the given id.
-// The returned *charm.URL may be modified freely.
-func (id *ResolvedURL) UserOwnedURL() *charm.URL {
```
mhilton, line 163: you've checked that nothing was relying on being able to freely modify the UserOwnedURL?

mhilton:
:+1: With the slight caveat that I think we should be changing the charm package's parsing rather than implement our own wrapper around it.

internal/router/router.go
rogpeppe, line 883: Yeah, we should take channels out of charm, but that's backwardly incompatible, so we'll leave it for another day if that's OK.

internal/router/router.go
rogpeppe, line 163: Yeah. There were a couple of cases actually (both within ResolvedURL itself).

rogpeppe:
:shipit:

jujugui:

PR #547, opened 2016-02-24
internal/mongodoc: use maps for channel-based fields

rogpeppe:
This provides a uniform form for channel-specific information in the mongodoc.

We also re-align some of the v4 tests with their v5 counterparts
and delete the Public field from BaseEntity, as it's no longer used
and it's not clear what it should mean now.

### DIFF
--- a/internal/charmstore/addentity.go
+++ b/internal/charmstore/addentity.go
@@ -513,11 +513,12 @@ func (s *Store) addEntity(entity *mongodoc.Entity) (err error) {
 		Write: perms,
 	}
 	baseEntity := &mongodoc.BaseEntity{
-		URL:         entity.BaseURL,
-		User:        entity.User,
-		Name:        entity.Name,
-		Public:      false,
-		ACLs:        acls,
+		URL:  entity.BaseURL,
+		User: entity.User,
+		Name: entity.Name,
+		ChannelACLs: map[mongodoc.Channel]mongodoc.ACL{
+			mongodoc.UnpublishedChannel: acls,
+		},
 		Promulgated: entity.PromulgatedURL != nil,
 	}
 	err = s.DB.BaseEntities().Insert(baseEntity)
@@ -599,7 +600,7 @@ func (s *Store) bundleCharms(ids []string) (map[string]charm.Charm, error) {
 			// be returned to the user along with other bundle errors.
 			continue
 		}
-		e, err := s.FindBestEntity(url, StableChannel, map[string]int{})
+		e, err := s.FindBestEntity(url, mongodoc.NoChannel, map[string]int{})
 		if err != nil {
 			if errgo.Cause(err) == params.ErrNotFound {
 				// Ignore this error too, for the same reasons
--- a/internal/charmstore/common_test.go
+++ b/internal/charmstore/common_test.go
@@ -8,6 +8,7 @@ import (
 	"gopkg.in/juju/charm.v6-unstable"
 	"gopkg.in/macaroon-bakery.v1/bakery"
 
+	"gopkg.in/juju/charmstore.v5-unstable/internal/mongodoc"
 	"gopkg.in/juju/charmstore.v5-unstable/internal/router"
 	"gopkg.in/juju/charmstore.v5-unstable/internal/storetesting"
 )
@@ -24,7 +25,7 @@ func (s *commonSuite) addRequiredCharms(c *gc.C, bundle charm.Bundle) {
 	defer store.Close()
 	for _, svc := range bundle.Data().Services {
 		u := charm.MustParseURL(svc.Charm)
-		if _, err := store.FindBestEntity(u, StableChannel, nil); err == nil {
+		if _, err := store.FindBestEntity(u, mongodoc.NoChannel, nil); err == nil {
 			continue
 		}
 		if u.Revision == -1 {
@@ -44,7 +45,7 @@ func (s *commonSuite) addRequiredCharms(c *gc.C, bundle charm.Bundle) {
 		}
 		err := store.AddCharmWithArchive(&rurl, ch)
 		c.Assert(err, gc.IsNil)
-		err = store.Publish(&rurl, StableChannel)
+		err = store.Publish(&rurl, mongodoc.StableChannel)
 		c.Assert(err, gc.IsNil)
 	}
 }
--- a/internal/charmstore/migrations.go
+++ b/internal/charmstore/migrations.go
@@ -175,12 +175,12 @@ func addDevelopmentACLs(db StoreDatabase) error {
 	baseEntities := db.BaseEntities()
 	var baseEntity mongodoc.BaseEntity
 	iter := baseEntities.Find(bson.D{{
-		"developmentacls", bson.D{{"$exists", false}},
-	}}).Select(bson.D{{"_id", 1}, {"acls", 1}}).Iter()
+		"channelacls.development", bson.D{{"$exists", false}},
+	}}).Select(bson.D{{"_id", 1}, {"channelacls", 1}}).Iter()
 	defer iter.Close()
 	for iter.Next(&baseEntity) {
 		if err := baseEntities.UpdateId(baseEntity.URL, bson.D{{
-			"$set", bson.D{{"developmentacls", baseEntity.ACLs}},
+			"$set", bson.D{{"channelacls.development", baseEntity.ChannelACLs[mongodoc.DevelopmentChannel]}},
 		}}); err != nil {
 			return errgo.Notef(err, "cannot add development ACLs to base entity id %s", baseEntity.URL)
 		}
--- a/internal/charmstore/migrations_test.go
+++ b/internal/charmstore/migrations_test.go
@@ -407,47 +407,56 @@ func (s *migrationsSuite) TestMigrateAddDevelopment(c *gc.C) {
 	}
 }
 
-func (s *migrationsSuite) TestMigrateAddDevelopmentACLs(c *gc.C) {
-	s.patchMigrations(c, getMigrations(migrationAddDevelopmentACLs))
-
-	// Populate the database with some base entities.
-	entities := []*mongodoc.BaseEntity{{
-		URL:  charm.MustParseURL("~charmers/django"),
-		Name: "django",
-		ACLs: mongodoc.ACL{
-			Read:  []string{"user", "group"},
-			Write: []string{"user"},
-		},
-	}, {
-		URL:  charm.MustParseURL("~who/rails"),
-		Name: "rails",
-		ACLs: mongodoc.ACL{
-			Read:  []string{"everyone"},
-			Write: []string{},
-		},
-	}, {
-		URL:  charm.MustParseURL("~who/mediawiki-scalable"),
-		Name: "mediawiki-scalable",
-		ACLs: mongodoc.ACL{
-			Read:  []string{"who"},
-			Write: []string{"dalek"},
-		},
-	}}
-	for _, e := range entities {
-		s.insertBaseEntity(c, e, migrationAddDevelopmentACLs)
-	}
-
-	// Start the server.
-	err := s.newServer(c)
-	c.Assert(err, gc.IsNil)
-
-	// Ensure base entities have been updated correctly.
-	s.checkCount(c, s.db.BaseEntities(), len(entities))
-	for _, e := range entities {
-		e.DevelopmentACLs = e.ACLs
-		s.checkBaseEntity(c, e, migrationAddDevelopmentACLs)
-	}
-}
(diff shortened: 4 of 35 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1564/
Test PASSed.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1565/
Test FAILed.

internal/charmstore/addentity.go
```diff
@@ -599,7 +600,7 @@ func (s *Store) bundleCharms(ids []string) (map[string]charm.Charm, error) {
 			// be returned to the user along with other bundle errors.
 			continue
 		}
-		e, err := s.FindBestEntity(url, StableChannel, map[string]int{})
+		e, err := s.FindBestEntity(url, mongodoc.StableChannel, map[string]int{})
```
mhilton: This should match the scheme we used in addRequiredCharms which is mongodoc.NoChannel

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1566/
Test PASSed.

internal/charmstore/common_test.go
```diff
@@ -20,11 +21,12 @@ type commonSuite struct {
 // addRequiredCharms adds any charms required by the given
 // bundle that are not already in the store.
 func (s *commonSuite) addRequiredCharms(c *gc.C, bundle charm.Bundle) {
+	c.Logf("addRequiredCharms")
```
mhilton: I'm not sure we still want this message since we worked out what was going wrong.

internal/charmstore/common_test.go
```diff
@@ -42,9 +44,10 @@ func (s *commonSuite) addRequiredCharms(c *gc.C, bundle charm.Bundle) {
 		} else {
 			rurl.PromulgatedRevision = -1
 		}
+		c.Logf("addRequiredCharms AddCharmWithArchive %v", &rurl)
```
mhilton: the same with this message

internal/charmstore/store.go
```diff
@@ -523,23 +503,23 @@ func (s *Store) FindBestEntity(url *charm.URL, channel Channel, fields map[strin
...
 	}
+
 	switch channel {
-	case StableChannel, NoChannel:
-		return s.findStableEntity(url, fields)
-	case DevelopmentChannel:
-		return s.findDevelopmentEntity(url, fields)
-	case UnpublishedChannel:
+	case mongodoc.UnpublishedChannel:
 		return s.findUnpublishedEntity(url, fields)
```
mhilton, line 517: not for this PR, but is it reasonable to use findEntityInChannel to find unpublished entities too?

mhilton:

internal/charmstore/addentity.go
```diff
@@ -513,11 +513,12 @@ func (s *Store) addEntity(entity *mongodoc.Entity) (err error) {
...
-		URL:         entity.BaseURL,
-		User:        entity.User,
-		Name:        entity.Name,
-		Public:      false,
-		ACLs:        acls,
+		URL:  entity.BaseURL,
+		User: entity.User,
+		Name: entity.Name,
+		ChannelACLs: map[mongodoc.Channel]mongodoc.ACL{
+			mongodoc.UnpublishedChannel: acls,
```
frankban, line 520: I wonder if we should just set the initial acls for all channels here, so that when you publish for the first time we already have a reasonable default (user both read and write).

internal/charmstore/common_test.go
```diff
@@ -20,11 +21,12 @@ type commonSuite struct {
 // addRequiredCharms adds any charms required by the given
 // bundle that are not already in the store.
 func (s *commonSuite) addRequiredCharms(c *gc.C, bundle charm.Bundle) {
+	c.Logf("addRequiredCharms")
 	store := s.newStore(c, true)
 	defer store.Close()
 	for _, svc := range bundle.Data().Services {
 		u := charm.MustParseURL(svc.Charm)
-		if _, err := store.FindBestEntity(u, StableChannel, nil); err == nil {
+		if _, err := store.FindBestEntity(u, mongodoc.NoChannel, nil); err == nil {
```
frankban, line 28: From Stable to NoChannel. I am sure this is ok, just FYI.

internal/charmstore/migrations.go
```diff
@@ -175,12 +175,12 @@ func addDevelopmentACLs(db StoreDatabase) error {
...
 	iter := baseEntities.Find(bson.D{{
-		"developmentacls", bson.D{{"$exists", false}},
-	}}).Select(bson.D{{"_id", 1}, {"acls", 1}}).Iter()
+		"channelacls.development", bson.D{{"$exists", false}},
+	}}).Select(bson.D{{"_id", 1}, {"channelacls", 1}}).Iter()
 	defer iter.Close()
 	for iter.Next(&baseEntity) {
 		if err := baseEntities.UpdateId(baseEntity.URL, bson.D{{
-			"$set", bson.D{{"developmentacls", baseEntity.ACLs}},
+			"$set", bson.D{{"channelacls.development", baseEntity.ChannelACLs[mongodoc.DevelopmentChannel]}},
```
frankban, line 183: I think this should populate "channelacls.development" with the value from `baseEntity.ChannelACLs[mongodoc.UnpublishedChannel]`.

internal/charmstore/migrations_test.go
```diff
@@ -407,47 +407,56 @@ func (s *migrationsSuite) TestMigrateAddDevelopment(c *gc.C) {
...
-	c.Assert(err, gc.IsNil)
-
-	// Ensure base entities have been updated correctly.
-	s.checkCount(c, s.db.BaseEntities(), len(entities))
-	for _, e := range entities {
-		e.DevelopmentACLs = e.ACLs
-		s.checkBaseEntity(c, e, migrationAddDevelopmentACLs)
-	}
-}
+// This test is commented out because mongodoc.BaseEntity no longer has a DevelopmentACLs
```
frankban, line 410: Shoudl we uncomment this test?

internal/charmstore/store.go
```diff
@@ -30,23 +30,6 @@ import (
 
 var logger = loggo.GetLogger("charmstore.internal.charmstore")
 
-// Channel is the name of a channel in which a charm may be published.
```
frankban, line 33: Moving these definitions to the mongodoc package seems a good idea, thanks.

internal/charmstore/store.go
```diff
@@ -375,9 +358,6 @@ func (s *Store) ensureIndexes() error {
 		s.DB.Entities(),
 		mgo.Index{Key: []string{"promulgated-url"}, Unique: true, Sparse: true},
 	}, {
-		s.DB.BaseEntities(),
```
frankban, line 378: Thank you!

internal/charmstore/store.go
```diff
@@ -985,15 +936,8 @@ func (s *Store) SetPromulgated(url *router.ResolvedURL, promulgate bool) error {
 // channel then the unpublished ACL is updated. This is only provided for
 // testing.
 func (s *Store) SetPerms(id *charm.URL, which string, acl ...string) error {
-	field := "acls"
```
frankban, line 988: Nice simplification.

internal/mongodoc/doc.go
```diff
@@ -330,6 +315,23 @@ func (b *IntBool) SetBSON(raw bson.Raw) error {
 	return nil
 }
 
+// Channel is the name of a channel in which a charm may be published.
```
frankban: s/a charm/a charm or bundle/

internal/v4/auth_test.go
```diff
@@ -435,6 +442,90 @@ var writeAuthorizationTests = []struct {
 		Code:    params.ErrUnauthorized,
 		Message: `unauthorized: access denied for user "kirk"`,
 	},
+}, {
```
frankban, line 445: Nice tests!

internal/v5/api.go
```diff
@@ -996,7 +994,9 @@ func checkExtraInfoKey(key string, field string) error {
 // GET id/meta/perm
 // https://github.com/juju/charmstore/blob/v4/docs/API.md#get-idmetaperm
 func (h *ReqHandler) metaPerm(entity *mongodoc.BaseEntity, id *router.ResolvedURL, path string, flags url.Values, req *http.Request) (interface{}, error) {
-	acls := entity.ACLs
+	// TODO: choose appropriate ACLs based on currently appropriate channel
```
frankban, line 997: And here we should use ?channel, correct?

internal/v5/api.go
```diff
@@ -1010,22 +1010,15 @@ func (h *ReqHandler) putMetaPerm(id *router.ResolvedURL, path string, val *json.
 	if err := json.Unmarshal(*val, &perms); err != nil {
 		return errgo.Mask(err)
 	}
-	isPublic := false
```
frankban, line 1013: Removing this public field is a small step for a man, one giant leap for mankind.

internal/v5/archive.go
```diff
@@ -311,9 +312,10 @@ func (h *ReqHandler) ServeBlobFile(w http.ResponseWriter, req *http.Request, id
 }
 
 func (h *ReqHandler) isPublic(id *router.ResolvedURL) bool {
-	baseEntity, err := h.Cache.BaseEntity(&id.URL, charmstore.FieldSelector("acls"))
+	baseEntity, err := h.Cache.BaseEntity(&id.URL, charmstore.FieldSelector("channelacls"))
 	if err == nil {
-		for _, p := range baseEntity.ACLs.Read {
+		// TODO choose appropriate ACLs based on entity.
```
frankban: Isn't it as simple as calling h.entityACLs at this point?
Mayeb we should just mention that in the comment...

frankban:
:+1: with some comments.
Thank you for this effort, and for fixing all the tests!

internal/charmstore/addentity.go
rogpeppe: done

internal/charmstore/common_test.go
rogpeppe: done

internal/charmstore/common_test.go
rogpeppe: done

internal/charmstore/store.go
rogpeppe, line 517: I'm not sure. I can't think of a way of updating ChannelEntities on initial publish in a non-racy way so that we guarantee that looking up an unversioned entity in the Unpublished channel will always return the latest unpublished entity.

internal/charmstore/addentity.go
rogpeppe, line 520: Yes, definitely. But in this PR we're keeping things more directly equivalent to the old semantics to avoid changing too much at the same time. That means that we're translating ACLs to its Unpublished equivalent but no more.

internal/charmstore/common_test.go
rogpeppe, line 28: Yeah, this was deliberate.

internal/charmstore/migrations.go
rogpeppe, line 183: I'm not sure what we should do about migrations in the face of field renames.
Technically each migration step should not need to change because of
a later step. Perhaps we'd be better off keeping all the old mongodoc types
around.

Alternatively, if we're sure that all the charmstore instances that we care about
have migrated their database with this step already, we could just delete
it.

rogpeppe, line 183: Also, yes, this migration is indeed wrong. Note that it won't work even if we use baseEntity.ChannelACLs[mongodoc.UnpublishedChannel] because that field won't exist when this migration step is being run. We need to think hard about migrations before merging this feature branch into master.

internal/charmstore/migrations_test.go
rogpeppe, line 410: Not until we do migrations.

internal/mongodoc/doc.go
rogpeppe: done (ish)

internal/v4/auth_test.go
rogpeppe, line 445: They're just copied from v5.

internal/v5/api.go
rogpeppe, line 997: yeah; added a TODO.

internal/v5/api.go
rogpeppe, line 1013: :)

internal/v5/archive.go
rogpeppe: changed TODO

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1567/
Test PASSed.

rogpeppe:
:shipit:

jujugui:

PR #549, opened 2016-02-26
internal/v5: resolve URLs based on channel parameter

rogpeppe:
Many of the tests were failing because they did not use the standard addPublicCharm
helper method, so we change them all to use addPublicCharm when possible.

### DIFF
--- a/internal/v4/api_test.go
+++ b/internal/v4/api_test.go
@@ -518,7 +518,7 @@ func (s *APISuite) TestEndpointGet(c *gc.C) {
 func (s *APISuite) TestAllMetaEndpointsTested(c *gc.C) {
 	// Make sure that we're testing all the metadata
 	// endpoints that we need to.
-	s.addPublicCharm(c, "wordpress", newResolvedURL("~charmers/precise/wordpress-23", 23))
+	s.addPublicCharmFromRepo(c, "wordpress", newResolvedURL("~charmers/precise/wordpress-23", 23))
 	rec := httptesting.DoRequest(c, httptesting.DoRequestParams{
 		Handler: s.srv,
 		URL:     storeURL("precise/wordpress-23/meta"),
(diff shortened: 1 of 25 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1572/
Test PASSed.

internal/v5/api.go
```diff
@@ -994,6 +1009,7 @@ func checkExtraInfoKey(key string, field string) error {
 // GET id/meta/perm
 // https://github.com/juju/charmstore/blob/v4/docs/API.md#get-idmetaperm
 func (h *ReqHandler) metaPerm(entity *mongodoc.BaseEntity, id *router.ResolvedURL, path string, flags url.Values, req *http.Request) (interface{}, error) {
```
mhilton, line 1011: Thinking about this update, does it now mean that if I attempt to update the development ACLs using meta perm I can only do it if there is already an entity published in the channel? Or what does PUT /v5/~charmers/wordpress/meta/perm?channel=stable do if there is no wordpress in a stable channel?

internal/v4/api_test.go
```diff
@@ -2455,6 +2450,102 @@ func (s *APISuite) TestHash256Laziness(c *gc.C) {
...
+	s.discharge = dischargeForUser("charmers")
+	for _, add := range urlChannelResolvingEntities {
+		err := s.store.AddCharmWithArchive(add.id, storetesting.NewCharm(nil))
+		c.Assert(err, gc.IsNil)
+		if add.channel != mongodoc.UnpublishedChannel {
+			err = s.store.Publish(add.id, add.channel)
+			c.Assert(err, gc.IsNil)
+		}
+	}
+	for i, test := range urlChannelResolvingTests {
```
frankban, line 2522: Should we test that this works the same on the id endpoints?

frankban:
:+1: with a comment.
Thanks for fixing all those tests!

mhilton:
:+1:

internal/v5/api.go
rogpeppe, line 1011: If there's no wordpress in a stable channel, then ~charmers/wordpress won't resolve.
But after PR about to be proposed, PUT /v5/~charmers/wordpress-0/meta/perm?channel=stable
will work (ignoring the channel attribute) and in the PR after that one, the meta/perm endpoint will inspect the channel parameter to decide which perms to use, so that PUT request will update the stable ACLs even if there is no wordpress in a stable channel.

internal/v4/api_test.go
rogpeppe, line 2522: I don't think so. We're using meta/any here only because it's the easiest way to tell us what the resolved id was. Given that the implementation only resolved URLs in one place, I don't think we'd be testing anything else by adding tests for other endpoints (there are lots of endpoints that return entity ids).

rogpeppe:
:shipit:

jujugui:

PR #550, opened 2016-02-26
internal/charmstore: no error when specifying a channel with explicit revno

rogpeppe:
It seems reasonable to be liberal when an explicit revision is provided, and
this will also allow us to use the channel attribute in meta/perm to
specify a channel different from that of the entity that was resolved.


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1574/
Test PASSed.

frankban:
:+1: ty!

mhilton:
:+1:

rogpeppe:
:shipit:

jujugui:

PR #551, opened 2016-02-26
make expand-id check the ACLs for all returned entities

mhilton:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1576/
Test PASSed.

internal/charmstore/addentity.go
```diff
@@ -437,7 +437,7 @@ func (s *Store) addCharm(c charm.Charm, p addParams) (err error) {
 	// the name of a bundle. This is racy, but it's the best we can
 	// do. Also check that there isn't an existing multi-series charm
 	// that would be replaced by this one.
-	entities, err := s.FindEntities(entity.BaseURL, nil)
+	entities, err := s.FindEntities(entity.BaseURL, mongodoc.UnpublishedChannel, nil)
```
rogpeppe: It looks like this is finding just unpublished entities. Isn't that wrong here? For example what if there is a duplicate name but all its entities are published?

internal/charmstore/addentity.go
```diff
@@ -487,7 +487,7 @@ func (s *Store) addBundle(b charm.Bundle, p addParams) error {
 
 	// Check that we're not going to create a bundle that duplicates
 	// the name of a charm. This is racy, but it's the best we can do.
-	entities, err := s.FindEntities(entity.BaseURL, nil)
+	entities, err := s.FindEntities(entity.BaseURL, mongodoc.UnpublishedChannel, nil)
```
rogpeppe: Ditto.

internal/charmstore/addentity.go
mhilton: Every entity, by definition is in the unpublished channel, perhaps we need to change the name of the unpublished channel to make the meaning more clear.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1578/
Test PASSed.

internal/charmstore/store.go
```diff
@@ -619,13 +616,11 @@ var seriesScore = map[string]int{
...
-// entities matching the given URL. If the given URL has no user then
-// the produced query will only match promulgated entities. If the given URL
-// channel is not "development" then the produced query will only match
-// published entities.
+// entities matching the given URL. If the given URL has no user then the
+// produced query will only match promulgated entities.
 func (s *Store) EntitiesQuery(url *charm.URL) *mgo.Query {
 	entities := s.DB.Entities()
-	query := make(bson.D, 1, 5)
+	query := make(bson.D, 1, 6)
```
rogpeppe: Is this still needed?

mhilton: no, good spot

internal/v4/api.go
```diff
@@ -302,9 +300,26 @@ func (h ReqHandler) serveExpandId(id *router.ResolvedURL, w http.ResponseWriter,
...
+			return errgo.Mask(err)
+		}
+	}
+
 	// Collect all the expanded identifiers for each entity.
 	response := make([]params.ExpandedId, 0, len(docs))
 	expandMultiSeries(docs, func(series string, doc *mongodoc.Entity) error {
+		eid := charmstore.EntityResolvedURL(doc)
+		// Always check the ACLs for the entity before adding to the response.
+		acl, err := h.EntityReadACLAnyChannel(eid)
```
rogpeppe: Shouldn't this just be using AuthorizeEntity like any other endpoint that lists entities?

internal/v4/api_test.go
```diff
@@ -1586,89 +1577,446 @@ func (s *APISuite) TestResolveURL(c *gc.C) {
 	}
 }
 
+var serveExpandIdCharms = []struct {
+	id    *router.ResolvedURL
+	charm charm.Charm
+}{{
+	id: newResolvedURL("cs:~charmers/precise/wordpress-0", 0),
+	// Note: this charms sets the ACL for all ~chamers/wordpress charms.
+	charm: testCharm{
```
rogpeppe: Isn't it a bit dubious to have a testCharm without any Charm field? I'm somewhat surprised this doesn't panic actually.

mhilton: I implemented the thing at the lowest level that if it's nil it uses a storetesting.NewCharm(nil) to make this all tickety-boo.

internal/v4/api.go
mhilton: As per our discussion this morning it should expand to all entities the user has access to through some ACL. this is not the same mechanism as supported by AuthorizeEntity, its also not clear that it should be.

frankban: Just export something like AuthorizeEntityAnyChannel?

internal/v4/common_test.go
```diff
@@ -194,17 +195,85 @@ func (s *commonSuite) addPublicCharmFromRepo(c *gc.C, charmName string, rurl *ro
...
+	}
+	return tc
+}
+
+func (s *commonSuite) addCharm(c *gc.C, ch charm.Charm, rurl *router.ResolvedURL) (*router.ResolvedURL, charm.Charm) {
+	if ch == nil {
+		ch = storetesting.NewCharm(nil)
+	}
+	if tch, ok := ch.(testCharm); ok {
+		rurl, ch = s.addCharm(c, tch.Charm, rurl)
```
rogpeppe: This recursive call seems a bit squiggly to me. It seems odd that we're passing in a non-nil charm.Charm where if you call any charm.Charm method it'll panic that somehow turns into a valid charm.Charm along the way...

Perhaps it might be nicer to do declare an interface that can optionally be implemented to add on-publish perms and channels; something like this
perhaps? This way, a charm is always either valid or not valid, not somewhere in between.

```
// PublishEntity is the type recognized by addCharm and addBundle
// as a way of associating ACL and channel information with a
// charm at publish time.
type PublishEntity interface {
    ACLs() map[mongodoc.Channel]mongodoc.ACL
    Channels() []mongodoc.Channel
}

// ... probably use this type in Store.AddCharmWithArchive and avoid
// the current type switch shenanigans (which are wrongly documented
// anyway)
type CharmWithArchive interface {
    charm.Charm
    charmstore.ArchiverTo
}


type charmWithPublishInfo struct {
    charmstore.CharmWithArchive
    PublishEntity
}

func charmWithACLs(ch CharmWithArchive, acls map[mongodoc.Channel]mongodoc.ACL) CharmWithArchive {
    return charmWithPublishInfo{
        PublishEntity: withACLs(ch, acls),
        CharmWithArchive: ch,
    }
}

func charmWithChannels(ch CharmWithArchive, channels ...mongodoc.Channel) CharmWithArchive {
    return charmWithPublishInfo{
        PublishEntity: withChannels(ch, acls),
        CharmWithArchive: ch,
    }
}

type bundleWithPublishInfo struct {
    charmstore.BundleWithArchive
    PublishEntity
}

func bundleWithACLs(b BundleWithArchive, acls map[mongodoc.Channel]mongodoc.ACL) CharmWithArchive {
    return bundleWithPublishInfo{
        PublishEntity: withACLs(ch, acls),
        BundleWithArchive: b,
    }
}

func bundleWithChannels(b BundleWithArchive, channels ...mongodoc.Channel) CharmWithArchive {


func withACLs(x interface{}, acls map[mongodoc.Channel]mongodoc.ACL) PublishEntity {
    info := publishEntityInfo{
        acls: acls,
    }
    if x, ok  := x.(PublishEntity); ok {
        info.channels = x.Channels()
    }
    return info
}

func withChannels(x interface{}, channels ...mongodoc.Channel) PublishEntity {
    info := publishEntityInfo{
        channels: channels,
    }
    if x, ok  := x.(PublishEntity); ok {
        info.acls = x.ACLs()
    }
    return info
}

type publishEntityInfo struct {
    acls     map[mongodoc.Channel]mongodoc.ACL
    channels []mongodoc.Channel
}

func (a publishEntityInfo) ACLs() map[mongodoc.Channel]mongodoc.ACL {
    return a.acls
}

func (a publishEntityInfo) Channels() []mongodoc.Channel {
    return a.channels
}
```

rogpeppe:
There's lots of good stuff here, but I'm not sure that changing expand-id is the right time to add it. All endpoints will need their permissions checking changed, so expand-id should just act like all the others until we change them, I think.

frankban:
I think that in our discussion we ended up with not considering expand-id special.
So let's keep the same logic in all endpoints. The original logic I suggested (and I think we agreed on this morning) is to check agains a single set of ACLs, being: 
- stable if the entity is stable AND ?channel != development
  OTHERWISE
- development if
  - the entity is development
  - the entity is both stable and development AND ?channel == development
    OTHERWISE
- unpublished.
  @mhilton sorry about that.
  Let's discuss tomorrow if there are any further concerns.

mhilton:
@frankban Really my takeaway from this morning was that expand-id would list every entity that you had access to through one of the ACLs. I'm now very confused.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1579/
Test PASSed.

rogpeppe:
LGTM, thanks and sorry for the churn.

frankban:
:+1: thank you!

mhilton:
:shipit:

jujugui:

PR #552, opened 2016-02-26
Filled in more information regarding resources.

kat-co:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1577/
Test PASSed.

docs/API.md
```diff
@@ -1702,10 +1703,29 @@ The above example is equivalent to the `meta/common-info` example above.
 
 **Not yet implemented**
 
-#### POST *id*/resources/name.stream
+#### GET *id*/meta/resources
```
ericsnowcurrently, line 1706: What about support for bulk calls (multiple ids)?

docs/API.md
```diff
@@ -1702,10 +1703,29 @@ The above example is equivalent to the `meta/common-info` example above.
...
+    Type        string
+    Path        string
+    Description string
+    Revision    int
+    Fingerprint string
+    Size        int64
+}
+```
+
+#### POST *id*/resources/name
```
ericsnowcurrently: Why is this separate from the PUT operation?  It implies that they aren't intrinsically linked (e.g. that later you can upload a different file for the same resource revision).  File information for a resource should not be available unless the resource file is also available (I.E. if you can list the resource then GET must work too).  Perhaps I've totally misunderstood?

docs/API.md
```diff
@@ -1713,23 +1733,15 @@ type ResourcesRevision struct {
...
 ```
 
-#### GET *id*/resources/name.stream[-revision]/arch/filename
+#### GET *id*/resources/name[/revision]
 
-Getting from the `/resources` path retrieves a charm resource from the charm
-with the given id. If version is not specified, it retrieves the latest version
-of the resource. The SHA-256 hash of the data is specified in the HTTP response
-headers.
+Getting from the `/resources` path retrieves a charm resource from the charm with the given id. If version is not specified, it retrieves the latest version of the resource. The SHA-256 hash of the data is specified in the HTTP response headers.
```
ericsnowcurrently: Is SHA-256 still what we are using?  I thought we were moving to SHA-384 for resources.

docs/API.md
```diff
@@ -1713,23 +1733,15 @@ type ResourcesRevision struct {
...
 
-#### PUT *id*/resources/[~user/]series/name.stream-revision/arch?sha256=hash
+#### PUT *id*/resources/[~user/]name/revision/?sha256=hash
```
ericsnowcurrently: Drop the [~user/] part.

docs/API.md
```diff
@@ -1713,23 +1733,15 @@ type ResourcesRevision struct {
...
-Putting to the `resources` path uploads a resource (an arbitrary "blob" of
-data) associated with the charm with id series/name, which must not be a
-bundle. Stream and arch specify which of the charms resource streams and which
-architecture the resource will be associated with, respectively. Revision
-specifies the revision of the stream that's being uploaded to.
+Putting to the `resources` path uploads a resource (an arbitrary "blob" of data) associated with the charm with id, which must not be a bundle. Revision specifies the revision of the stream that's being uploaded to.
 
-The hash value must specify the hash of the stream. If the same series, name,
-stream, revision combination is PUT again, it must specify the same hash.
+The hash value must specify the hash of the stream. If the same series, name, stream, revision combination is PUT again, it must specify the same hash.
```
ericsnowcurrently: This needs to drop references to streams and series.

docs/API.md
ericsnowcurrently: I expect this is a hold-over from when a revision might have more than one file, depending on series, arch, etc.  That no longer applies.

docs/API.md
ericsnowcurrently, line 1706: Would this be handled through the "meta" GET?

Regardless, there should be some mention of the "meta" GET here, no?

docs/API.md
kat-co: Yeah, I had the same thought, but figured this was the way the charmstore folks wanted the workflow structured. I'll propose a unified PUT and see what feedback they have.

docs/API.md
kat-co: Yes, oversight, sorry.

docs/API.md
kat-co: Oops, yes. Also dropped the revision portion since the server will decide what revision the resource is at.

docs/API.md
kat-co: Yep, sorry.

docs/API.md
kat-co, line 1706: You are correct, it looks like this would be handled by the meta/any endpoint. Other sections that support the meta/any endpoint don't seem to refer to it in their section, so I think it's intended that meta/any be common knowledge. Agree, disagree?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1581/
Test PASSed.

docs/API.md
```diff
@@ -1702,34 +1703,33 @@ The above example is equivalent to the `meta/common-info` example above.
...
 
-#### PUT *id*/resources/[~user/]series/name.stream-revision/arch?sha256=hash
+Getting from the `/resources` path retrieves a charm resource from the charm with the given id. If version is not specified, it retrieves the latest version of the resource. The SHA-384 hash of the data is specified in the HTTP response headers.
 
-Putting to the `resources` path uploads a resource (an arbitrary "blob" of
-data) associated with the charm with id series/name, which must not be a
-bundle. Stream and arch specify which of the charms resource streams and which
-architecture the resource will be associated with, respectively. Revision
-specifies the revision of the stream that's being uploaded to.
+#### PUT *id*/resources/*name*/?sha384=hash
```
ericsnowcurrently, line 1730: We use POST for this for charms (POST id/archive?hash=sha384hash).  So this should be POST too.

ericsnowcurrently:
We also need to add support for resources to the "PUT _id_/publish" endpoint.

docs/API.md
```diff
@@ -1702,34 +1703,33 @@ The above example is equivalent to the `meta/common-info` example above.
 
 **Not yet implemented**
 
-#### POST *id*/resources/name.stream
+#### GET *id*/meta/resources
 
-Posting to the resources path creates a new version of the given stream
-for the charm with the given id. The request returns the new version.
+Getting from this path will return metadata regarding the resources for the charm with id of `id`.
```
rogpeppe, line 1708: To fit with the standard phrasing used in other places:

```
The `meta/resources` path returns information on all the resources associated with the given charm *id*.
```

docs/API.md
```diff
@@ -1702,34 +1703,33 @@ The above example is equivalent to the `meta/common-info` example above.
...
 
 ```go
-type ResourcesRevision struct {
-        Revision int
+type ResourcesResponse struct {
+    Resources []Resource
```
rogpeppe, line 1712: I guess this could also be:

```
Resources map[string] Resource
```

to emphasise the fact that there's only one resource per name.
(or is that actually the case?)

docs/API.md
```diff
@@ -1702,34 +1703,33 @@ The above example is equivalent to the `meta/common-info` example above.
...
+```
 
-Getting from the `/resources` path retrieves a charm resource from the charm
-with the given id. If version is not specified, it retrieves the latest version
-of the resource. The SHA-256 hash of the data is specified in the HTTP response
-headers.
+#### GET *id*/resources/*name*[/revision]
 
-#### PUT *id*/resources/[~user/]series/name.stream-revision/arch?sha256=hash
+Getting from the `/resources` path retrieves a charm resource from the charm with the given id. If version is not specified, it retrieves the latest version of the resource. The SHA-384 hash of the data is specified in the HTTP response headers.
```
rogpeppe, line 1728: Probably worth matching the /archive phrasing here:

```
The `/resources` path returns the charm resource from the charm with the given id. etc.
```

docs/API.md
```diff
@@ -1702,34 +1703,33 @@ The above example is equivalent to the `meta/common-info` example above.
...
-Putting to the `resources` path uploads a resource (an arbitrary "blob" of
-data) associated with the charm with id series/name, which must not be a
-bundle. Stream and arch specify which of the charms resource streams and which
-architecture the resource will be associated with, respectively. Revision
-specifies the revision of the stream that's being uploaded to.
+#### PUT *id*/resources/*name*/?sha384=hash
 
-The hash value must specify the hash of the stream. If the same series, name,
-stream, revision combination is PUT again, it must specify the same hash.
+Putting to the `resources` path uploads a resource (an arbitrary "blob" of data) associated with the charm with id, which must not be a bundle. The hash value must specify the hash of the blob. A new revision of the resource will be created if the SHA-384 hash does not match the prior recorded version.
```
rogpeppe, line 1732:
```
 Note that the resource is independent of the actual revision of the
 charm; once uploaded, a resource is available in all revisions
 of *id*.
```

docs/API.md
rogpeppe, line 1706: Yes, it was deemed unnecessary to mention meta/any in all meta endpoints.
Every meta endpoint automatically gets the benefit of bulk call support.

rogpeppe:
Superceded by #598

PR #556, opened 2016-03-02
internal/v5: tests for meta/perm semantics

rogpeppe:
Also update entitycache so that it correctly passes back
the not-found error that it was given and make internal/v[45]
avoid throwing away the original error too.

Also, in internal/v[45], align assertPut with assertGet, making
put-with-admin-privileges the special case rather than the default.

### DIFF
--- a/internal/v4/api_test.go
+++ b/internal/v4/api_test.go
@@ -570,8 +570,8 @@ func (s *APISuite) addTestEntities(c *gc.C) []*router.ResolvedURL {
 		// Associate some extra-info data with the entity.
 		key := e.URL.Path() + "/meta/extra-info/key"
 		commonkey := e.URL.Path() + "/meta/common-info/key"
-		s.assertPut(c, key, "value "+e.URL.String())
-		s.assertPut(c, commonkey, "value "+e.URL.String())
+		s.assertPutAsAdmin(c, key, "value "+e.URL.String())
+		s.assertPutAsAdmin(c, commonkey, "value "+e.URL.String())
 	}
 	return testEntities
 }
@@ -620,13 +620,13 @@ func (s *APISuite) TestMetaPerm(c *gc.C) {
 		err := s.store.AddCharmWithArchive(u, storetesting.NewCharm(nil))
 		c.Assert(err, gc.IsNil)
 	}
-	s.assertGet(c, "wordpress/meta/perm?channel=unpublished", params.PermResponse{
-		Read:  []string{"charmers"},
-		Write: []string{"charmers"},
+	s.doAsUser("charmers", func() {
+		s.assertGet(c, "wordpress/meta/perm?channel=unpublished", params.PermResponse{
+			Read:  []string{"charmers"},
+			Write: []string{"charmers"},
+		})
 	})
-	e, err := s.store.FindBaseEntity(charm.MustParseURL("precise/wordpress-23"), nil)
-	c.Assert(err, gc.IsNil)
-	c.Assert(e.ChannelACLs, gc.DeepEquals, map[mongodoc.Channel]mongodoc.ACL{
+	s.assertChannelACLs(c, "precise/wordpress-23", map[mongodoc.Channel]mongodoc.ACL{
 		mongodoc.UnpublishedChannel: {
 			Read:  []string{"charmers"},
 			Write: []string{"charmers"},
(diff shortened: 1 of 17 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1588/
Test PASSed.

internal/v4/api_test.go
```diff
@@ -681,36 +683,45 @@ func (s *APISuite) TestMetaPerm(c *gc.C) {
...
-		Handler: s.srv,
-		Do:      bakeryDo(nil),
-		URL:     storeURL("precise/wordpress-23/meta/perm"),
-		ExpectBody: params.PermResponse{
-			Read:  []string{"bob", "charlie"},
-			Write: []string{"charmers"},
-		},
+	s.doAsUser("bob", func() {
+		// Check that we aren't allowed to put to the newly published entity as bob.
+		s.assertPutUnauthorized(c, "~charmers/precise/wordpress/meta/perm/read?channel=development", []string{}, `unauthorized: access denied for user "bob"`)
```
mhilton: This might not be , but looks like the inner part of this function contradicts the outer part. If you are doing a PutUnauthorized why do it as bob?

internal/v4/api_test.go
```diff
@@ -885,6 +917,85 @@ func (s *APISuite) TestMetaPerm(c *gc.C) {
 			Write: []string{},
 		},
 	})
+
+	// Restore some write rights to the stable channel.
+	s.assertPutAsAdmin(c, "trusty/wordpress-1/meta/perm/write", []string{"bob"})
+
+	// ~charmers/trusty/wordpress-1 has been published only to the
+	// stable channel. If we specify a channel in a perm put
```
frankban: s/put/PUT

frankban: s/a channel/a different channel/ ?

internal/v4/api_test.go
```diff
@@ -681,36 +683,45 @@ func (s *APISuite) TestMetaPerm(c *gc.C) {
...
+		s.assertGetUnauthorized(c, "~charmers/precise/wordpress/meta/perm/read?channel=development", `unauthorized: access denied for user "charmers"`)
 	})
 
-	e, err = s.store.FindBaseEntity(charm.MustParseURL("precise/wordpress-23"), nil)
-	c.Assert(err, gc.IsNil)
-	c.Assert(e.ChannelACLs, gc.DeepEquals, map[mongodoc.Channel]mongodoc.ACL{
+	s.doAsUser("bob", func() {
+		httptesting.AssertJSONCall(c, httptesting.JSONCallParams{
+			Handler: s.srv,
+			Do:      bakeryDo(nil),
```
mhilton, line 705: There is something tricksy here having to know to use bakeryDo if you want you're doAsUser wrapper to work. I don't think I'm too keen on this pattern. I think it will lead to confusion for no real benefit.

internal/v5/common_test.go
```diff
@@ -331,6 +333,91 @@ func (s *commonSuite) addRequiredCharms(c *gc.C, bundle charm.Bundle) {
 	}
 }
 
+func (s *commonSuite) assertPut(c *gc.C, url string, val interface{}) {
```
frankban, line 336: Thanks for moving those here.

internal/v4/api_test.go
frankban: I think this is asserting that Bob is unauthorized to PUT.

frankban:
:+1: thank you!

internal/v4/api_test.go
```diff
@@ -885,6 +917,85 @@ func (s *APISuite) TestMetaPerm(c *gc.C) {
...
+			Read:  []string{"bob"},
+			Write: []string{"admin"},
+		})
+		s.assertGet(c, "wordpress/meta/perm?channel=development", params.PermResponse{
+			Read:  []string{"bob", "charlie"},
+			Write: []string{"charmers"},
+		})
+	})
+
+	// We can't write to a channel that the charm's not in.
```
mhilton, line 959: I think this should be allowed. It's not a fundemental part of the design, but an artifact of the implementation that disallows this.

internal/v4/common_test.go
```diff
@@ -291,6 +293,91 @@ func (s *commonSuite) addRequiredCharms(c *gc.C, bundle charm.Bundle) {
...
+func (s *commonSuite) assertGet(c *gc.C, url string, expectVal interface{}) {
+	httptesting.AssertJSONCall(c, httptesting.JSONCallParams{
+		Handler:    s.srv,
+		Do:         bakeryDo(nil),
+		URL:        storeURL(url),
+		ExpectBody: expectVal,
+	})
+}
+
+// assertGetUnauthorized asserts that a GET to the given URL results
```
mhilton: wow, that's not what I expected at all reading the tests above. I expected that this was assertGet without any authorization in the request. I think this needs a much clearer name.

internal/v4/common_test.go
```diff
@@ -291,6 +293,91 @@ func (s *commonSuite) addRequiredCharms(c *gc.C, bundle charm.Bundle) {
...
+		URL:          storeURL(url),
+		ExpectStatus: http.StatusUnauthorized,
+		ExpectBody: params.Error{
+			Code:    params.ErrUnauthorized,
+			Message: expectMessage,
+		},
+	})
+}
+
+// assertGetUnauthorized asserts that a PUT to the given URL with the
```
mhilton: same as above

internal/v4/common_test.go
```diff
@@ -291,6 +293,91 @@ func (s *commonSuite) addRequiredCharms(c *gc.C, bundle charm.Bundle) {
...
+			Message: expectMessage,
+		},
+	})
+}
+
+// doAsUser calls the given function, discharging any authorization
+// request as the given user name.
+func (s *commonSuite) doAsUser(user string, f func()) {
+	old := s.discharge
+	s.discharge = dischargeForUser(user)
```
mhilton, line 376:
```
defer func (){ s.discharge = old }()
```

perhaps? that way we don't leak the authorization on a panic, like if the test fails.

mhilton:
If anything I find the tests harder to understand now than before. Also I don't think it's worth doing that re-factoring now.

internal/v4/api_test.go
rogpeppe: Isn't that what the comment is saying? Perhaps you could suggest a rewording?

internal/v4/api_test.go
rogpeppe, line 705: You have to use bakeryDo to get any authorization to work.
That's been the case for ages. I do think that the doAsUser pattern
is better than arbitrarily placed assignments to s.discharger - at
least it makes the scope clear (it helped me when trying to make changes
to the test, anyway)

internal/v4/api_test.go
mhilton: As mentioned below, I think the function name is ambiguous. perhaps assertPutNotAuthorized would be better, or perhaps assertPutStatusUnauthorized or possibly make the function assertPutError and make http.StatusUnauthorized a parameter.

internal/v4/api_test.go
rogpeppe, line 959: How would you propose changing it?
Currently we have a simple model that resolves entities before we get to any entity-specific endpoint. That's nice and easy to explain - all endpoints within an entity are treated the same.

To do it differently, you'd need to avoid resolving until the actual endpoint is encountered, but that makes bulk updates much harder (what do you do for a PUT to wordpress-1/meta/any which includes both extra-info and perm?) and everything more complex AFAICS, for not _that_ much gain.

By making the security decision in one place rather than in each individual handler, we gain more confidence in our security. I don't think that this is worth compromising that for.

Another approach that's compatible with the current scheme would be to have new meta endpoints that are specifically oriented to changing channel-specific ACLs, for example meta/perm.stable, meta/perm.development, etc.

internal/v4/common_test.go
rogpeppe: changed to assertPutIsUnauthorized

internal/v4/common_test.go
rogpeppe: ditto

internal/v4/common_test.go
rogpeppe, line 376: If we get a panic, then SetUpTest will still deal with restoring s.discharge.
But maybe worth doing anyway.

internal/v4/api_test.go
rogpeppe: both done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1589/
Test PASSed.

mhilton:
:+1:

rogpeppe:
:shipit:

jujugui:

PR #575, opened 2016-03-15
internal/charmstore: use standard debugstatus.Handler.

mhilton:
Use the standard debugstatus.Handler to serve the standard debug
endpoints. This moves all debug endpoints to be rooted at /debug and
not have version specific debug handlers.


mhilton:
Note: this has temporarily made it so that the restricted debug endpoints, like /debug/pprof, are only available by using admin credentials. This will be changed in a subsequent PR.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1647/
Test PASSed.

internal/charmstore/debug.go
```diff
@@ -91,146 +335,125 @@ func checkES(si *SearchIndex) func() error {
...
+	code := http.StatusInternalServerError
+	resp := new(bytes.Buffer)
+	defer func() {
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+		w.WriteHeader(code)
+		resp.WriteTo(w)
+	}()
 
-		fmt.Fprintln(resp, "Testing v4...")
+	fmt.Fprintln(resp, "Testing v4...")
```
rogpeppe: This looks like it's a hangover from debugging some time ago.

internal/charmstore/debug.go
```diff
@@ -14,60 +14,304 @@ import (
...
+
+func (h *debugReqHandler) check() map[string]debugstatus.CheckResult {
+	checkers := []debugstatus.CheckerFunc{
+		debugstatus.ServerStartTime,
+		h.checkElasticSearch,
+	}
+	store, err := h.store()
+	if err != nil {
+		checkers = append(checkers, func() (string, debugstatus.CheckResult) {
+			return "store available", debugstatus.CheckResult{
```
rogpeppe, line 154: Shouldn't this test be included when it succeeds too?

mhilton, line 154: It's kind of implicit.

internal/charmstore/debug.go
```diff
@@ -14,60 +14,304 @@ import (
...
+}
+
+func (h *debugReqHandler) store() (*Store, error) {
+	if h.store_ == nil && h.storeError == nil {
+		h.store_, h.storeError = h.h.params.pool.RequestStore()
+	}
+	return h.store_, h.storeError
+}
+
+func (h *debugReqHandler) check() map[string]debugstatus.CheckResult {
```
rogpeppe, line 146: // check implements the Check function as found in
// debugstatus.Handler.Check.

?

internal/charmstore/debug.go
```diff
@@ -14,60 +14,304 @@ import (
...
 	appver "gopkg.in/juju/charmstore.v5-unstable/version"
 )
 
-// GET /debug/info .
-func serveDebugInfo(http.Header, *http.Request) (interface{}, error) {
-	return appver.VersionInfo, nil
+type debugParams struct {
+	ServerParams
+	pool            *Pool
+	loopbackHandler http.Handler
```
rogpeppe, line 37: A comment on this field might be nice.

internal/charmstore/debug.go
```diff
@@ -14,60 +14,304 @@ import (
...
+	}
+	if u != h.params.AuthUsername || p != h.params.AuthPassword {
+		return errgo.WithCausef(nil, params.ErrUnauthorized, "username or password mismatch")
+	}
+	return nil
+}
+
+func (h *debugHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// The charmstore router will have stripped "/debug" off the
+	// front of the path, put it back.
```
rogpeppe: s/,/;/

rogpeppe: Might be good to state _why_ we're putting it back.
i.e.
// ... put it back. We need it because the handlers that debugstatus
// defined have the /debug prefix.

internal/charmstore/debug.go
```diff
@@ -14,60 +14,304 @@ import (
...
+		return time.Time{}, time.Time{}, errgo.Notef(err, "Cannot query logs")
+	}
+	return
+}
+
+type debugCheckRequest struct {
+	httprequest.Route `httprequest:"GET /debug/check"`
+}
+
+// GET /debug/check.
```
rogpeppe: It would be nice to have a comment on this saying how it
differs from /debug/status and why.

internal/charmstore/debug.go
```diff
@@ -14,60 +14,304 @@ import (
...
-func debugCheck(checks map[string]func() error) http.Handler {
-	return router.HandleJSON(func(http.Header, *http.Request) (interface{}, error) {
-		n := len(checks)
-		type result struct {
-			name string
-			err  error
+type debugHandler struct {
+	router    *httprouter.Router
+	params    debugParams
+	checks    map[string]func() error
```
rogpeppe, line 50: Let's have some comments saying what these fields do. It's not
clear how or why the checks here differ from the ones used
for servicing debug/status, or what openPaths is for.

internal/charmstore/debug.go
mhilton, line 37: done

internal/charmstore/debug.go
mhilton: done

internal/charmstore/debug.go
mhilton, line 50: done

internal/charmstore/debug.go
mhilton, line 146: done

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1659/
Test PASSed.

internal/charmstore/debug.go
rogpeppe, line 154: Yeah, but it's kinda strange to have a test entry that only exists when there's an error.

mhilton:
This is unlikely to be needed now

PR #580, opened 2016-03-22
Add a collection and doc types for resources.

ericsnowcurrently:
We store resources in a separate collection from entities.  There are 2 doc types:
- Resource - info about a resource at a revision
- Resources - maps resource name to revision for a channel/resolved-URL pair


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1664/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1665/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1670/
Test PASSed.

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,182 @@
...
+	"gopkg.in/errgo.v1"
+	"gopkg.in/juju/charm.v6-unstable"
+	"gopkg.in/juju/charm.v6-unstable/resource"
+	"gopkg.in/juju/charmrepo.v2-unstable/csclient/params"
+	"gopkg.in/mgo.v2/bson"
+)
+
+// NewResourceQuery formats the provided details into a mongo query for
+// the identified resource.
+func NewResourceQuery(cURL *charm.URL, resName string, revision int) bson.D {
```
kat-co, line 20: Can we expand cURL into charmURL?

internal/mongodoc/resource_test.go
```diff
@@ -0,0 +1,369 @@
+// Copyright 2016 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package mongodoc_test // import "gopkg.in/juju/charmstore.v5-unstable/internal/mongodoc"
```
kat-co: Looks like an old import made it in there.

kat-co:
LGTM

internal/charmstore/store.go
```diff
@@ -1086,6 +1095,12 @@ func (s StoreDatabase) BaseEntities() *mgo.Collection {
 	return s.C("base_entities")
 }
 
+// Resources returns the mongo collection where resources are stored.
+func (s StoreDatabase) Resources() *mgo.Collection {
+	// TODO(ericsnow) Re-use the Entities collection?
```
natefinch: maybe we can remove this todo?  I think we decided we're not reusing entities, because resources aren't really the same thing, right?

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,182 @@
...
+
+	if doc.UploadTime.IsZero() {
+		return errgo.New("missing upload timestamp")
+	}
+
+	return nil
+}
+
+// SortResources sorts the provided resource docs.
+func SortResources(resources []*Resource) {
```
natefinch, line 115: I presume this will be used later (since it's not used now).  There seems to be no reason to pass around pointers to resources instead of just the value itself.  There's no reason to use a pointer, it just makes everything uglier and more difficult.

internal/mongodoc/package_test.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2015 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package mongodoc_test // import "gopkg.in/juju/charmstore.v5-unstable/internal/mongodoc"
```
natefinch, line 4: You only need one canonical import path specifier per package, so let;'s not proliferate them everywhere, otherwise we have to update all of them when the package moves.

natefinch:
LGTM with a couple small caveats.

internal/mongodoc/resource.go
ericsnowcurrently, line 20: We're pretty consistent on the spelling throughout the various code bases.  So I'd rather stick with the shortened name.  Otherwise I agree.

internal/mongodoc/package_test.go
ericsnowcurrently, line 4: While you are correct, note that each existing file in this code base duplicates the specifier.  Given that the precedent is mildly inconsistent (only 3 files don't do it), I don't mind dropping the specifier.

internal/charmstore/store.go
ericsnowcurrently: done

internal/mongodoc/resource.go
ericsnowcurrently, line 115: While I agree, this is an existing code base with established patterns that I've tried to respect.  Other mongo docs are passed around as pointers.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1673/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1682/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1684/
Test PASSed.

internal/mongodoc/package_test.go
mhilton, line 4: In this case the specifier doesn't match the package name, so almost certainly shouldn't be there. They would have been put there in the first place, by some over-eager automated script.

internal/mongodoc/resource.go
mhilton, line 20: I'd suggest that cURL is unusual for the charmstore, but as that isn't particularly consistent within itself I'm not that worried.

internal/mongodoc/package_test.go
rogpeppe, line 4: If in doubt, run github.com/rogpeppe/misc/cmd/importcomment.

I think it's reasonable to have the comment on all files, as there's no "primary" file in any given package.

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+	"time"
+
+	"gopkg.in/errgo.v1"
+	"gopkg.in/juju/charm.v6-unstable"
+	"gopkg.in/juju/charm.v6-unstable/resource"
+	"gopkg.in/juju/charmrepo.v2-unstable/csclient/params"
+	"gopkg.in/mgo.v2/bson"
+)
+
+// NewResourceQuery formats the provided details into a mongo query for
```
rogpeppe, line 18: Other than a very few things, we've tried to keep business logic outside of the mongodoc package. The place for this is in the internal/charmstore package (and even then it might not warrant exporting, depending on what requirements are).

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	return query
+}
+
+// Resource holds the in-database representation of a charm resource
+// at a particular revision.
+type Resource struct {
+	// CharmURL identifies the unresolved charm associated with this
+	// resource.
+	CharmURL *charm.URL `bson:"unresolved-charm-url"`
```
mhilton, line 40: Could you add a comment saying what unresolved means in this context? 

Also, if it's a URL without a revision that's fine. If it's a url without a series or revision then I think I'd prefer it be called BaseURL to be consistent with the rest of the database.

rogpeppe, line 40: We call this BaseURL in other places.

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+		{"name", resName},
+	}
+	if revision >= 0 {
+		query = append(query, bson.DocElem{"revision", revision})
+	}
+	return query
+}
+
+// Resource holds the in-database representation of a charm resource
+// at a particular revision.
```
rogpeppe, line 36: Perhaps mention that the (url, name, revision) triple forms the primary key tuple?

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+func (doc Resource) HashString() string {
+	return fmt.Sprintf("%x", doc.Fingerprint)
+}
+
+// SortResources sorts the provided resource docs.
+func SortResources(resources []*Resource) {
+	sort.Sort(resourcesByName(resources))
+}
+
+type resourcesByName []*Resource
```
mhilton, line 119: As this sorts by name and then revision could you add a comment somewhere that tells you what the sort actually does?

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Name is the name of the resource as defined in the charm
+	// metadata.
+	Name string `bson:"name"`
+
+	// Revision identifies the specific revision of the resource.
+	Revision int `bson:"revision"`
+
+	// Fingerprint is the checksum of the resource file.
+	Fingerprint []byte `bson:"fingerprint"`
```
rogpeppe, line 50: Perhaps call this BlobHash to match the same thing in the Entity document?

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Size is the size of the resource file, in bytes.
+	Size int64 `bson:"size"`
```
rogpeppe, line 53: BlobSize

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// BlobName holds the name that the resource blob is given in the
+	// blob store.
+	BlobName string
+
+	// UploadTime is the UTC timestamp from when the resource file was
```
rogpeppe, line 59: No need to mention UTC here, I think.
Just "is the time the resource file was stored in the blob store".

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+	// stored in the blob store.
+	UploadTime time.Time
+}
+
+// Validate ensures that the doc is valid.
+func (doc Resource) Validate() error {
+	if doc.CharmURL == nil {
+		return errgo.New("missing charm URL")
+	}
+	if doc.CharmURL.Revision != -1 {
```
rogpeppe, line 69: Perhaps better to check that *BaseURL(doc.CharmURL) == *doc.CharmURL ?

mhilton:
:+1: Modulo a few thoughts.

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+		return errgo.New("missing name")
+	}
+	if doc.Revision < 0 {
+		return errgo.Newf("got negative revision %d", doc.Revision)
+	}
+
+	if len(doc.Fingerprint) == 0 {
+		return errgo.New("missing fingerprint")
+	}
+	fp, err := resource.NewFingerprint(doc.Fingerprint)
```
rogpeppe, line 86: Honestly, I wouldn't worry too much about verifying whether the hash looks OK here. Checking that it's non-empty should be fine, and it's fail-safe.

internal/mongodoc/resource.go
rogpeppe, line 50: Also for consistency, I'd suggest storing it in ASCII hex.

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+	return nil
+}
+
+// HashString returns the resource's hex-encoded fingerprint.
+func (doc Resource) HashString() string {
+	return fmt.Sprintf("%x", doc.Fingerprint)
+}
+
+// SortResources sorts the provided resource docs.
+func SortResources(resources []*Resource) {
```
rogpeppe, line 115: Let's be consistent about whether we use *Resource or Resource.
Given that everywhere else we use *Entity and *BaseEntity, I'd suggest that we use *Resource throughout, so I'd suggest defining the methods on *Resource too.

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+	sort.Sort(resourcesByName(resources))
+}
+
+type resourcesByName []*Resource
+
+func (sorted resourcesByName) Len() int      { return len(sorted) }
+func (sorted resourcesByName) Swap(i, j int) { sorted[i], sorted[j] = sorted[j], sorted[i] }
+func (sorted resourcesByName) Less(i, j int) bool {
+	if sorted[i].Name < sorted[j].Name {
```
rogpeppe, line 124: This doesn't sort the resources properly - it ignores the URL.

Also, this doesn't implement Less correctly - the results
could be inconsistent and could actually cause a panic.

I think you want something more like this:

```
 func (s resourcesByName) Less(i, j int) bool {
      r0, r1 := s[i], s[j]
      if *r0.BaseURL != *r1.BaseURL {
           return r0.BaseURL.String() < r1.BaseURL.String()
      }
      if r0.Name != r1.Name {
          return r0.Name < r1.Name
      }
      return r0.Revision < r1.Revision
}
```

internal/mongodoc/resource.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	if channel != params.NoChannel {
+		query = append(query, bson.DocElem{"channel", channel})
+	}
+	return query
+}
+
+// Resources identifies the set of resource revisions for a resolved
+// charm, relative to a specific channel,
+type Resources struct {
```
rogpeppe, line 148: I think I'd put this data in the base entity along with the other channel-related fields.
Something like this perhaps?

```
// ChannelResources holds the set of resources currently
// associated with each channel.
ChannelResources map[params.Channel] []ResourceId

type ResourceId struct {
    Name string
    Revision int
}
```

internal/mongodoc/resource_test.go
```diff
@@ -0,0 +1,369 @@
...
+	query := mongodoc.NewResourceQuery(cURL, "eggs", -1)
+
+	c.Check(query, jc.DeepEquals, bson.D{
+		{"unresolved-charm-url", charm.MustParseURL("cs:spam")},
+		{"name", "eggs"},
+	})
+}
+
+func (s *ResourceSuite) TestValidateFull(c *gc.C) {
+	doc := mongodoc.Resource{
```
rogpeppe, line 51: Let's make all these into a table-driven test for consistency with all the other code, please. It makes it much easier to see the consistency between the tests.

internal/mongodoc/resource_test.go
```diff
@@ -0,0 +1,369 @@
...
+)
+
+type ResourceSuite struct{}
+
+func (s *ResourceSuite) TestNewResourceQueryWithRevision(c *gc.C) {
+	cURL := charm.MustParseURL("cs:trusty/spam-2")
+
+	query := mongodoc.NewResourceQuery(cURL, "eggs", 3)
+
+	c.Check(query, jc.DeepEquals, bson.D{
```
rogpeppe, line 32: This really doesn't test anything worth testing. Without testing this against an actual mongo database, this doesn't give us any assurance at all that the code will really work - we're just testing that what we return is what we return, not that it will actually work as the expected query.

More reason for the resource query and associated tests to live in the internal/charmstore package.

rogpeppe:
This seems like a good start thanks, modulo a bunch of comments and suggestions above.

internal/mongodoc/resource.go
ericsnowcurrently, line 18: Fair enough.  I was trying to keep code that has to refer to the BSON field names next to the doc, but I'll stick to your guidelines.

mhilton:
I've landed this as #593

PR #592, opened 2016-04-05
Allow resolver@charmstore group to resolve any charm url.

alesstimec:


internal/router/router.go
```diff
@@ -406,20 +412,27 @@ func (r *Router) willIncludeMetadata(req *http.Request) {
 }
 
 func (r *Router) serveMetaGet(rurl *ResolvedURL, req *http.Request) (interface{}, error) {
+	key, path := handlerKey(req.URL.Path)
+	if key == "any" {
+		// TODO: consider whether we might want the capability to
+		// have different permissions for different meta endpoints.
+		if err := r.Context.AuthorizeEntity(rurl, req, resolverGroup); err != nil {
```
mhilton, line 419: This is probably dangerous. I suspect for the any key we would end up with the resolve group having permissions for everything. I don't think we individually check the meta endpoints.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1697/
Test PASSed.

internal/router/router.go
alesstimec, line 419: How so?

PR #583, opened 2016-03-24
Add Store.ListResources().

ericsnowcurrently:
This patch also sets the stage for adding the other resource-related methods.

Note: This PR is based on another (#580).

PR #594, opened 2016-04-06
resources store methods list

mhilton:
This is an updated version of #583


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1702/
Test PASSed.

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+var resourceNotFound = errgo.Newf("resource not found")
+
+// ListResources returns the set of resources for the charm. If the
+// unpublished channel is specified then set is composed of the latest
+// revision for each resource. Otherwise it holds the revisions declared
+// when the charm/channel pair was published.
+func (s *Store) ListResources(entity *mongodoc.Entity, channel params.Channel) ([]*mongodoc.Resource, error) {
+	if entity.URL.Series == "bundle" {
```
rogpeppe, line 62: Or just return nil, nil ?

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+		return nil, errgo.Newf("bundles do not have resources")
+	}
+	if entity.CharmMeta == nil {
+		return nil, errgo.Newf("entity missing charm metadata")
+	}
+	baseEntity, err := s.FindBaseEntity(entity.URL, FieldSelector("channelresources"))
```
rogpeppe, line 68: We really need to make this go through the cache at some point.

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return nil, errgo.Mask(err)
+	}
+	var docs []*mongodoc.Resource
+	for name, meta := range entity.CharmMeta.Resources {
+		doc, err := s.latestResource(baseEntity, channel, name)
```
rogpeppe: When the channel is unpublished, presumably this _could_ be more efficient than making a query for each resource - it could do a single query on the resources collection for all resources with a matching URL and filter out all those that aren't mentioned in the charm metadata. If you don't fancy doing that now, perhaps add a TODO?

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+		if errgo.Cause(err) == resourceNotFound {
+			// Convert the "not found" error into placeholder. We do
+			// this instead of pre-populating the database with
+			// placeholders. This keeps the machinery simpler while
+			// still correct.
```
rogpeppe: I don't understand this. Why are we returning resources that don't exist?

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+			return errgo.WithCausef(nil, params.ErrDuplicateUpload, "")
+		}
+		return errgo.Notef(err, "cannot insert resource")
+	}
+	return nil
+}
+
+func (s *Store) storeResource(blob ResourceBlob) (string, error) {
+	name := bson.NewObjectId().Hex()
+	// TODO(ericsnow) We will finish this in a follow-up patch.
```
rogpeppe: // TODO(mhilton) actually store the resource blob

?

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+	return name, nil
+}
+
+func (s *Store) setResource(entity *mongodoc.Entity, channel params.Channel, name string, revision int) error {
```
rogpeppe: // setResource adds or updates the resource for the given entity,
// associating the resource with the given name and revision
// with the given channel.

?

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+	doc, err := s.resource(entity.BaseURL, name, revision)
+	if err != nil {
+		return errgo.Mask(err, errgo.Is(resourceNotFound))
+	}
+
+	if !charmHasResource(entity.CharmMeta, doc.Name) {
+		return errgo.Mask(err)
+	}
+
+	update := bson.D{{fmt.Sprintf("channelresources.%s.%s", channel, name), revision}}
```
rogpeppe: Looks like this could be a problem if resource names might contain dots or other nasty characters.

You'll have to either permanently exclude nasty characters from the resource name, encode resource names, or use a slice in the doc, I think.

internal/charmstore/resources_test.go
```diff
@@ -0,0 +1,256 @@
...
+	for i, doc := range docs {
+		meta := ch.Meta().Resources[doc.Name]
+		blob, err := os.Open(filepath.Join(ch.Path, meta.Path))
+		c.Assert(err, jc.ErrorIsNil)
+		docs[i].Revision = addResource(c, store, entity, channel, doc, blob)
+	}
+	return docs
+}
+
+func addResource(c *gc.C, store *Store, entity *mongodoc.Entity, channel params.Channel, doc *mongodoc.Resource, blobReader io.Reader) int {
```
rogpeppe: A doc comment might be nice here. Why do we need to use the unexported charmstore API here BTW?

internal/charmstore/resources.go
rogpeppe: Is this method ever going to be useful in production code? Won't we always be publishing all the resources for a given channel at the same time?

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+		err := s.DB.Resources().Find(bson.D{{"baseurl", baseEntity.URL}, {"name", name}}).Sort("-revision").One(&res)
+		if err == mgo.ErrNotFound {
+			return -1, resourceNotFound
+		}
+		if err != nil {
+			return -1, errgo.Mask(err)
+		}
+		return res.Revision, nil
+	}
+	revision, ok := baseEntity.ChannelResources[channel][name]
```
rogpeppe: Maybe get this easy case out of the way first to save indentation in the harder case?

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+	if channel == params.UnpublishedChannel {
+		return errgo.New("cannot publish to unpublished channel")
+	}
+
+	doc, err := s.resource(entity.BaseURL, name, revision)
+	if err != nil {
+		return errgo.Mask(err, errgo.Is(resourceNotFound))
+	}
+
+	if !charmHasResource(entity.CharmMeta, doc.Name) {
```
rogpeppe: Maybe do this test before hitting the database to fetch the resource info?

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+		return res.Revision, nil
+	}
+	revision, ok := baseEntity.ChannelResources[channel][name]
+	if !ok {
+		return -1, resourceNotFound
+	}
+	return revision, nil
+}
+
+func (s *Store) resource(url *charm.URL, name string, revision int) (*mongodoc.Resource, error) {
```
rogpeppe, line 226: // resource returns the resource with the given URL, name and revision.
// If the resource was not found, it returns an error with a resourceNotFound
// cause.

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+	query := newResourceQuery(url, name, revision)
+	var doc mongodoc.Resource
+	err := s.DB.Resources().Find(query).One(&doc)
+	if err == mgo.ErrNotFound {
+		err = resourceNotFound
```
rogpeppe, line 231: or params.ErrNotFound I guess.

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return nil, errgo.Mask(err)
+	}
+	if err := doc.Validate(); err != nil {
+		return nil, errgo.Notef(err, "got bad data from DB")
+	}
+	return &doc, nil
+}
+
+func (s *Store) publishedResources(curl *charm.URL, channel params.Channel) ([]*mongodoc.Resource, error) {
```
rogpeppe: I think I'd leave this out until we actually need it. As it is, it's neither tested nor used AFAICS.

internal/charmstore/resources.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	baseEntity, err := s.FindBaseEntity(curl, FieldSelector("channelresources"))
+	if err != nil {
+		return nil, errgo.Mask(err, errgo.Is(params.ErrNotFound))
+	}
+	if len(baseEntity.ChannelResources[channel]) == 0 {
+		return nil, nil
+	}
+	resources := make([]*mongodoc.Resource, 0, len(baseEntity.ChannelResources[channel]))
+	for name, revision := range baseEntity.ChannelResources[channel] {
```
rogpeppe: This could probably be more efficient (concurrent or a single query).

internal/charmstore/resources_test.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func extractResources(c *gc.C, cURL *charm.URL, ch *charm.CharmDir) []*mongodoc.Resource {
+	copied := *cURL
+	cURL = &copied
+	cURL.Revision = -1
+	cURL.Series = ""
+	var docs []*mongodoc.Resource
+	for _, meta := range ch.Meta().Resources {
+		data, err := ioutil.ReadFile(filepath.Join(ch.Path, meta.Path))
```
rogpeppe: I don't get this. The Path attribute is supposed to represent where
the resource is stored inside the runtime charm data directory, not
the actual charm itself. Storing the resources inside the charm itself
seems kinda wrong. Also, this looks like we're entrenching the pattern
of using charm dirs, where we'd really like to move away from them
towards in-memory charms etc.

I'd prefer to see some in-memory solution, perhaps fabricating known
contents for the resources, or using a function argument to make that
pluggable if needed.

internal/charmstore/resources_test.go
```diff
@@ -0,0 +1,256 @@
...
+	expected[1].Revision = addResource(c, store, entity, channel, expected[1], nil)
+	expected[2].Revision = addResource(c, store, entity, channel, expected[2], nil)
+
+	docs, err := store.ListResources(entity, channel)
+	c.Assert(err, jc.ErrorIsNil)
+
+	checkResourceDocs(c, docs, expected)
+}
+
+func addResources(c *gc.C, store *Store, curl *charm.URL, channel params.Channel, entity *mongodoc.Entity, ch *charm.CharmDir) []*mongodoc.Resource {
```
rogpeppe: Doc comment please. And like I said below, I'd prefer if this didn't rely on files in the testing charm repo.

internal/charmstore/store.go
```diff
@@ -730,14 +730,24 @@ func (s *Store) UpdateEntity(url *router.ResolvedURL, update bson.D) error {
 // url. If there are no entries in update then no update is performed,
 // and no error is returned.
 func (s *Store) UpdateBaseEntity(url *router.ResolvedURL, update bson.D) error {
+	if err := s.updateBaseEntity(mongodoc.BaseURL(&url.URL), update); err != nil {
+		return errgo.NoteMask(err, fmt.Sprintf("cannot update base entity for %q", url), errgo.Is(params.ErrNotFound))
```
rogpeppe: Why not just do this inside UpdateBaseEntity rather than having two functions?

rogpeppe:
This looks nice in general, thanks, but the resource names need sorting out, and I think
the testing could be improved somewhat.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1705/
Test PASSed.

internal/charmstore/resources.go
```diff
@@ -0,0 +1,244 @@
...
+		*resource = r
+		if _, ok := resources[r.Name]; !ok {
+			resources[r.Name] = make(map[int]*mongodoc.Resource)
+		}
+		resources[r.Name][r.Revision] = resource
+		if r.Revision >= latest[r.Name] {
+			latest[r.Name] = r.Revision
+		}
+	}
+	return resources, latest, iter.Close()
```
rogpeppe: I think it would be better to return zero values if the iterator fails.

```
if err := iter.Close(); err != nil {
    return nil, nil, errgo.Mask(err)
}
```

internal/charmstore/resources.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// charmResources returns all of the currently stored resources for a charm.
+func (s *Store) charmResources(baseURL *charm.URL) (map[string]map[int]*mongodoc.Resource, map[string]int, error) {
+	resources := make(map[string]map[int]*mongodoc.Resource)
+	latest := make(map[string]int)
+	iter := s.DB.Resources().Find(bson.D{{"baseurl", baseURL}}).Iter()
+	var r mongodoc.Resource
+	for iter.Next(&r) {
+		resource := new(mongodoc.Resource)
```
rogpeppe: or:
    resource := *r

then use &r when assigning to the map?

internal/charmstore/resources.go
```diff
@@ -69,143 +69,153 @@ func (s *Store) ListResources(entity *mongodoc.Entity, channel params.Channel) (
...
 		UploadTime: time.Now().UTC(),
-	}
-	if s.insertResource(doc); err != nil {
+	})
+	if err != nil {
 		if err := s.BlobStore.Remove(blobName); err != nil {
 			logger.Errorf("cannot remove blob %s after error: %v", blobName, err)
 		}
-		return errgo.Mask(err)
+		return nil, errgo.Mask(err)
```
rogpeppe, line 153: Do we want to preserve the ErrDuplicateUpload cause maybe?

internal/charmstore/resources.go
mhilton, line 62: done

internal/charmstore/resources.go
mhilton, line 68: Not now

internal/charmstore/resources.go
mhilton: I've changed how the mechanism works so that it gets the list of possible resources first and then works out which ones it wants.

internal/charmstore/resources.go
mhilton: No idea, I've stopped doing that. For now I just miss out any missing ones, but perhaps it should be an error.

internal/charmstore/resources.go
mhilton: I implemented upload, mainly because it made testing easier

internal/charmstore/resources.go
mhilton: I've replaced it with a method that does the things we will actually do, publish a set of resources.

internal/charmstore/resources.go
mhilton: Updated as per your previous suggestion.

internal/charmstore/resources_test.go
mhilton: done. The unexported charmstore API seemed like a good idea at the time.

internal/charmstore/resources.go
mhilton: This has been replaced

internal/charmstore/resources.go
mhilton: This has been replaced.

internal/charmstore/resources.go
mhilton, line 231: Possibly I think I'll look into that in the later PRs

rogpeppe:
LGTM with a couple of trivial suggestions, thanks!

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1706/
Test PASSed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1707/
Test PASSed.

mhilton:
:shipit:

jujugui:

PR #597, opened 2016-04-08
add placeholders for resources that are not present

mhilton:


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1712/
Test PASSed.

internal/charmstore/resources_test.go
```diff
@@ -170,11 +170,16 @@ func (s *resourceSuite) TestListResourcesResourceNotFound(c *gc.C) {
 	sortResources(expected)
 	err := store.PublishResources(entity, channel, resourceRevisions(expected[1:]))
```
rogpeppe: I don't think PublishResources should allow publishing without all the resources
that are needed by the charm, so I'd write this test a bit differently.

rogpeppe:
Looks reasonable except for the fact that we shouldn't
be allowed to do what the test is doing here.
Land this then fix that if it's easier.

jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1720/
Test FAILed.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1721/
Test PASSed.

rogpeppe:
LGTM thanks!

urosj:
👍 However, have a bit of reservations how the whole experience would look like. If we can reserve a name for the resource but not require the resource to be uploaded, then it seems fine to return -1 for the revision.

mhilton:
:shipit:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1722/
Test PASSed.

PR #603, opened 2016-04-19
internal/v5: implement meta/can-ingest endpoint

rogpeppe:
This will make it possible for charmload to ignore charms
that are under manual control.

### DIFF
--- a/internal/mongodoc/doc.go
+++ b/internal/mongodoc/doc.go
@@ -200,6 +200,12 @@ type BaseEntity struct {
 	// set of resource names holding the currently published resource
 	// version for that channel and resource name.
 	ChannelResources map[params.Channel][]ResourceRevision
+	
+	// NoIngest is set to true when a charm or bundle has been uploaded
+	// with a POST request. Since the ingester only uses PUT requests
+	// at present, this signifies that someone has taken over control from
+	// the ingester.
+	NoIngest bool	`bson:",omitempty"`
 }
 
 // ResourceRevision specifies an association of a resource name to a
--- a/internal/v5/archive.go
+++ b/internal/v5/archive.go
@@ -56,6 +56,11 @@ func (h *ReqHandler) serveArchive(id *charm.URL, w http.ResponseWriter, req *htt
 		// TODO: investigate using 100-Continue statuses to prevent
 		// unnecessary uploads.
 		defer io.Copy(ioutil.Discard, req.Body)
+
+		// Add "noingest" so that we don't need to fetch the base entity
+		// twice in the common case when uploading a charm.
+		h.Cache.AddBaseEntityFields(charmstore.FieldSelector("noingest"))
+
 		if err := h.authorizeUpload(id, req); err != nil {
 			return errgo.Mask(err, errgo.Any)
 		}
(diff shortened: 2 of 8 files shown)


jujugui:
Test FAILed.
http://ci.jujugui.org:8080//job/charmstore/1740/
Test FAILed.

internal/mongodoc/doc.go
```diff
@@ -200,6 +200,12 @@ type BaseEntity struct {
 	// set of resource names holding the currently published resource
 	// version for that channel and resource name.
 	ChannelResources map[params.Channel][]ResourceRevision
+	
+	// NoIngest is set to true when a charm or bundle has been uploaded
+	// with a POST request. Since the ingester only uses PUT requests
+	// at present, this signifies that someone has taken over control from
+	// the ingester.
+	NoIngest bool	`bson:",omitempty"`
```
mhilton, line 208: I don't think I like the name NoIngest. It seems to represent the wrong thing. I think I'd prefer a name that represents what happened and then the can-ingest endpoint can decide based on that whether it's allowed. So perhaps this should be "Posted" or "Pushed".

mhilton:
The code looks good as far as it goes. I'm slightly worried about the concept though. Won't this result in a situation where the revision numbering will get out of sync with legacy charm stores. Is that an OK situation?

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1741/
Test PASSed.

internal/mongodoc/doc.go
rogpeppe, line 208: I'm not sure. I considered Posted or Pushed, but I decided against that because we're actually interested in what will be returned from the can-ingest endpoint, and we might decide to set the NoIngest flag based on other things too (for example, we might let people explicitly turn off ingestion without uploading something, in which case "Posted" would be inaccurate).

rogpeppe:
> Won't this result in a situation where the revision numbering will get out of sync with legacy charm stores?

Absolutely. But we're already in that position if someone uploads a charm manually. This at least will make it possible to avoid a war between manual and ingested uploads.

internal/v5/archive.go
```diff
@@ -193,6 +198,22 @@ func (h *ReqHandler) servePostArchive(id *charm.URL, w http.ResponseWriter, req
...
 	}
+	// Find the base entity before trying to update its noingest status
+	// as if this isn't the first upload of the entity, the base entity
+	// will already be cached, so we won't need any more round trips
+	// to mongo than usual.
+	baseEntity, err := h.Cache.BaseEntity(&rid.URL, charmstore.FieldSelector("noingest"))
+	if err != nil || !baseEntity.NoIngest {
+		if err := h.Store.DB.BaseEntities().UpdateId(mongodoc.BaseURL(&rid.URL), bson.D{{
+			"$set", bson.D{{
+				"noingest", true,
```
urosj, line 209: I'd suggest to use "pushed" instead to better reflect why we set that?

urosj:
LGTM with mandatory bike-shedding :)

rogpeppe:
:shipit:

jujugui:

PR #665, opened 2016-08-01
Check terms existence

alesstimec:
This PR prevents users from uploading charms that require agreement to non existent terms (such charms would not be deployable).


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1862/
Test PASSed.

PR #670, opened 2016-08-04
Entity document: include publishing info in a single published field.

frankban:
Replace the Stable and Edge fields: this scales better in preparation for adding two additional channels.
The corresponding migration is implemented in the branch.

Also refactor the code so that there is a single point where the list of channels is defined, in order.

### DIFF
--- a/internal/charmstore/addentity.go
+++ b/internal/charmstore/addentity.go
@@ -560,17 +560,32 @@ func (s *Store) addCharm(c charm.Charm, p addParams) (err error) {
 }
 
 // setEntityChannels associates the entity with the given channels, ignoring
-// unknown channels.
+// unknown channels and the unpublished channel.
 func setEntityChannels(entity *mongodoc.Entity, chans []params.Channel) {
+	entity.Published = make(map[params.Channel]bool, len(chans))
 	for _, c := range chans {
-		switch c {
-		case params.EdgeChannel:
-			entity.Edge = true
-		case params.StableChannel:
-			entity.Stable = true
+		if ValidChannels[c] && c != params.UnpublishedChannel {
+			entity.Published[c] = true
 		}
 	}
 }
+
+// OrderedChannels holds the list of valid channels in order of publishing
+// status, most stable first.
+var OrderedChannels = []params.Channel{
+	params.StableChannel,
+	params.EdgeChannel,
+	params.UnpublishedChannel,
+}
+
+// ValidChannels holds the set of all allowed channels for an entity.
+var ValidChannels = func() map[params.Channel]bool {
+	channels := make(map[params.Channel]bool, len(OrderedChannels))
+	for _, ch := range OrderedChannels {
+		channels[ch] = true
+	}
+	return channels
+}()
 
 // addBundle adds a bundle to the entities collection with the given
 // parameters. If p.URL cannot be used as a name for the bundle then the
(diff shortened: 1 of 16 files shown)


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1873/
Test PASSed.

internal/v5/api.go
```diff
@@ -1460,12 +1448,16 @@ func (h *ReqHandler) servePromulgate(id *router.ResolvedURL, w http.ResponseWrit
...
 
-// validPublishChannels holds the set of channels that can
-// be the target of a publish request.
-var validPublishChannels = map[params.Channel]bool{
-	params.EdgeChannel:   true,
-	params.StableChannel: true,
-}
+// ValidChannels holds the set of all allowed channels
+// that can be passed as a "?channel=" parameter.
+var ValidChannels = func() map[params.Channel]bool {
```
rogpeppe: I think the distinction between this and charmstore.ValidChannels is a bit confusing. Can't we make do with only one ValidChannels value?

internal/charmstore/addentity.go
```diff
@@ -562,16 +562,30 @@ func (s *Store) addCharm(c charm.Charm, p addParams) (err error) {
...
-			entity.Stable = true
+		if ValidChannels[c] {
+			entity.Published[c] = true
 		}
 	}
 }
 
+// OrderedChannels holds the list of valid channels in order of publishing
+// status, most stable first.
+var OrderedChannels = []params.Channel{
```
rogpeppe, line 575: If we had UnpublishedChannel here, then we could avoid defining a separate ValidChannels variable in v5, I think.

mhilton:
:+1:

internal/charmstore/store.go
```diff
@@ -1068,9 +1059,10 @@ func (s *Store) SetPromulgated(url *router.ResolvedURL, promulgate bool) error {
 
 // SetPerms sets the ACL specified by which for the base entity with the
 // given id. The which parameter is in the form "[channel].operation",
-// where channel, if specified, is one of "edge" or "stable" and operation
-// is one of "read" or "write". If which does not specify a channel then
-// the unpublished ACL is updated. This is only provided for testing.
+// where channel, if specified, is the string corresponding to one of the
+// ValidChannels and operation is one of "read" or "write". If which does
+// not specify a channel then the unpublished ACL is updated.
```
rogpeppe: It looks to me as if which must always specify a channel, otherwise the db update will be wrong.

That is, the which parameter is in the form "channel.operation" not "[channel].operation".

rogpeppe:
This is an excellent change, thanks. I have a few suggestions around things that confused me when reading, but otherwise this LGTM.

internal/charmstore/store.go
frankban: Done.

internal/charmstore/addentity.go
frankban, line 575: Done.

internal/v5/api.go
frankban: Done.

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1874/
Test PASSed.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1875/
Test PASSed.

PR #671, opened 2016-08-04
Introduce beta and candidate channels.

frankban:
Also add a migration for populating their corresponding base entity ACLs.


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1877/
Test PASSed.

internal/charmstore/migrations.go
```diff
@@ -222,3 +226,35 @@ func migratePublishedEntities(db StoreDatabase) error {
...
+	baseEntities := db.BaseEntities()
+	iter := baseEntities.Find(bson.D{{
+		// Assume that, if a base entity does not have the "channelacls.beta"
+		// field, then the "channelacls.candidate" one is also missing and the
+		// document must be migrated.
+		"channelacls.beta", bson.D{{"$exists", false}},
+	}}).Select(map[string]int{"channelacls": 1}).Iter()
+
+	// For every resulting base entity populate "channelacls.beta" and
+	// "channelacls.candidate" with contents from "channelacls.unpublished".
```
mhilton, line 242: Why unpublished? I'd have assumed that edge would have been closer?

frankban, line 242: Unpublished is supposed to be more restrictive, so safer default?

mhilton:
:+1:

internal/charmstore/migrations.go
rogpeppe, line 242: I think I agree that "edge" might be a better default, as "edge" is more restrictive than both "beta" and "candidate" as I understand it.

internal/charmstore/migrations_integration_test.go
```diff
@@ -384,6 +395,14 @@ var migrationFromDumpBaseEntityTests = []struct {
 				Read:  []string{"bobgroup"},
 				Write: []string{"bob", "someoneelse"},
 			},
+			params.BetaChannel: {
+				Read:  []string{"bobgroup"},
+				Write: []string{"bob", "someoneelse"},
+			},
```
rogpeppe, line 427: ISTM that it would be good to have an entity with different unpublished and edge ACLs so that we're verifying that the new channels are populated from the correct source.

internal/v4/api_test.go
```diff
@@ -668,6 +668,14 @@ func (s *APISuite) TestMetaPerm(c *gc.C) {
 			Read:  []string{"charmers"},
 			Write: []string{"charmers"},
 		},
+		params.BetaChannel: {
```
rogpeppe, line 671: This is getting really quite repetitive I wonder if some kind of helper function might make things a little more concise.

rogpeppe:
LGTM but I wonder about devising a shorthand for all those verbose ACLs in tests.

internal/charmstore/migrations_integration_test.go
frankban, line 427: Good idea, done.

internal/v4/api_test.go
frankban, line 671: I think I'll handle this in another branch.

frankban:
:shipit:

jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1878/
Test PASSed.

frankban:
Thanks for the reviews!
:shipit:

jujugui:

PR #689, opened 2016-10-29
use 100-continue with resource POST

jrwren:
This allows the client to use Expect: 100-continue header to prevent
request body upload if there is an error.


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1913/
Test PASSed.

jrwren:
definitely DO NOT DO THIS. :)

PR #693, opened 2016-11-07
all: move to macaroon.v2-unstable

rogpeppe:
There's still a dependency on macaroon.v1 because one
of the charmrepo params uses it. That will change when
we can move juju to using macaroon.v2.

We delete the agent tests because they were a hassle to
update and the agent package is going to be deleted in the next PR.


jujugui:
Test PASSed.
http://ci.jujugui.org:8080//job/charmstore/1917/
Test PASSed.

jrwren (approved):
This is great. It reads like a how-to-better-use-bakery.

internal/v5/api.go
```diff
@@ -114,10 +125,13 @@ func New(pool *charmstore.Pool, config charmstore.ServerParams, rootPath string)
 	var permChecker permChecker
 	if config.IdentityAPIURL != "" && config.AgentUsername != "" && config.AgentKey != nil {
 		logger.Infof("group membership enabled at %s; agent %s", config.IdentityAPIURL, config.AgentUsername)
-		identityClient := idmclient.New(idmclient.NewParams{
+		identityClient, err := idmclient.New(idmclient.NewParams{
 			BaseURL: config.IdentityAPIURL,
 			Client:  agent.NewClient(config.AgentUsername, config.AgentKey),
 		})
+		if err != nil {
+			panic(err)
```
mhilton, line 133: Strictly speaking I'd say that as the IdentityAPIURL originally came from user input, the error case here cannot be attributed to being a bug and so a panic isn't appropriate. I suspect we should rethink the idmclient API if we need this to complete without errors.

On the other hand I know how hard it is to fail to parse a URL, so I'm not that worried about it.

mhilton (approved):
LGTM with one thought

internal/v5/api.go
rogpeppe: Yes. I think I'll do it in the next PR if that's OK.

rogpeppe:
:shipit:

jujugui:

PR #702, opened 2016-12-20
add s3 storage for blobs

jrwren:
Extract a blobstore "store" interface.
Implement provider pattern.
Keep backward compatible blobstore by default.

### DIFF
--- a/cmd/blobtool/main.go
+++ b/cmd/blobtool/main.go
@@ -1,31 +1,29 @@
-// Copyright 2012, 2013, 2014 Canonical Ltd.
+// Copyright 2016 Canonical Ltd.
 // Licensed under the AGPLv3, see LICENCE file for details.
 
-package main // import "gopkg.in/juju/charmstore.v5-unstable/cmd/charmd"
+package main // import "gopkg.in/juju/charmstore.v5-unstable/cmd/blobtool"
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
-	"net/http"
 	"os"
 	"path/filepath"
 
-	"github.com/gorilla/handlers"
 	"github.com/juju/loggo"
 	"gopkg.in/errgo.v1"
-	"gopkg.in/macaroon-bakery.v2-unstable/bakery"
-	"gopkg.in/macaroon-bakery.v2-unstable/httpbakery"
 	"gopkg.in/mgo.v2"
 	"gopkg.in/natefinch/lumberjack.v2"
 
-	"gopkg.in/juju/charmstore.v5-unstable"
 	"gopkg.in/juju/charmstore.v5-unstable/config"
-	"gopkg.in/juju/charmstore.v5-unstable/elasticsearch"
+	"gopkg.in/juju/charmstore.v5-unstable/internal/charmstore"
+	"gopkg.in/juju/charmstore.v5-unstable/internal/mongodoc"
 )
 
 var (
-	logger        = loggo.GetLogger("charmd")
-	loggingConfig = flag.String("logging-config", "", "specify log levels for modules e.g. <root>=TRACE")
+	logger        = loggo.GetLogger("blobtool")
+	loggingConfig = flag.String("logging-config", "", "specify log levels for modules e.g. TRACE or '<root>=DEBUG;mgo=WARN'")
+	filter        = flag.String("filter", "", `json filter string to use as mongodb query. e.g. '{"user":"evarlast"}' or '{"size":{"$lt": 10000000}}'`)
 )
 
 func main() {
@@ -44,13 +42,13 @@ func main() {
 			os.Exit(1)
 		}
 	}
-	if err := serve(flag.Arg(0)); err != nil {
+	if err := run(flag.Arg(0)); err != nil {
 		fmt.Fprintf(os.Stderr, "%v\n", err)
 		os.Exit(1)
 	}
 }
 
-func serve(confPath string) error {
+func run(confPath string) error {
 	logger.Infof("reading configuration")
 	conf, err := config.Read(confPath)
 	if err != nil {
@@ -69,26 +67,11 @@ func serve(confPath string) error {
 	}
 	db := session.DB(dbName)
 
-	var es *elasticsearch.Database
-	if conf.ESAddr != "" {
-		es = &elasticsearch.Database{
-			Addr: conf.ESAddr,
-		}
+	if conf.BlobStorageProviders == nil {
+		return errgo.New("a provider-config section must be in the input config file")
 	}
 
-	keyring := bakery.NewPublicKeyRing()
-	err = addPublicKey(keyring, conf.IdentityLocation, conf.IdentityPublicKey)
-	if err != nil {
-		return errgo.Mask(err)
-	}
-	if conf.TermsLocation != "" {
-		err = addPublicKey(keyring, conf.TermsLocation, conf.TermsPublicKey)
-		if err != nil {
-			return errgo.Mask(err)
-		}
-	}
-
-	logger.Infof("setting up the API server")
+	logger.Infof("instantiating the store")
 	cfg := charmstore.ServerParams{
 		AuthUsername:            conf.AuthUsername,
 		AuthPassword:            conf.AuthPassword,
@@ -101,7 +84,6 @@ func serve(confPath string) error {
 		MaxMgoSessions:          conf.MaxMgoSessions,
 		HTTPRequestWaitDuration: conf.RequestTimeout.Duration,
 		SearchCacheMaxAge:       conf.SearchCacheMaxAge.Duration,
-		PublicKeyLocator:        keyring,
 		BlobStorageProviders:    conf.BlobStorageProviders,
 	}
 
@@ -113,33 +95,54 @@ func serve(confPath string) error {
 		}
 	}
 
-	server, err := charmstore.NewServer(db, es, "cs", cfg, charmstore.Legacy, charmstore.V4, charmstore.V5)
+	pool, err := charmstore.NewPool(db, nil, nil, cfg)
 	if err != nil {
-		return errgo.Notef(err, "cannot create new server at %q", conf.APIAddr)
+		return errgo.Notef(err, "cannot create a new store")
 	}
-	handler := server.(http.Handler)
-	if conf.AccessLog != "" {
-		accesslog := &lumberjack.Logger{
-			Filename:   conf.AccessLog,
-			MaxSize:    500, // megabytes
-			MaxBackups: 3,
-			MaxAge:     28, //days
-		}
-		handler = handlers.CombinedLoggingHandler(accesslog, handler)
+	store := pool.Store()
+	defer store.Close()
+
+	logger.Infof("updating entities")
+	if err := action(store); err != nil {
+		return errgo.Notef(err, "cannot update entities")
 	}
-	logger.Infof("starting the API server")
-	return http.ListenAndServe(conf.APIAddr, handler)
+
+	logger.Infof("done")
+	return nil
 }
 
-func addPublicKey(ring *bakery.PublicKeyRing, loc string, key *bakery.PublicKey) error {
-	if key != nil {
-		return ring.AddPublicKeyForLocation(loc, false, key)
-	}
-	pubKey, err := httpbakery.PublicKeyForLocation(http.DefaultClient, loc)
+func action(store *charmstore.Store) error {
+	entities := store.DB.Entities()
+	var filterMap map[string]interface{}
+	err := json.Unmarshal([]byte(*filter), &filterMap)
 	if err != nil {
-		return errgo.Mask(err)
+		return errgo.Notef(err, "could not json unmarshal filter: %s", filter)
 	}
-	return ring.AddPublicKeyForLocation(loc, false, pubKey)
+	iter := entities.Find(filterMap).Iter()
+	defer iter.Close()
+
+	successCounter := 0
+	failCounter := 0
+	var entity mongodoc.Entity
+	for iter.Next(&entity) {
+		logger.Debugf("processing %s", entity.URL)
+		blob, err := store.OpenBlob(charmstore.EntityResolvedURL(&entity))
+		if err != nil {
+			logger.Warningf("cannot open archive data for %s: %s", entity.URL, err)
+			failCounter++
+			continue
+		}
+		defer blob.Close()
+		logger.Debugf("copying %s %s", entity.URL, entity.BlobName)
+		store.BlobStore.Put(blob, entity.BlobName, blob.Size, blob.Hash, nil)
+		successCounter++
+		if successCounter%100 == 0 {
+			logger.Infof("%d entities written", successCounter)
+		}
+	}
+	logger.Infof("%d entities written. %d failed to read.", successCounter, failCounter)
+
+	return nil
 }
 
 var mgoLogger = loggo.GetLogger("mgo")
--- a/cmd/charmd/config.yaml
+++ b/cmd/charmd/config.yaml
@@ -23,3 +23,30 @@ identity-location: https://api.jujucharms.com/identity/v1/discharger
 # Uncomment to test with a terms service running locally
 #terms-location: localhost:8085
 access-log: /var/log/charmstore/access.log
+provider-config:
+    - type: jujublobstore
+      mongo_addr: localhost:27017
+      mongo_dbname: juju
+      bucket_name: entitystore
+      mode: read-write
+# An example config using s3 and ceph:
+#    - type: s3
+#      bucket_name: charmstoretestbucket
+#      disable_ssl: true
+#      endpoint: http://10.0.5.117
+#      key: O5W6PMIQZ83ODYCGVIGJ
+#      secret: 6aqf5vyRONOvGFJkvH65xW7ttxZIKNZx0c2cPMTA
+#      mode: read-write
+# An example config using localfs. bucket_name is a relative or absolute directory name.
+#    - type: localfs
+#      bucket_name: localstore
+#      mode: read-write
+# An example config using s3 and a fake s3 server.
+    #    - type: s3
+    #  bucket_name: charmstoretestbucket
+    #  disable_ssl: true
+    #  endpoint: "192.168.15.197:8000"
+    #  key: accessKey1
+    #  region: us-east-1
+    #  s3_force_path_style: true
+    #  secret: verySecretKey1
--- a/internal/blobstore/blobstore.go
+++ b/internal/blobstore/blobstore.go
@@ -1,4 +1,4 @@
-// Copyright 2014 Canonical Ltd.
+// Copyright 2014-2016 Canonical Ltd.
 // Licensed under the AGPLv3, see LICENCE file for details.
 
 package blobstore // import "gopkg.in/juju/charmstore.v5-unstable/internal/blobstore"
@@ -76,22 +76,65 @@ func NewContentChallengeResponse(chal *ContentChallenge, r io.ReadSeeker) (*Cont
 	}, nil
 }
 
+// Store stores data blobs in the configured store.
+type Store struct {
+	store
+}
+
 // Store stores data blobs in mongodb, de-duplicating by
 // blob hash.
-type Store struct {
+type store interface {
+	Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (*ContentChallenge, error)
+	PutUnchallenged(r io.Reader, name string, size int64, hash string) error
+	Open(name string) (ReadSeekCloser, int64, error)
+	Remove(name string) error
+}
+
+// gridStore stores data blobs in mongodb gridfs, de-duplicating by
+// blob hash.
+type gridStore struct {
 	mstore blobstore.ManagedStorage
+	db     *mgo.Database
+	prefix string
 }
 
 // New returns a new blob store that writes to the given database,
 // prefixing its collections with the given prefix.
+// This is the legacy function. Prefer using NewGridFSFromProviderConfig
 func New(db *mgo.Database, prefix string) *Store {
 	rs := blobstore.NewGridFS(db.Name, prefix, db.Session)
 	return &Store{
-		mstore: blobstore.NewManagedStorage(db, rs),
+		&gridStore{
+			mstore: blobstore.NewManagedStorage(db, rs),
+			db:     db,
+			prefix: prefix,
+		},
 	}
 }
 
-func (s *Store) challengeResponse(resp *ContentChallengeResponse) error {
+// NewBlobstoreFromProviderConfig returns a new blob stor that writes to mongodb
+// as defined in the given ProviderConfig
+func NewBlobstoreFromProviderConfig(pc *ProviderConfig) *Store {
+	if pc.MongoAddr == "" {
+		logger.Errorf("gridfs config with empty mongo_addr")
+	}
+	session, err := mgo.Dial(pc.MongoAddr)
+	if err != nil {
+		panic(errgo.Notef(err, "cannot dial mongo at %q", pc.MongoAddr))
+	}
+	dbName := pc.MongoDBName
+	if pc.MongoDBName == "" {
+		dbName = "juju"
+		logger.Debugf("gridfs mongo_dbname was empty, defaulting to juju")
+	}
+	db := session.DB(dbName)
+	if pc.BucketName == "" {
+		logger.Debugf("gridfs bucket_name was empty, mgo will default to fs")
+	}
+	return New(db, pc.BucketName)
+}
+
+func (s *gridStore) challengeResponse(resp *ContentChallengeResponse) error {
 	id, err := strconv.ParseInt(resp.RequestId, 10, 64)
 	if err != nil {
 		return errgo.Newf("invalid request id %q", id)
@@ -106,7 +149,7 @@ func (s *Store) challengeResponse(resp *ContentChallengeResponse) error {
 // satisfied by a client to prove that they have access to the content.
 // If the proof has already been acquired, it should be passed in as the
 // proof argument.
-func (s *Store) Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (*ContentChallenge, error) {
+func (s *gridStore) Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (*ContentChallenge, error) {
 	if proof != nil {
 		err := s.challengeResponse(proof)
 		if err == nil {
@@ -140,12 +183,12 @@ func (s *Store) Put(r io.Reader, name string, size int64, hash string, proof *Co
 // storage, with the provided name. The content should have the given
 // size and hash. In this case a challenge is never returned and a proof
 // is not required.
-func (s *Store) PutUnchallenged(r io.Reader, name string, size int64, hash string) error {
+func (s *gridStore) PutUnchallenged(r io.Reader, name string, size int64, hash string) error {
 	return s.mstore.PutForEnvironmentAndCheckHash("", name, r, size, hash)
 }
 
 // Open opens the entry with the given name.
-func (s *Store) Open(name string) (ReadSeekCloser, int64, error) {
+func (s *gridStore) Open(name string) (ReadSeekCloser, int64, error) {
 	r, length, err := s.mstore.GetForEnvironment("", name)
 	if err != nil {
 		return nil, 0, errgo.Mask(err)
@@ -154,6 +197,6 @@ func (s *Store) Open(name string) (ReadSeekCloser, int64, error) {
 }
 
 // Remove the given name from the Store.
-func (s *Store) Remove(name string) error {
+func (s *gridStore) Remove(name string) error {
 	return s.mstore.RemoveForEnvironment("", name)
 }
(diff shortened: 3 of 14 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2019/

cmd/blobtool/main.go
```diff
@@ -0,0 +1,159 @@
...
+	logger.Infof("updating entities")
+	if err := action(store); err != nil {
+		return errgo.Notef(err, "cannot update entities")
+	}
+
+	logger.Infof("done")
+	return nil
+}
+
+func action(store *charmstore.Store) error {
```
mhilton, line 114: shouldn't this function be called something like updateEntities?

cmd/blobtool/main.go
```diff
@@ -0,0 +1,159 @@
...
+	entities := store.DB.Entities()
+	var filterMap map[string]interface{}
+	err := json.Unmarshal([]byte(*filter), &filterMap)
+	if err != nil {
+		return errgo.Notef(err, "could not json unmarshal filter: %s", filter)
```
mhilton, line 119: how about "invalid filter \"%s\""

cmd/blobtool/main.go
```diff
@@ -0,0 +1,159 @@
...
+	var entity mongodoc.Entity
+	for iter.Next(&entity) {
+		logger.Debugf("processing %s", entity.URL)
+		blob, err := store.OpenBlob(charmstore.EntityResolvedURL(&entity))
+		if err != nil {
+			logger.Warningf("cannot open archive data for %s: %s", entity.URL, err)
+			failCounter++
+			continue
+		}
+		defer blob.Close()
```
mhilton, line 135: deferring in a loop is a bit of a smell. As it is if the filter is wide enough you'll have a large number of open blobs by the end of it. Perhaps factor out some of the loop body into a function so that blobs are closed as you go along. Alternatively as you don't seem to shortcut out of the loop body anyway just close the blob at the end.

cmd/blobtool/main.go
```diff
@@ -0,0 +1,159 @@
...
+		logger.Debugf("copying %s %s", entity.URL, entity.BlobName)
+		store.BlobStore.Put(blob, entity.BlobName, blob.Size, blob.Hash, nil)
```
mhilton, line 137: This will almost certainly return an error. it is probably a good idea to check for it.

internal/blobstore/blobstore.go
```diff
@@ -76,22 +76,65 @@ func NewContentChallengeResponse(chal *ContentChallenge, r io.ReadSeeker) (*Cont
 	}, nil
 }
 
+// Store stores data blobs in the configured store.
+type Store struct {
```
mhilton, line 80: this doesn't seem to be particularly worthwhile. As we're in an internal package we have quite a lot more leeway to make non backwards compatible changes. Make Store the interface you desire (as in store below).

internal/blobstore/blobstore.go
```diff
@@ -76,22 +76,65 @@ func NewContentChallengeResponse(chal *ContentChallenge, r io.ReadSeeker) (*Cont
...
 // blob hash.
-type Store struct {
+type store interface {
+	Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (*ContentChallenge, error)
+	PutUnchallenged(r io.Reader, name string, size int64, hash string) error
+	Open(name string) (ReadSeekCloser, int64, error)
+	Remove(name string) error
+}
+
+// gridStore stores data blobs in mongodb gridfs, de-duplicating by
```
mhilton, line 93: I suggest separating the gridfs bits to a different file and concentrating on the interface in this file. But that may just be java thinking coming back to me.

internal/blobstore/blobstore.go
```diff
@@ -76,22 +76,65 @@ func NewContentChallengeResponse(chal *ContentChallenge, r io.ReadSeeker) (*Cont
...
+// blob hash.
+type gridStore struct {
 	mstore blobstore.ManagedStorage
+	db     *mgo.Database
+	prefix string
 }
 
 // New returns a new blob store that writes to the given database,
 // prefixing its collections with the given prefix.
+// This is the legacy function. Prefer using NewGridFSFromProviderConfig
```
mhilton, line 103: Probably remove this function now. fix all the places in the code that called it.

internal/blobstore/blobstore.go
```diff
@@ -76,22 +76,65 @@ func NewContentChallengeResponse(chal *ContentChallenge, r io.ReadSeeker) (*Cont
...
+			mstore: blobstore.NewManagedStorage(db, rs),
+			db:     db,
+			prefix: prefix,
+		},
+	}
+}
+
+// NewBlobstoreFromProviderConfig returns a new blob stor that writes to mongodb
+// as defined in the given ProviderConfig
+func NewBlobstoreFromProviderConfig(pc *ProviderConfig) *Store {
```
mhilton, line 117: Isn't this supposed to be NewGridFSFromProviderConfig?

internal/blobstore/config.go
```diff
@@ -0,0 +1,40 @@
+// Copyright 2016 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package blobstore // import "gopkg.in/juju/charmstore.v5-unstable/internal/blobstore"
+
+// ProviderConfig holds configuration for where blobs are stored.
```
mhilton, line 6: I'd suggest adding this to blobstore.go once gridfs bits are removed?

internal/blobstore/config.go
```diff
@@ -0,0 +1,40 @@
...
+type ProviderConfig struct {
```
mhilton, line 7: Bundling every type config into one structure is likely to get confusing very quickly. One of the really nice things with the yaml parser is you can have a few goes with the same data. One approach would be to just parse the type field and then decide what config type to produce based on that.

internal/blobstore/local.go
```diff
@@ -0,0 +1,57 @@
...
+type localFSStore struct {
+	path string
+}
+
+func NewLocalFS(pc *ProviderConfig) *Store {
+	return &Store{&localFSStore{path: pc.BucketName}}
+}
+
+func (s *localFSStore) Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (_ *ContentChallenge, err error) {
+	err = s.PutUnchallenged(r, name, size, hash)
```
mhilton, line 25: why not just `return nil, s.PutUnchallenged(r, name, size, hash)`

internal/blobstore/local.go
```diff
@@ -0,0 +1,57 @@
...
+	return
+}
+
+func (s *localFSStore) PutUnchallenged(r io.Reader, name string, size int64, hash string) error {
+	w, err := os.Create(path.Join(s.path, name))
+	if err != nil {
+		logger.Errorf("localfs could not put %s", name)
+		return err
+	}
+	io.Copy(w, r)
```
mhilton, line 35: please check for an error here.

internal/blobstore/local.go
```diff
@@ -0,0 +1,57 @@
...
+	return &Store{&localFSStore{path: pc.BucketName}}
+}
+
+func (s *localFSStore) Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (_ *ContentChallenge, err error) {
+	err = s.PutUnchallenged(r, name, size, hash)
+	return
+}
+
+func (s *localFSStore) PutUnchallenged(r io.Reader, name string, size int64, hash string) error {
+	w, err := os.Create(path.Join(s.path, name))
```
mhilton, line 30: you'll want to close w.

Also strictly speaking you should use filepath.Join for this.

internal/blobstore/local.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		f.Close()
+		return nil, 0, err
+	}
+	logger.Debugf("localfs openned %s", name)
+	return f, stat.Size(), nil
+}
+
+func (s *localFSStore) Remove(name string) error {
+	return os.Remove(name)
```
mhilton, line 56: shouldn't you joins s.path here too?

internal/blobstore/local_test.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2016 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package blobstore
+
+var _ store = (*localFSStore)(nil)
```
mhilton, line 6: Would you mind awfully writing some tests please?

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+	"gopkg.in/errgo.v1"
+)
+
+// multiStore is a store.
+var _ store = (*multiStore)(nil)
+
+// multiStore holds the Stores to use as the blobstore.
+type multiStore struct {
+	stores []*Store
+	// paralle arrays because i'm sloppy
```
mhilton, line 18: parallel

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+	storeTypes []string
+	storeModes []string
+}
+
+// NewMultiStore returns a new multiStore which is backed by
+// the given ProviderConfig.
+func NewMultiStore(bsps []ProviderConfig) *Store {
```
mhilton, line 25: I think the nicer API here is if the multi store takes a `[]Store` and put the metadat knowledge in the Store interface.

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return s
+}
+
+func (s *multiStore) Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (cc *ContentChallenge, err error) {
+	for i := range s.stores {
+		if s.storeModes[i] == "read-write" || s.storeModes[i] == "write-only" {
+			cc, err = s.stores[i].Put(r, name, size, hash, proof)
+			if err != nil {
+				logger.Errorf("multiStore PutUnchallenged %s in %s with hash %s err was: %v trying next(if available)", name, s.storeTypes[i], hash, err)
```
mhilton, line 61: Put rather than PutUnchallenged

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+			s.stores[i] = NewLocalFS(&bsp)
+		default:
+			panic("unknown provider-config type: " + bsp.Type)
+		}
+		logger.Debugf("%s blob storage provider configured by FaillbackStore: %v\n", bsp.Type, bsp)
+	}
+	return s
+}
+
+func (s *multiStore) Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (cc *ContentChallenge, err error) {
```
mhilton, line 56: Add a doc comment here please. Especially say what the semantics are if multiple stores are writable.

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+	for i := range s.stores {
+		if s.storeModes[i] == "read-write" || s.storeModes[i] == "write-only" {
+			cc, err = s.stores[i].Put(r, name, size, hash, proof)
```
mhilton, line 59: This seems dodgy to me if there are multiple writable stores then it will attempt to write to them all, but you will only ever see the result from the final attempt. Also if there are no writable stores a nil error is returned which seems wrong.

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+			if err != nil {
+				logger.Errorf("multiStore PutUnchallenged %s in %s with hash %s err was: %v trying next(if available)", name, s.storeTypes[i], hash, err)
+			}
+		}
+	}
+	return
+}
+
+func (s *multiStore) PutUnchallenged(r io.Reader, name string, size int64, hash string) (err error) {
```
mhilton, line 68: The same comments as for Put

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+			err = s.stores[i].PutUnchallenged(r, name, size, hash)
+			if err != nil {
+				logger.Errorf("multiStore PutUnchallenged %s in %s with hash %s err was: %v trying next(if available)", name, s.storeTypes[i], hash, err)
+			}
+		}
+	}
+	return
+}
+
+func (s *multiStore) Open(name string) (ReadSeekCloser, int64, error) {
```
mhilton, line 80: Please add a doc comment here

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+			logger.Debugf("multiStore open %s not found in %s err was: %v trying next(if available)", name, s.storeTypes[i], err)
+		}
+	}
+	return nil, 0, errgo.Newf("file not found %s", name)
+}
+
+func (s *multiStore) Remove(name string) (err error) {
+	for i := range s.stores {
+		if s.storeModes[i] == "read-write" || s.storeModes[i] == "write-only" {
+			err = s.stores[0].Remove(name)
```
mhilton, line 96: That 0 is extremely suspect.

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+			if err == nil {
+				return f, st, err
+			}
+			logger.Debugf("multiStore open %s not found in %s err was: %v trying next(if available)", name, s.storeTypes[i], err)
+		}
+	}
+	return nil, 0, errgo.Newf("file not found %s", name)
+}
+
+func (s *multiStore) Remove(name string) (err error) {
```
mhilton, line 93: Please add a doc comment.

internal/blobstore/s3store.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil { //this only happens on invalid bucket name that is when a name has any of /:@
+		panic(err)
+	}
+	err = bucket.PutBucket(s3.Private)
+	if err != nil {
+		log.Println("Failed to create bucket", err)
+	}
+}
+
+func (s *s3Store) Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (_ *ContentChallenge, err error) {
```
mhilton, line 48: Doc comments for all of these please.

internal/blobstore/s3store.go
```diff
@@ -0,0 +1,120 @@
...
+	err := bucket.PutReader(name, r, size, "application/octet-stream", s3.Private)
+	if err != nil {
+		logger.Errorf("put failed :%s", err)
+		return errgo.Mask(err)
+	}
+	logger.Debugf("successful put %s in bucket %s", name, s.bucket)
+	return nil
+}
+
+func (s *s3Store) Open(name string) (ReadSeekCloser, int64, error) {
```
mhilton, line 65: I can see you're not happy with this implementation, and neither am I. perhaps you can use List up front to get the size, and then wrap rc in a type that provides a limited in memory data cache so that you can seek back to the start if necessary (which is the only sensible place to seek). There might already be such a thing implemented somewhere.

internal/blobstore/s3store.go
```diff
@@ -0,0 +1,120 @@
...
+		bucket: pc.BucketName,
+		getS3:  getter,
+	}
+	s.createBucket()
+	return s
+}
+
+func (s *s3Store) createBucket() {
+	bucket, err := s.getS3().Bucket(s.bucket)
+	if err != nil { //this only happens on invalid bucket name that is when a name has any of /:@
```
mhilton, line 39: I always worry about statements like this when it isn't explicitly in a functions documentation.

internal/blobstore/s3store.go
```diff
@@ -0,0 +1,120 @@
...
+		logger.Errorf("put failed :%s", err)
+		return errgo.Mask(err)
+	}
+	logger.Debugf("successful put %s in bucket %s", name, s.bucket)
+	return nil
+}
+
+func (s *s3Store) Open(name string) (ReadSeekCloser, int64, error) {
+	svc := s.getS3()
+	bucket, _ := svc.Bucket(s.bucket)
```
mhilton, line 67: perhaps create an s.getBucket method?

internal/blobstore/s3store_test.go
```diff
@@ -0,0 +1,74 @@
...
+	c.Assert(err, gc.IsNil)
+	defer rc.Close()
+	c.Assert(length, gc.Equals, int64(len(content)))
+
+	data, err := ioutil.ReadAll(rc)
+	c.Assert(err, gc.IsNil)
+	c.Assert(string(data), gc.Equals, content)
+
+}
+
```
mhilton, line 59: Please test Remove too, and some error cases if possible.

mhilton (changes requested):
In general this is great stuff and most of my comments are minor thoughts. However it is lacking tests so I would like to see those before this merges.

cmd/blobtool/main.go
```diff
@@ -0,0 +1,159 @@
...
+
+var (
+	logger        = loggo.GetLogger("blobtool")
+	loggingConfig = flag.String("logging-config", "", "specify log levels for modules e.g. TRACE or '<root>=DEBUG;mgo=WARN'")
+	filter        = flag.String("filter", "", `json filter string to use as mongodb query. e.g. '{"user":"evarlast"}' or '{"size":{"$lt": 10000000}}'`)
+)
+
+func main() {
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "usage: %s [options] <config path>\n", filepath.Base(os.Args[0]))
```
rogpeppe, line 31: I'd like to see a little more description here please. It would be nice to have some idea what the tool is and how to use it without delving into the source.

rogpeppe: Also "blobtool" is a very non-specific name. How about giving it a name that actually implies what it might do?

cmd/charmd/config.yaml
```diff
@@ -23,3 +23,30 @@ identity-location: https://api.jujucharms.com/identity/v1/discharger
 # Uncomment to test with a terms service running locally
 #terms-location: localhost:8085
 access-log: /var/log/charmstore/access.log
+provider-config:
```
rogpeppe, line 26: store-config ?

cmd/charmd/config.yaml
```diff
@@ -23,3 +23,30 @@ identity-location: https://api.jujucharms.com/identity/v1/discharger
...
+    - type: jujublobstore
+      mongo_addr: localhost:27017
+      mongo_dbname: juju
+      bucket_name: entitystore
+      mode: read-write
+# An example config using s3 and ceph:
+#    - type: s3
+#      bucket_name: charmstoretestbucket
+#      disable_ssl: true
```
rogpeppe, line 35: Isn't this implied by the http URL?

internal/blobstore/blobstore.go
```diff
@@ -76,22 +76,65 @@ func NewContentChallengeResponse(chal *ContentChallenge, r io.ReadSeeker) (*Cont
...
+// Store stores data blobs in the configured store.
+type Store struct {
+	store
+}
+
 // Store stores data blobs in mongodb, de-duplicating by
 // blob hash.
-type Store struct {
+type store interface {
+	Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (*ContentChallenge, error)
```
rogpeppe, line 87: We never use this method. Why not delete it and rename PutUnchallenged to Put?

internal/blobstore/config.go
rogpeppe: +1.

I'd suggest putting each provider implementation independently in its own file (possibly in its own package).

cmd/charmd/config.yaml
rogpeppe: (it seems like it's never used anyway)

cmd/charmd/config.yaml
```diff
@@ -23,3 +23,30 @@ identity-location: https://api.jujucharms.com/identity/v1/discharger
...
+#      endpoint: http://10.0.5.117
+#      key: O5W6PMIQZ83ODYCGVIGJ
+#      secret: 6aqf5vyRONOvGFJkvH65xW7ttxZIKNZx0c2cPMTA
+#      mode: read-write
```
rogpeppe, line 39: Do we actually need to have the capability for non-read-write blob stores at this point?
If we do, then perhaps "read-only: true" (defaults to false) might be a better way of doing it.

internal/blobstore/blobstore.go
rogpeppe: +1
Just export the interface as Store.

internal/blobstore/multistore.go
rogpeppe: +1

internal/blobstore/multistore.go
rogpeppe: How about making the type and mode part of the Store interface?
Then you won't need the parallel slices.

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+		case "jujublobstore":
+			s.stores[i] = NewBlobstoreFromProviderConfig(&bsp)
+		case "s3":
+			s.stores[i] = NewS3(&bsp)
+		case "localfs":
+			s.stores[i] = NewLocalFS(&bsp)
+		default:
+			panic("unknown provider-config type: " + bsp.Type)
+		}
+		logger.Debugf("%s blob storage provider configured by FaillbackStore: %v\n", bsp.Type, bsp)
```
rogpeppe, line 51: FaillbackStore?

internal/blobstore/multistore.go
rogpeppe: Another reason this seems questionable is that we're given a reader which will be consumed by the first Put so the second put will be putting an empty blob.

internal/blobstore/s3store.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (s *s3Store) createBucket() {
+	bucket, err := s.getS3().Bucket(s.bucket)
+	if err != nil { //this only happens on invalid bucket name that is when a name has any of /:@
+		panic(err)
+	}
+	err = bucket.PutBucket(s3.Private)
+	if err != nil {
+		log.Println("Failed to create bucket", err)
```
rogpeppe, line 44: Further to mhilton's comments above, the function should definitely return an error in this case, surely?

internal/blobstore/s3store.go
rogpeppe: I'm not keen either. Reading all of a blob into memory is not OK.
I don't think we should use this until we can work out a decent implementation.
The S3 API does support retrieving an arbitrary byte range (using the Range header) and although amz.v3/s3 doesn't support it, it shouldn't be too much work to make it do so.

We do need support for seeking to arbitrary locations, as we make use of that to read parts of charm archives.

internal/blobstore/s3store.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (s *s3Store) Put(r io.Reader, name string, size int64, hash string, proof *ContentChallengeResponse) (_ *ContentChallenge, err error) {
+	err = s.PutUnchallenged(r, name, size, hash)
+	return
+}
+
+func (s *s3Store) PutUnchallenged(r io.Reader, name string, size int64, hash string) error {
+	svc := s.getS3()
+	bucket, _ := svc.Bucket(s.bucket) // Ignoring the error because we know this bucket name is valid.
+	err := bucket.PutReader(name, r, size, "application/octet-stream", s3.Private)
```
rogpeppe, line 56: This isn't doing any blob de-duplication when we put the same content twice. That was a big motivation behind the blob store implementation - are we really prepared to discard that functionality? It would be good to find out how much space we're actually saving with blob-deduplication in the charm store currently.

internal/blobstore/s3store.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		logger.Errorf("put failed :%s", err)
+		return errgo.Mask(err)
```
rogpeppe, line 59: Don't log, just return errgo.Notef(err, "whatever message")

internal/charmstore/server.go
```diff
@@ -90,6 +91,12 @@ type ServerParams struct {
 	// RootKeyPolicy holds the default policy used when creating
 	// macaroon root keys.
 	RootKeyPolicy mgostorage.Policy
+
+	// StorageProviders holds a list of storage providers in priority order.
+	// During a migration it will likely be configured to s3 and gridfs.
+	// AWS recommends using ~/.aws/configuration file for secure credentials,
```
rogpeppe, line 97: Does this recommendation really apply to our cloud-deployed charm store?

internal/charmstore/store.go
```diff
@@ -227,6 +227,17 @@ func (p *Pool) requestStoreNB(always bool) (*Store, error) {
 	return store, nil
 }
 
+// newBlobstore creates a new blobstore based on config and the given StoreDatabase.
+func (p *Pool) newBlobstore(db *mgo.Database) *blobstore.Store {
+	// If no BlobStoreProviders are in the config, use the legacy config of
+	// same mongo as entity database and entitystore as the gridfs prefix.
+	if 0 == len(p.config.BlobStorageProviders) {
```
rogpeppe, line 234: No need for the "0 ==" idiom in Go, please.

internal/charmstore/store.go
```diff
@@ -227,6 +227,17 @@ func (p *Pool) requestStoreNB(always bool) (*Store, error) {
 	return store, nil
 }
 
+// newBlobstore creates a new blobstore based on config and the given StoreDatabase.
+func (p *Pool) newBlobstore(db *mgo.Database) *blobstore.Store {
+	// If no BlobStoreProviders are in the config, use the legacy config of
```
rogpeppe, line 232: I'd prefer it if this was done at a different level. I think it might be better if the default provider was added when the config was read.

internal/charmstore/store.go
```diff
@@ -227,6 +227,17 @@ func (p *Pool) requestStoreNB(always bool) (*Store, error) {
...
+	// same mongo as entity database and entitystore as the gridfs prefix.
+	if 0 == len(p.config.BlobStorageProviders) {
+		logger.Debugf("using default gridfs blobstore with prefix entitystore - no provider-config")
+		return blobstore.New(db, "entitystore")
+	}
+	return blobstore.NewMultiStore(p.config.BlobStorageProviders)
```
rogpeppe, line 238: This is pretty heavyweight to be doing on every charmstore request. Better to change the interface so we don't need to create the multistore and parse all the configuration every time. We should chat about this - there are a bunch of possibilities.

internal/blobstore/multistore.go
```diff
@@ -0,0 +1,100 @@
...
+			if err != nil {
+				logger.Errorf("multiStore PutUnchallenged %s in %s with hash %s err was: %v trying next(if available)", name, s.storeTypes[i], hash, err)
+			}
+		}
+	}
+	return
+}
+
+func (s *multiStore) Open(name string) (ReadSeekCloser, int64, error) {
+	for i := range s.stores {
```
rogpeppe, line 81: This should almost certainly do the opens concurrently, perhaps with a delay after starting each one.

internal/blobstore/blobstore.go
rogpeppe: Also, I'd like to see more doc comments on the methods here to make their contracts clear.

internal/blobstore/blobstore.go
```diff
@@ -76,22 +76,65 @@ func NewContentChallengeResponse(chal *ContentChallenge, r io.ReadSeeker) (*Cont
...
+	}
+}
+
+// NewBlobstoreFromProviderConfig returns a new blob stor that writes to mongodb
+// as defined in the given ProviderConfig
+func NewBlobstoreFromProviderConfig(pc *ProviderConfig) *Store {
+	if pc.MongoAddr == "" {
+		logger.Errorf("gridfs config with empty mongo_addr")
+	}
+	session, err := mgo.Dial(pc.MongoAddr)
```
rogpeppe, line 121: ISTM that as we've got things in this PR, we'll be making a network connection to mongodb on every charmstore request. Probably not what we want.

rogpeppe: Also, panic is never the right reaction to a network error.

rogpeppe (changes requested):
I like the general thrust, but I think this could do with a bunch of thought before it lands, and the potential performance ramifications are large.

There are some significant issues that would prevent us being able to use this in production.

How about making a feature branch so we can land this in several stages without needing to be too picky to start with?

internal/blobstore/blobstore.go
jrwren: Thanks. For some reason I struggled with making this read as I wanted and ended up with this. I think my desire was to leave all other code untouched and wrapping this interface in a struct meant not visiting every place the current Store struct is used as *Store and changing it to Store. I've visited each use and updated to Store.

internal/blobstore/blobstore.go
jrwren: I like the type per file and filename matches the type convention, but we haven't been doing that AFAICT.

internal/blobstore/blobstore.go
jrwren: I call it Blobstore, since it is not directly GridFS, it is the github.com/juju/blobstore layer that is built upon GridFS. I'd rather use a name which reflects that it is not direct GridFS use.

PR #722, opened 2017-02-23
internal/blobstore: fix removing of blobs

rogpeppe:
Previously, we refused to remove a multipart blob if it was a completed
upload. That doesn't fit with the way that they're used - an upload is
created and then another step associates that upload with a document.

This change changes things so that each upload may be associated with an
"owner" (the thing that's going to refer to the data) and that's set
before actually doing the association.

This means that we can avoid the same blob being used for two different
entities as well as meaning that we have a better idea of when we can
remove orphaned blob data.


jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2058/

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2059/

mhilton (approved):
LGTM

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2060/

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2061/

fabricematrat (approved):

rogpeppe:
:shipit:

jujugui:

PR #733, opened 2017-03-03
Delete ES element.

fabricematrat:


jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2084/

internal/charmstore/store.go
```diff
@@ -1284,7 +1286,13 @@ func (s *Store) DeleteEntity(id *router.ResolvedURL) error {
 			return errgo.Notef(err, "cannot remove compatibility blob %s", name)
 		}
 	}
-
+	if s.ES == nil || s.ES.Database == nil {
```
mhilton: you don't need to do this check in both places, just call s.ES.delete and let it return nil if its not configured.

mhilton (approved):
LGTM with one comment

urosj (changes requested):
When charm/5 is deleted, charm/4 needs to be reindexed. We have to keep the promise of indexing the last revision/

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2085/

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2086/

PR #748, opened 2017-06-26
internal/blobstore: only swift needs to know about container

rogpeppe:
The concept of "container" was exposed in the ObjectStore
interface, but it's not inherent to an object store - of
the two current implementations, only one needs
to know about containers currently.

This also fixes the MongoDB blobstore implementation
so that it continues to use the same paths as the
currently deployed version, so deploying without
Swift won't cause all the blobs to "disappear".

### DIFF
--- a/internal/blobstore/blobstore.go
+++ b/internal/blobstore/blobstore.go
@@ -33,19 +33,18 @@ func NewHash() hash.Hash {
 // object store.
 type ObjectStore interface {
 	// Get gets an object.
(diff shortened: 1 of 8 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2116/

mhilton (approved):
LGTM

frankban (approved):
Looks good, nice simplification!

fabricematrat (approved):

rogpeppe:
:shipit:

jujugui:

PR #752, opened 2017-06-27
internal/charmstore: make mongo backend work when session dies

rogpeppe:
Also factor the direct Swift dependency out of the
charmstore server parameters, and stop using
a Swift blobstore in some packages where it
doesn't give extra test confidence.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"gopkg.in/errgo.v1"
+	"gopkg.in/goose.v2/identity"
 	"gopkg.in/macaroon-bakery.v2-unstable/bakery"
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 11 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2123/

internal/charmstore/server.go
```diff
@@ -89,44 +90,25 @@ type ServerParams struct {
...
-
-	// SwiftTenant holds the swift tenant to use for connecting to swift.
-	SwiftTenant string `yaml:"swift-tenant"`
-
-	// SwiftAuthMode holds the swift auth mode to use for connecting to swift.
-	SwiftAuthMode string `yaml:"swift-authmode"`
+	// NewBlobBackend returns a new blobstore backend
+	// that may use the given MongoDB database.
+	// If this is nil, a MongoDB backend will be used.
+	NewBlobBackend func(db *mgo.Database) blobstore.Backend
```
mhilton, line 111: I understand why this is here, but given the code you actually the value won't currently be used with any non-nil value for NewBlobBackend. It seems a real shame to leek the abstraction like that.

mhilton (approved):
LGTM with a thought

internal/charmstore/server.go
rogpeppe: I think we probably will need to use mongo for the swift backend.

config/config.go
```diff
@@ -42,35 +43,77 @@ type Config struct {
...
+}
+
+type BlobStoreType string
+
+const (
+	MongoDBBlobStore BlobStoreType = "mongodb"
+	SwiftBlobStore   BlobStoreType = "swift"
+)
+
+type SwiftAuthMode struct {
```
frankban: Docstring.

frankban (approved):

internal/charmstore/server.go
rogpeppe: done

config/config.go
rogpeppe: done

rogpeppe:
:shipit:

jujugui:

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2124/

PR #753, opened 2017-06-28
internal/mongodoc: implement storage scheme for hashes

rogpeppe:
With the new blob storage scheme, we want to store
the hash of all parts within the mongodoc entities
(because we're not planning to keep refcounts separately)
so the multipart index needs to store the hashes too.

To save space, we store the hashes as a contiguous
byte array (this saves about 10 bytes per index entry)
but we use a Go representation that is more convenient.

### DIFF
--- a/internal/mongodoc/hash.go
+++ b/internal/mongodoc/hash.go
@@ -1,0 +1,52 @@
+// Copyright 2017 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package mongodoc
+
+import (
+	"crypto/sha512"
+	"encoding/hex"
+
+	"gopkg.in/errgo.v1"
+	"gopkg.in/mgo.v2/bson"
+)
+
+// Hashes holds a slice of hash objects
+// that will encode to MongoDB in a space-efficient
+// way. Each of the strings must be a hex-encoded
+// SHA384 checksum as produced by blobstore.NewHash.
+type Hashes []string
+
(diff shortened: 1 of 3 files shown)


jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2125/

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2126/

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2127/

fabricematrat (approved):

frankban (approved):
very nice ty!

rogpeppe:
:shipit:

jujugui:

internal/mongodoc/hash.go
```diff
@@ -0,0 +1,52 @@
...
+	"encoding/hex"
+
+	"gopkg.in/errgo.v1"
+	"gopkg.in/mgo.v2/bson"
+)
+
+// Hashes holds a slice of hash objects
+// that will encode to MongoDB in a space-efficient
+// way. Each of the strings must be a hex-encoded
+// SHA384 checksum as produced by blobstore.NewHash.
```
mhilton, line 17: I'm still very dubious about the value of this, whilst the space saving may be worth it it will make it significantly harder to understand the mongodb document in isolation. I'm not certain that is a price worth paying.

mhilton (commented):
I was still going to apprive this, but that confused github. I still have reservations about the encoding.

PR #759, opened 2017-07-03
internal/charmstore: report blob stats

rogpeppe:
We add a size field to the blobref collection so
that we can easily find the sizes when garbage
collecting. We leave reporting on entity/resource
blob sizes for another day.

### DIFF
--- a/internal/monitoring/monitoring.go
+++ b/internal/monitoring/monitoring.go
@@ -28,11 +28,53 @@ var (
 		Name:      "blobstore_gc_duration",
 		Help:      "The processing duration a garbage collection in seconds",
 	})
+
+	blobCount = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "charmstore",
+		Subsystem: "archive",
+		Name:      "blob_count",
+		Help:      "The total number of stored blobs.",
+	})
+
+	maxBlobSize = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "charmstore",
+		Subsystem: "archive",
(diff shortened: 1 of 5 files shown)


mhilton (approved):
LGTM

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2137/

fabricematrat (approved):

internal/monitoring/monitoring.go
```diff
@@ -28,11 +28,49 @@ var (
...
+
+	meanBlobSize = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "charmstore",
+		Subsystem: "archive",
+		Name:      "mean_blob_size",
+		Help:      "The mean stored blob size",
+	})
 )
 
+type BlobStats struct {
```
frankban: Docstrings. Also, since this is already called BlobStats, maybe call the fields just Count, MaxSize and MeanSize?

frankban (approved):
Looks good!
I wonder if, instead of having the GC return stats, we could pass a statsCollector function to GC, which is called once for each blob ref. this way maybe the stats code becomes less important in GC? Take it or leave it.

internal/monitoring/monitoring.go
rogpeppe: done

rogpeppe:
:shipit:

jujugui:

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2139/

PR #760, opened 2017-07-11
Fix migration and bulk insert for blob refs

fabricematrat:


jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2149/

mhilton (approved):
LGTM thanks

PR #764, opened 2017-07-19
internal/charmstore: fix race in blobstore migration

rogpeppe:
We were reading from the entity and writing to it concurrently.
The race detector did in fact pick up the mistake, but we
hadn't run with the race detector enabled...

Also, adjust the logs a little so it's clear which entities are
being migrated, and log any errors when an invalid
hash is used with the GC.

### DIFF
--- a/internal/blobstore/blobstore.go
+++ b/internal/blobstore/blobstore.go
@@ -292,8 +292,9 @@ func NewRefs(n int) *Refs {
 // Add records that the given hash is referenced.
 // It ignores the hash if it's invalid.
 func (r *Refs) Add(hash string) {
-	data, ok := decodeHash(hash)
-	if !ok {
+	data, err := decodeHash(hash)
+	if err != nil {
+		logger.Errorf("cannot add bad hash %q: %v", hash, err)
 		return
 	}
 	r.refs[data] = struct{}{}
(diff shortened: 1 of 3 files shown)


fabricematrat (approved):
LGTM

jujugui:
http://ci-cge.jujugui.org:8080//job/charmstore/2155/

internal/blobstore/blobstore.go
```diff
@@ -292,8 +292,9 @@ func NewRefs(n int) *Refs {
 // Add records that the given hash is referenced.
 // It ignores the hash if it's invalid.
 func (r *Refs) Add(hash string) {
-	data, ok := decodeHash(hash)
-	if !ok {
+	data, err := decodeHash(hash)
+	if err != nil {
+		logger.Errorf("cannot add bad hash %q: %v", hash, err)
```
mhilton, line 297: This really shouldn't be an error I don't think. We use this with refs that are empty a lot prev5blobhashes that aren't specified for instance.

internal/charmstore/store.go
```diff
@@ -431,6 +431,9 @@ func (s *Store) ensureIndexes() error {
 // deleting all blobs that have not been referenced since
 // the given time.
 func (s *Store) BlobStoreGC(before time.Time) error {
+	// BEWARE: if this code does not add all the relevant blob
```
mhilton, line 434: who is this comment for?

mhilton (approved):
LGTM with a few thoughts.

internal/blobstore/blobstore.go
fabricematrat: @mhilton We added a check to not use add when it's empty in here https://github.com/juju/charmstore/pull/764/files#diff-20be4fa0bfcb7416de93e699d65b2d2aR459

internal/charmstore/store.go
rogpeppe: you!

rogpeppe: (just wanted something to say that this is important code)

rogpeppe:
:shipit:

jujugui:

PR #786, opened 2018-03-09
internal/{charmstore,v5}: fix list results

rogpeppe:
When some list results had different ACLs from others,
we filtered out all but the most recent revision, so
we could fail to show any results even though some
results could be shown.

We fix that by changing ListQuery so that it returns
all matching charms that are in different channels;
then we filter those results after retrieving from MongoDB.

This changes the semantics of sorting by owner because
sorting is now done based on the returned results which
don't necessarily contain the actual owner, so we sort by
the User field in the returned URL, which means that
promulgated charms will always be returned first in that case.


jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2174/

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2175/

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2176/

internal/v5/list_test.go
```diff
@@ -7,13 +7,13 @@ import (
...
 	"encoding/json"
 	"net/http"
-	"sort"
 	"strings"
 
 	"github.com/juju/idmclient"
 	"github.com/juju/loggo"
 	jc "github.com/juju/testing/checkers"
 	"github.com/juju/testing/httptesting"
+	"github.com/kr/pretty"
```
mhilton: "that shouldn't be there" - @rogpeppe 2018

mhilton (approved):
LGTM

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2177/

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2178/

fabricematrat (approved):

rogpeppe:
:shipit:

jujugui:

PR #813, opened 2018-07-15
internal/v5: include content hash header in docker resource downloads

mhilton:


jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2226/

internal/v5/resources.go
```diff
@@ -132,7 +134,14 @@ func (h *ReqHandler) serveDownloadResourceDocker(id *router.ResolvedURL, r *mong
 		}
 		resp.Password = password
 	}
-	return httprequest.WriteJSON(w, http.StatusOK, resp)
+	buf, err := json.Marshal(resp)
```
rogpeppe, line 137: OK I guess, but please add a comment saying that this is just a temporary hack.

internal/v5/resources.go
```diff
@@ -132,7 +134,14 @@ func (h *ReqHandler) serveDownloadResourceDocker(id *router.ResolvedURL, r *mong
...
+	if err != nil {
+		return errgo.Mask(err)
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set(params.ContentHashHeader, fmt.Sprintf("%x", sha512.Sum384(buf)))
+	_, err = w.Write(buf)
+	return errgo.Mask(err)
```
rogpeppe: Please don't return an error if the write fails. There's nothing we can do in that case other than produce an inconsistent body.

rogpeppe (approved):
LGTM I suppose, but I don't really like putting in cruft like this that will stay around forever with no good reason other than the fact it was a bit more convenient at the time.

rogpeppe:
Having looked at how deeply the "resource as bytes" paradigm is embedded in juju, this seems more reasonable to me, so LGTM without reservation this time (other than the error return), thanks.

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2240/

mhilton:
:shipit:

jujugui:

PR #817, opened 2018-07-23
internal/v5: Support symlink for files in archive.

fabricematrat:


jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2229/

internal/charmstore/archive.go
```diff
@@ -135,11 +137,27 @@ func (s *Store) OpenBlobFile(blob *Blob, filePath string) (io.ReadCloser, int64,
 		if err != nil {
 			return nil, 0, errgo.Notef(err, "unable to read file %q", filePath)
 		}
+		if file.Mode()&os.ModeSymlink != 0 {
+			defer content.Close()
+			url, err := ioutil.ReadAll(content)
+			if err != nil {
+				return nil, 0, errgo.Notef(err, "cannot read archive data for symlink %s", file.Name)
+			}
+			return nil, 0, ErrRedirect{URL: string(url)}
```
rogpeppe: It's conventional to return pointers for errors, so that it's easy to remember.

internal/charmstore/archive.go
```diff
@@ -135,11 +137,27 @@ func (s *Store) OpenBlobFile(blob *Blob, filePath string) (io.ReadCloser, int64,
 		if err != nil {
 			return nil, 0, errgo.Notef(err, "unable to read file %q", filePath)
 		}
+		if file.Mode()&os.ModeSymlink != 0 {
```
rogpeppe, line 143: Please document this behaviour in the doc comment for OpenBlobFile.

internal/charmstore/archive.go
```diff
@@ -135,11 +137,27 @@ func (s *Store) OpenBlobFile(blob *Blob, filePath string) (io.ReadCloser, int64,
...
+				return nil, 0, errgo.Notef(err, "cannot read archive data for symlink %s", file.Name)
+			}
+			return nil, 0, ErrRedirect{URL: string(url)}
+		}
 		return content, fileInfo.Size(), nil
 	}
 	return nil, 0, errgo.WithCausef(nil, params.ErrNotFound, "file %q not found in the archive", filePath)
 }
 
+type ErrRedirect struct {
```
rogpeppe: // ErrRedirect holds the error returned by OpenBlobFile when a symbolic link
    // is opened. The Path field holds the name of the file that the symbolic link
    // refers to.

internal/charmstore/archive.go
```diff
@@ -135,11 +137,27 @@ func (s *Store) OpenBlobFile(blob *Blob, filePath string) (io.ReadCloser, int64,
...
+	URL string
```
rogpeppe: This isn't a URL. Please call it "Path" not "URL".

internal/charmstore/archive.go
```diff
@@ -135,11 +137,27 @@ func (s *Store) OpenBlobFile(blob *Blob, filePath string) (io.ReadCloser, int64,
...
+}
+
+func (e ErrRedirect) Error() string {
+	return fmt.Sprintf("redirect to %v", e.URL)
```
rogpeppe: Let's make this look a bit more like an error. How about: `fmt.Sprintf("file is a symbolic link to %q", e.Path)` ?

internal/v5/archive.go
```diff
@@ -334,6 +335,17 @@ func (h *ReqHandler) serveArchiveFile(id *router.ResolvedURL, w http.ResponseWri
 func (h *ReqHandler) ServeBlobFile(w http.ResponseWriter, req *http.Request, id *router.ResolvedURL, blob *charmstore.Blob) error {
 	r, size, err := h.Store.OpenBlobFile(blob, req.URL.Path)
 	if err != nil {
+		if redirect, ok := err.(charmstore.ErrRedirect); ok {
```
rogpeppe: How about checking whether it's *not* a symlink and returning early if so, to save a level of indirection?

internal/v5/archive.go
```diff
@@ -334,6 +335,17 @@ func (h *ReqHandler) serveArchiveFile(id *router.ResolvedURL, w http.ResponseWri
...
+			p, err := router.RelativeURLPath(req.URL.Path, path.Join(path.Dir(req.URL.Path), redirect.URL))
```
rogpeppe: I think it might be worth calling path.Clean on redirect and checking that the result doesn't begin with ".." (so it's not possible to have links that point outside the charm itself.

Also, I've just thought of another thing which is... not trivial to fix: what about symlinks to directories?

For example, say there's a charm archive that contains:

     foo -> hooks
     hooks/install

What do we see if we try to open $charmid/archive/foo/install ?

internal/v5/archive_test.go
```diff
@@ -1207,14 +1207,33 @@ func (s *ArchiveSuite) assertArchiveFileContents(c *gc.C, zipFile *zip.ReadClose
...
-	c.Assert(rec.Body.Bytes(), gc.DeepEquals, expectBytes)
+	c.Assert(string(rec.Body.Bytes()), gc.DeepEquals, string(expectBytes))
 	headers := rec.Header()
 	c.Assert(headers.Get("Content-Length"), gc.Equals, strconv.Itoa(len(expectBytes)))
 	// We only have text files in the charm repository used for tests.
 	c.Assert(headers.Get("Content-Type"), gc.Equals, "text/plain; charset=utf-8")
 	assertCacheControl(c, rec.Header(), true)
 }
 
+func retrieveExpectedBytes(c *gc.C, zipFile *zip.ReadCloser, filePath string) (expectBytes []byte) {
```
rogpeppe, line 1218: This might be better named readZipFilePath, but honestly I think we could just leave things as they were and just test the symlink behaviour directly in the symlink test rather than calling assertArchiveFileContents.

internal/v5/archive_test.go
```diff
@@ -1178,25 +1179,24 @@ func (s *ArchiveSuite) TestArchiveFileGet(c *gc.C) {
 	s.assertArchiveFileContents(c, zipFile, "~charmers/utopic/all-hooks-0/archive/hooks/install")
 }
 
+func (s *ArchiveSuite) TestSymLinkArchiveFileGet(c *gc.C) {
```
rogpeppe, line 1182: I'd like to see a few more tests of this behaviour. For example, does it work when the symlink is absolute, contains multiple slashes, trailing slashes? Does it work when the URL path contains a trailing slash or isn't clean?

rogpeppe (changes requested):
This is a good start, thanks. I'd like to see more tests, and we should probably chat about what to do about symlinks to directories.

internal/v5/archive.go
fabricematrat: it's not possible to have links that point outside the charm itself as we only read in the zip file.

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2230/

internal/charmstore/archive.go
fabricematrat: done.

internal/charmstore/archive.go
fabricematrat: done.

internal/charmstore/archive.go
fabricematrat: done.

internal/charmstore/archive.go
fabricematrat: done.

internal/charmstore/archive.go
fabricematrat: done.

internal/v5/archive.go
fabricematrat: done.

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2231/

internal/v5/archive.go
```diff
@@ -23,6 +23,7 @@ import (
 	"gopkg.in/juju/charmstore.v5/internal/charmstore"
 	"gopkg.in/juju/charmstore.v5/internal/mongodoc"
 	"gopkg.in/juju/charmstore.v5/internal/router"
+	"path"
```
mhilton, line 26: this seems in the wrong place

internal/v5/archive.go
```diff
@@ -333,6 +334,17 @@ func (h *ReqHandler) serveArchiveFile(id *router.ResolvedURL, w http.ResponseWri
...
 	r, size, err := h.Store.OpenBlobFile(blob, req.URL.Path)
+	if redirect, ok := err.(*charmstore.ErrRedirect); ok {
+		p, err := router.RelativeURLPath(req.URL.Path, path.Join(path.Dir(req.URL.Path), redirect.Path))
+		if err != nil {
+			return errgo.Notef(err, "cannot make relative URL from %q and %q", req.URL.Path, path.Join(path.Dir(req.URL.Path), redirect.Path))
+		}
+		// We can't use the http.redirect function because it tries to build an absolute url but the path in the
+		// request URL has been changed.
+		w.Header().Set("Location", p)
+		w.WriteHeader(http.StatusMovedPermanently)
```
mhilton, line 345: I think StatusFound would be better. We cannot say it is moved perminently only until a new zip file is uploaded.

internal/v5/archive_test.go
mhilton: I'd say if a zipfile contains an absolute symlink then it is broken.

mhilton (changes requested):

PR #819, opened 2018-08-01
internal/blobstore: use different swift library

mhilton:


jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2232/

cmd/charmd/main.go
```diff
@@ -123,15 +122,25 @@ func serve(conf *config.Config) error {
...
-			Region:     conf.SwiftRegion,
-			TenantName: conf.SwiftTenant,
+		authVersion := 0
+		switch conf.SwiftAuthMode {
+		case "legacy":
+			authVersion = 1
+		case "userpass":
+			authVersion = 2
+		case "userpassv3":
+			authVersion = 3
```
rogpeppe: maybe add a default case here?

internal/blobstore/swift.go
```diff
@@ -8,71 +8,128 @@ import (
...
 	}
-	if hash != fmt.Sprintf("%x", h.Sum(nil)) {
-		err := s.client.DeleteObject(s.container, name)
+
+	for i := 0; i < 2; i++ {
+		f, err := s.connection.ObjectCreate(s.container, name, true, "", "", swift.Headers{"Content-Length": fmt.Sprintf("%d", size)})
 		if err != nil {
-			logger.Errorf("could not delete object from container after a hash mismatch was detected: %v", err)
+			if errgo.Cause(err) == swift.AuthorizationFailed {
+				continue
```
rogpeppe: If this happens twice in a row, do we really want to return a nil error?

internal/blobstore/swift.go
```diff
@@ -8,71 +8,128 @@ import (
...
+		if size > maxBufferSize {
+			if _, err := io.Copy(w, r); err != nil {
+				f.Close()
+				return errgo.Mask(err)
+			}
+		}
+		if err := f.Close(); err != nil {
+			return errgo.Mask(err)
+		}
+		if hash != fmt.Sprintf("%x", h.Sum(nil)) {
```
rogpeppe: I think I'd reverse the order of the expressions here to put the just-calculated hash on the left. YMMV.

cmd/charmd/main.go
rogpeppe: Also, given that these are fairly arbitrary constants, have we got sufficient test coverage to be sure they're actually correct?

rogpeppe (changes requested):
Looks great in general, thanks. Hopefully the new library will be more reliable than the old one. I've a few comments and suggestions, but nothing too hard to fix.

cmd/charmd/main.go
mhilton: In the default case it is autodetected. Also they aren't that arbitrary they are protocol versions 1,2, or 3. Adding tests for this would require a significant amount of infrastructure. I'd be more inclined to remove it entirely.

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2233/

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2234/

internal/blobstore/swift.go
```diff
@@ -8,71 +8,129 @@ import (
...
+		break
 	}
-	return nil
+	if f != nil {
+		err1 := f.Close()
+		if err == nil {
+			err = err1
+		}
+	}
+	if err == nil && fmt.Sprintf("%x", h.Sum(nil)) != hash {
```
rogpeppe: if err != nil {
        return errgo.Mask(err)
    }

?

internal/blobstore/swift.go
```diff
@@ -8,71 +8,129 @@ import (
...
+		err1 := s.connection.ObjectDelete(s.container, name)
+		if err1 != nil {
+			logger.Errorf("could not delete object from container after a hash mismatch was detected: %v", err1)
+		}
+		err = errgo.New("hash mismatch")
```
rogpeppe: Any chance of a test for this case?

rogpeppe (approved):
LGTM with a couple more minor suggestions, thanks!

internal/blobstore/swift.go
mhilton: not easily, this hasn't really changed from the previous code so if it wasn't tested before it won't be now.

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2235/

mhilton:
:shipit:

jujugui:

PR #837, opened 2018-09-25
internal/charmstore: Add duration tracking on ElasticSearch population

mhemmings:


jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2261/

PR #841, opened 2018-09-27
Track more things for API responses.

mhemmings:
This adds endpoint and HTTP status tracking to Prometheus to allows us to gain more insightful metrics.

### DIFF
--- a/internal/monitoring/responsewriter.go
+++ b/internal/monitoring/responsewriter.go
@@ -1,0 +1,40 @@
+package monitoring
+
+import "net/http"
+
+// ResponseWriter wraps the http.ResponseWriter to give us the ability to
+// retrieve the HTTP status code at any point after the response has been written.
+//
+// Note: This is does not implement any of the optional methods implemented by
+// ResponseWriters. (i.e. CloseNotifier, Flusher, Hijacker)
+type ResponseWriter struct {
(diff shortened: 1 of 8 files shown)


jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2265/

mhilton (approved):
LGTM

internal/monitoring/responsewriter.go
```diff
@@ -0,0 +1,37 @@
+package monitoring
+
+import "net/http"
+
+// ResponseWriter wraps the http.ResponseWriter to give us the ability to
+// retrieve the HTTP status code at any point after the response has been written.
+type ResponseWriter struct {
+	http.ResponseWriter
```
rogpeppe, line 11: This hides all the other methods that the http ResponseWriter implementation implements that aren't part of the standard interface (e.g. Flusher, Hijacker, Pusher, CloseNotifier, ReaderFrom, StringWriter). If we don't care about any of them, then please add a comment to that effect, otherwise it might be worth passing through some of those interfaces explicitly.

rogpeppe (approved):
LGTM modulo one reservation, thanks.

jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2266/

PR #842, opened 2018-09-27
internal/monitoring: Track when ElasticSearch is syncing.

mhemmings:
We are only tracking when ES sync happens here. We don't explicitly care about precise duration as it's fairly meaningless. Having this change will show us when it's happening and for how long (i.e. is it seconds or hours?), and is easy to graph/analyse.


jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2267/

mhilton (approved):
LGTM

PR #855, opened 2018-10-19
internal/{v4,v5}: implement read-only mode

rogpeppe:
We'd like to be able to stop people making changes
to the charm store while we're upgrading, so
add a read-only mode to allow that possibility.


jujugui:
http://ci-gce.jujugui.org:8080//job/charmstore/2288/

internal/router/util.go
```diff
@@ -82,7 +82,9 @@ func errorToResp1(err error) (int, interface{}) {
 		status = http.StatusNotFound
 	case params.ErrBadRequest, params.ErrInvalidEntity:
 		status = http.StatusBadRequest
-	case params.ErrForbidden, params.ErrEntityIdNotAllowed:
+	case params.ErrForbidden,
+		params.ErrEntityIdNotAllowed,
+		params.ErrReadOnly:
 		status = http.StatusForbidden
```
mhilton, line 88: I feel this is more of a 503 Service Unavailable type situation. It should almost certainly be a 5XX error. There's nothing about the request that means it's forbidden.

mhilton (changes requested):
This doesn't look sufficient to me. It doesn't cover stats updates , for instance. and I suspect there are other unexpected places we write to the database.

internal/router/util.go
rogpeppe: You're forbidden from writing to the store. It seems that StatusForbidden might be appropriate. 503 doesn't seem quite right to me, as much of the service is actually available.

mhilton (approved):
As discussed

rogpeppe:
$$merge$$
:shipit:

jujugui:

PR #875, opened 2019-05-10
internal/series: add eoan series.

mhilton:
This also removes precise and trusty from search results. Search tests
have been reworked to not require hardcoded series names.

### DIFF
--- a/internal/charmstore/store.go
+++ b/internal/charmstore/store.go
@@ -802,7 +802,8 @@ var seriesScore = map[string]int{
 	"zesty":   8,
 	"artful":  9,
 	"cosmic":  10,
-	"disco":   0, // TODO: make this higher when disco is released
+	"disco":   11,
+	"eoan":    0, // TODO: make this higher when eoan is released
 
 	// When we find a multi-series charm (no series) we
 	// will always choose it in preference to a series-specific
--- a/internal/series/series.go
+++ b/internal/series/series.go
@@ -34,28 +34,40 @@ type SeriesInfo struct {
 	SearchBoost float64
 }
 
+const (
+	boostBundle     = 1.1255
+	boostLTS1       = 1.125
+	boostLTS2       = 1.1125
+	boostLTS3       = 1.11
+	boostS1         = 1.102
+	boostS2         = 1.101
+	boostS3         = 1.1
+	boostUnreleased = 0.9
+)
+
 // Series contains the data charmstore knows about series names
 var Series = map[string]SeriesInfo{
 	// Bundle
-	"bundle": {false, "", true, 1.1380},
+	"bundle": {false, "", true, boostBundle},
 
 	// Ubuntu
 	"oneiric": {true, Ubuntu, false, 0},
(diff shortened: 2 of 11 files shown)


internal/series/series.go
```diff
@@ -34,28 +34,40 @@ type SeriesInfo struct {
...
+	boostBundle     = 1.1255
+	boostLTS1       = 1.125
+	boostLTS2       = 1.1125
+	boostLTS3       = 1.11
+	boostS1         = 1.102
+	boostS2         = 1.101
+	boostS3         = 1.1
+	boostUnreleased = 0.9
+)
+
```
urosj, line 47: Nice!

internal/series/series.go
```diff
@@ -34,28 +34,40 @@ type SeriesInfo struct {
...
 
 	// Ubuntu
 	"oneiric": {true, Ubuntu, false, 0},
-	"precise": {true, Ubuntu, true, 1.1125},
+	"precise": {true, Ubuntu, false, 0},
 	"quantal": {true, Ubuntu, false, 0},
 	"raring":  {true, Ubuntu, false, 0},
 	"saucy":   {true, Ubuntu, false, 0},
-	"trusty":  {true, Ubuntu, true, 1.125},
+	"trusty":  {true, Ubuntu, false, 0},
```
urosj, line 59: We need to review this due to ESM and maybe revert indexing later.

internal/storetesting/search.go
```diff
@@ -0,0 +1,287 @@
...
+
+	"gopkg.in/juju/charmstore.v5/internal/mongodoc"
+	"gopkg.in/juju/charmstore.v5/internal/router"
+	"gopkg.in/juju/charmstore.v5/internal/series"
+)
+
+// SearchSeries contains the list of charm series that will be indexed in
+// the search in descending order of preference.
+var SearchSeries = func() []string {
+	var ss []string
```
urosj, line 20: Not sure it makes sense to execute this more than once.

urosj (commented):
Didn't finish. Will continue later.

internal/storetesting/search.go
mhilton: It is only executed once, at init time.

internal/charmstore/store.go
```diff
@@ -802,7 +802,8 @@ var seriesScore = map[string]int{
 	"zesty":   8,
 	"artful":  9,
 	"cosmic":  10,
-	"disco":   0, // TODO: make this higher when disco is released
+	"disco":   11,
+	"eoan":    0, // TODO: make this higher when eoan is released
```
urosj, line 806: Is it worth it? Why not just 12?

urosj (commented):
LGTM

internal/charmstore/store.go
mhilton: 12 would make it the default for any charm that was requested (and had a eoan charm available). That's probably not what you really want.

mhemmings (approved):

PR #881, opened 2019-05-24
internal/charmstore: add revision indexes to aid sorting

mhilton:


mhemmings (approved):

mhilton:
This is unlikely to be needed now

PR #882, opened 2019-05-29
internal/store: delete resources

mhilton:
Add support in the store for deleting resources. This has the same
semantics as deleting entities.

### DIFF
--- a/internal/charmstore/resources.go
+++ b/internal/charmstore/resources.go
@@ -137,6 +137,74 @@ func mapRevisions(resourceRevisions []mongodoc.ResourceRevision) map[string]int 
 	return revisions
 }
 
+// DeleteResource deletes the resource with the given id from the store.
+// If the resource is the currently published revision for any channel or
+// the last revision for base entity and resource name an error will be
+// returned with an ErrForbidden cause.
+func (s *Store) DeleteResource(id *router.ResolvedURL, rev mongodoc.ResourceRevision) error {
+	// Find all resources that use the Base URL of id and the
+	// resource name so we can refuse to delete the last resource.
+	var resources []mongodoc.Resource
+	err := s.DB.Resources().Find(bson.D{
+		{"baseurl", mongodoc.BaseURL(&id.URL)},
+		{"name", rev.Name},
+	}).All(&resources)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+
+	var res *mongodoc.Resource
+	for i := range resources {
+		if resources[i].Revision == rev.Revision {
+			res = &resources[i]
+			break
+		}
+	}
+
+	if res == nil {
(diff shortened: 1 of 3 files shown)


internal/charmstore/resources.go
```diff
@@ -137,6 +137,74 @@ func mapRevisions(resourceRevisions []mongodoc.ResourceRevision) map[string]int
 	return revisions
 }
 
+// DeleteResource deletes the resource with the given id from the store.
+// If the resource is the currently published revision for any channel or
+// the last revision for base entity and resource name an error will be
+// returned with an ErrForbidden cause.
```
cmars, line 143: A 404 or 410 seems more appropriate, but as long as it's well defined and works well with the existing clients, juju, etc, it's a win.

internal/charmstore/resources.go
```diff
@@ -137,6 +137,74 @@ func mapRevisions(resourceRevisions []mongodoc.ResourceRevision) map[string]int
...
+	if err != nil {
+		return errgo.Mask(err)
+	}
+
+	var res *mongodoc.Resource
+	for _, r := range resources {
+		if r.Revision == rev.Revision {
+			r := r
+			res = &r
+		}
```
cmars, line 161: break?

internal/charmstore/resources.go
```diff
@@ -137,6 +137,74 @@ func mapRevisions(resourceRevisions []mongodoc.ResourceRevision) map[string]int
...
+		{"name", rev.Name},
+	}).All(&resources)
+	if err != nil {
+		return errgo.Mask(err)
+	}
+
+	var res *mongodoc.Resource
+	for _, r := range resources {
+		if r.Revision == rev.Revision {
+			r := r
```
cmars: Would using the slice index be a more concise and direct way to get this done? Something like:

```
for i := range resources { ...
  if resources[i].Revision == rev.Revision {
    res = &resources[i]
    break
...
```

internal/charmstore/resources.go
```diff
@@ -137,6 +137,74 @@ func mapRevisions(resourceRevisions []mongodoc.ResourceRevision) map[string]int
...
+	// Remove the resource.
+	err = s.DB.Resources().Remove(bson.D{
+		{"baseurl", mongodoc.BaseURL(&id.URL)},
+		{"name", rev.Name},
+		{"revision", rev.Revision},
+	})
+	if err != nil {
+		if err == mgo.ErrNotFound {
+			// Someone else got there first.
+			err = params.ErrNotFound
```
cmars, line 200: This reassignment is weird to me, but maybe it's necessary or idiomatic in this project or an errgo thing?

cmars (approved):

frankban (approved):
Looks good with notes from Casey. Thanks!

PR #884, opened 2019-07-25
Reject uploads of bundles containing overlays

achilleasa:
This PR enables the charmstore to detect attempts to upload bundles whose `bundle.yaml` entry contains multiple documents and return an error to indicate that multi-document bundles are not currently supported by the charmstore.

On a side-note, the charm.v6 version imported by this PR includes helpers for dealing with multi-doc bundles (`LocalBundleDataSource`,`StreamBundleDataSource` and `ReadAndMergeBundleData`). This should make it easier to update the charmstore code so it can deal with multi-doc bundles in the future if that's something we want to do.


achilleasa:
Note: CI bails out complaining the mongo has shut down

mhilton (approved):
LGTM thanks

achilleasa:
@mhilton I have performed an audit of the code in the [gobwas/glob](https://github.com/gobwas/glob) that we have added to our dependencies and found no red flags. More specifically, the package only deals with string comparisons and performs no I/O (files, sockets etc.).

For additional context: the package is used by juju when a local charm is uploaded to a controller to filter the charm archive files (e.g. the '.tox/' folder contents) before re-packaging the charm archive (we landed this as a fix for https://bugs.launchpad.net/juju/+bug/1813799)

PR #885, opened 2019-08-05
Support uploads of bundles with per charm channels

achilleasa:
This PR enables charm store users to upload bundles that reference charms from different channels using bundles such as the one below:

```yaml
applications:
  dummy-in-edge:
    charm: cs:~juju-qa/bionic/dummy-in-edge
    channel: edge
  other-dummy-in-beta:
    charm: cs:~juju-qa/bionic/other-dummy-in-beta
    channel: beta
  wordpress:
    charm: wordpress
```

In the above example, the `dummy-in-edge` and `other-dummy-in-beta` charms have been uploaded to the charm store, published to the `edge` and `beta` channels (but **not** on `stable`) and made public via `charm grant $charm-url everyone`.

Prior to this PR, any attempts to upload the above bundle would fail with:
```
ERROR cannot post archive: bundle verification failed: ["application \"dummy-in-edge\" refers to non-existent charm \"cs:~juju-qa/bionic/dummy-in-edge\"","application \"other-dummy-in-beta\" refers to non-existent charm \"cs:~juju-qa/bionic/other-dummy-in-beta\""]
```

The changes from this PR, bring in the latest `charm.v6` dependency that includes an optional, per-application channel attribute. If defined, it will be used by the charm store code when resolving charms.

NOTE:
This PR includes a newer `charm.v6` version than the one included in #884. It is therefore important to ensure that this PR does not get merged before #884 lands!


internal/charmstore/addentity.go
```diff
@@ -894,3 +894,26 @@ func (e *entityCharm) Actions() *charm.Actions {
...
+}
+
+// requiredCharms returns a list of a bundle's charms and the preferred channel
+// for obtaining them.
+func requiredCharms(bd *charm.BundleData) []requiredCharm {
+	list := make([]requiredCharm, 0, len(bd.Applications))
+	for _, app := range bd.Applications {
+		list = append(list, requiredCharm{
+			charm:   app.Charm,
+			channel: params.Channel(app.Channel),
```
mhilton, line 910: I presume this is choosing stable by default.

internal/charmstore/addentity.go
```diff
@@ -894,3 +894,26 @@ func (e *entityCharm) Actions() *charm.Actions {
...
+		})
+	}
+
+	sort.Slice(list, func(a, b int) bool {
```
mhilton, line 914: normally we sort things so we have consistent output, I'm not sure that applies here, AFAICS we only use the output for validating the output and the order doesn't really matter there.

mhilton (approved):
LGTM, a couple of thoughts (take or leave)

internal/charmstore/addentity.go
achilleasa: Yes, an empty channel pulls from stable (this matches the current behavior).

internal/charmstore/addentity.go
achilleasa: I have included the sort to match the behavior of bundledata.RequiredCharms (see: https://github.com/juju/charm/blob/v6/bundledata.go#L425) which the store code invoked prior to my changes. Unfortunately, that implementation is oblivious to channels so I had to introduce a tuple type to keep track of {charm, channel} values and call slice.Sort on it.

PR #889, opened 2019-11-28
Issuing long lived delegatable macaroons and docker macaroons.

alesstimec:
- also added "focal" series

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/prometheus/client_golang v0.0.0-20180319131721-d49167c4b9f3
 	github.com/rogpeppe/fastuuid v0.0.0-20150106093220-6724a57986af
 	golang.org/x/net v0.0.0-20181029044818-c44066c5c816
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -5,6 +5,7 @@ require (
 	github.com/ajstarks/svgo v0.0.0-20181006003313-6ce6a3bcf6cd // indirect
 	github.com/dgrijalva/jwt-go v0.0.0-20160705203006-01aeca54ebda
 	github.com/gorilla/handlers v0.0.0-20170224193955-13d73096a474
+	github.com/juju/charmstore v5.5.1+incompatible // indirect
```
mhilton: This shouldn't be being imported, I guess there is a bad import somewhere.

mhilton (approved):
LGTM thanks just the weird dependency problem

go.mod
alesstimec: "go mod tidy" fixed the issue

PR #890, opened 2020-01-14
More privacy in perms

frankban:
Change the perms API endpoints so that they don't leak information about other users if the current user doesn't have write permissions on the entity.

### DIFF
--- a/internal/v5/api.go
+++ b/internal/v5/api.go
@@ -1085,14 +1085,9 @@ func (h *ReqHandler) metaPerm(entity *mongodoc.BaseEntity, id *router.ResolvedUR
 	if err != nil {
 		return nil, errgo.Mask(err)
 	}
-	acls := entity.ChannelACLs[ch]
-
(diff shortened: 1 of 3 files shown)


mhilton (approved):
LGTM

frankban:
The failure is intermittent, it succeeded in go 1.12.

internal/v5/api.go
```diff
@@ -1169,6 +1175,24 @@ func (h *ReqHandler) metaPermWithKey(entity *mongodoc.BaseEntity, id *router.Res
 	return nil, errgo.WithCausef(nil, params.ErrNotFound, "unknown permission")
 }
 
+func (h *ReqHandler) visibleACL(acls mongodoc.ACL) (mongodoc.ACL, error) {
```
urosj, line 1178: Could you add description of what the function does and expected results?

urosj (commented):
LGTM, but please add the description of the change.

PR #893, opened 2020-01-16
meta/perm: less naive implementation of privacy features

frankban:
The previous change to the meta/perm endpoint broke the "charm grant/revoke" commands, which rely on the GET perm/meta to always return the real ACLs even when the user only has read permission. We try to be more smart here about how the permissions are returned to the user.

### DIFF
--- a/internal/v5/api.go
+++ b/internal/v5/api.go
@@ -1107,20 +1107,25 @@ func (h *ReqHandler) putMetaPerm(id *router.ResolvedURL, path string, val *json.
 		return errgo.Mask(err)
 	}
 	// TODO use only one UpdateField operation?
-	updater.UpdateField(string("channelacls."+ch+".read"), perms.Read, &audit.Entry{
-		Op:     audit.OpSetPerm,
-		Entity: &id.URL,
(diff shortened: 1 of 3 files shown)


mhilton (approved):
LGTM

internal/v5/api.go
```diff
@@ -1175,24 +1180,40 @@ func (h *ReqHandler) metaPermWithKey(entity *mongodoc.BaseEntity, id *router.Res
...
-	ok, err := h.auth.User.Allow(acls.Write)
+	auth := h.auth
+	// Authenticate the user if not authenticated already, in order to check
+	// write permissions as well.
+	if auth.User == nil {
+		var err error
+		auth, err = h.Authenticate(req)
+		if err != nil {
+			// The user cannot authenticate, but the entity was readable by everyone anyway.
+			return respForEveryone, nil
```
tasdomas, line 1196: Do we assume this was readable by everyone because we got this far?

tasdomas (approved):

internal/v5/api.go
frankban: Yes.

PR #909, opened 2020-10-09
all: update dependencies

mhilton:
Update the github.com/juju/charm and github.com/juju/charmrepo
dependencies. The behaviour of charm.ParseURL has changed in the latest
version to assume that scheme-less URLs are charmhub ones. The
charmstore uses scheme-less URLs a lot, especially in tests, and always
uses them to refer to charmstore URLs. A shim layer has been added for
charmstore to maintain the existing behaviour.


frankban (approved):
Looks good!

